PR #14, opened 2023-01-20
Thread.cpp: change ucontext to ucontext_t

adro79:
Fix error when building on Linux (At least on Arch Linux)

Signed-off-by: adro79 <57686179+adro79@users.noreply.github.com>

### DIFF
--- a/src/lib/mu/Mu/Thread.cpp
+++ b/src/lib/mu/Mu/Thread.cpp
@@ -74,7 +74,7 @@ Thread::Thread(Process* p, bool appthread) :
 #endif
 
 #ifdef PLATFORM_LINUX
-    _threadState = (size_t*)(new ucontext);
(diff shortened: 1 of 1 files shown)


marcomeyerVFX:
not only on Arch, this fix is also required on any EL9.1 Distro

geffrak (approved):

geffrak:
Awesome! It doesn't seem to break anything with RHEL7.

rogernelson (approved):
Thank you for this. Great to know that RV can be built on other Linux distros.

adro79:
> Thank you for this. Great to know that RV can be built on other Linux distros.

Regarding building on Arch Linux, this is the only fix I was able to provide.
The other errors I found are mentioned in #15 and @marcomeyerVFX seems to have found the fixes.

PR #29, opened 2023-01-23
Correct boost.cmake byproducts

geffrak:
On my MacBook with macOS Ventura and Xcode 14.2, when I try to rebuild the "RV" target using Jetbrains CLion, it always rebuilds everything that depends on Boost.

Debugging with `-d explain` tells me that RV_DEPS_BOOST gets constantly rebuilt because of some boost dylibs listed in the BUILD_BYPRODUCTS not existing.

Example of the issue:
```
/opt/homebrew/bin/cmake --build /Users/geffrak/code/git/github.com/org-geffrak/OpenRV/cmake-build-debug --target RV -- -d explain 
ninja explain: output RV_DEPS_BOOST/install/lib/libboost_atomic.dylib doesn't exist
```

Before this commit, `_boost_libs` contains libraries that are not requested to the build of Boost. With this change, we explicitly request stuff to the build of Boost using that list.

None of the removed libraries are used by OpenRV, and OpenRV is not using all the remaining libraries.

### DIFF
--- a/cmake/dependencies/boost.cmake
+++ b/cmake/dependencies/boost.cmake
@@ -35,36 +35,19 @@ SET(_install_dir
 )
 
 SET(_boost_libs
-    atomic
     chrono
-    container
-    context
-    contract
-    coroutine
     date_time
     filesystem
     graph
     iostreams
     locale
-    log
-    log_setup
-    numpy39
-    prg_exec_monitor
     program_options
-    python39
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):

rogernelson (approved):
Thank you! This was so annoying.

Issue #25, opened 2023-01-21
atomics ops download fail (hash mishmatch) during windows build


finnschi:
Atomic Ops download has a hash mismatch, 

downloading this package: https://github.com/ivmai/libatomic_ops/archive/refs/heads/master.zip 

downloadings heads version which was changed yesterday from what I can tell the hash changed. 

I changed the hash in cmake/dependencies/atomic_ops.cmake
```
SET(_download_hash
      7b6cf4a584813e531b336c7d966bd929
) 

```
and will try to build again.  

maybe the hash check should be removed when downloading from head? 

```
  -- [download 99% complete]
  -- [download 100% complete]
  -- verifying file...
         file='C:/OpenRV/cmake-build/RV_DEPS_DOWNLOAD/RV_DEPS_ATOMIC_OPS_7.7.0.zip'
  -- MD5 hash of
      C:/OpenRV/cmake-build/RV_DEPS_DOWNLOAD/RV_DEPS_ATOMIC_OPS_7.7.0.zip
    does not match expected value
      expected: '8da38242d61d746e070e021a735798b9'
        actual: '7b6cf4a584813e531b336c7d966bd929'
  -- Hash mismatch, removing...
  CMake Error at C:/OpenRV/cmake-build/cmake/dependencies/RV_DEPS_ATOMIC_OPS-prefix/src/RV_DEPS_ATOMIC_OPS-stamp/download-RV_DEPS_ATOMIC_OPS.cmake:170 (message):
    Each download failed!
```

PR #31, opened 2023-01-23
Fix atomics ops download fail during windows build

bernie-laberge:
This issue adresses the following reported issue:
https://github.com/AcademySoftwareFoundation/OpenRV/issues/25

**Problem:**
The download of the libatomic_ops is now failing the hash verification causing a build break on Windows. 

**Cause:**
The windows build of RV depends on a recent version of libatomic_ops so that --disable-gpl can be used. To do so we were downloading the head of the repo while verifying it against a fixed hash. This hash is not longer valid since a recent commit in the libatomic_ops repo thus the error.

**Solution:**
Now downloading the libatomic_ops archive from a specific commit (https://github.com/ivmai/libatomic_ops/commit/044573903530c4a8e8318e20a830d4a0531b2035) instead of the HEAD of the repo.

Also in this commit:
Adding two .gitignore files to prevent generated files from being added inadvertently to the repo.


geffrak (approved):

src/lib/files/Gto/Gto/generated/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
rogernelson, line 1: It's probably a good idea to add a `!.gitignore` after the `*` to ensure any future changes to the gitignore are not ignored.

src/lib/mu/MuLang/generated/.gitignore
```diff
@@ -0,0 +1 @@
+*
```
rogernelson, line 1: Same here for the `!.gitignore`.

rogernelson (approved):

src/lib/files/Gto/Gto/generated/.gitignore
bernie-laberge: Excellent idea. I'll do it. Thanks Roger !

PR #35, opened 2023-01-24
Include sysctl on Mac only

### LINKED ISSUE #20: #include <sys/sysctl.h>  deprecated from glibc on recent distros add conditional import that only runs if the distro is old enough.
#include <sys/sysctl.h>  Deprecated from glibc on EL9.

Please add conditional statement that checks how recent your system is to OpenRV/src/lib/base/TwkUtil/SystemInfo.cpp,

rogernelson:
Closes #20 

The sysctl include is not needed on CentOS7 and as of glibc 2.32 it has been removed.  So support distros/compilers later than that by removing it completely on Linux.

### DIFF
--- a/src/lib/base/TwkUtil/SystemInfo.cpp
+++ b/src/lib/base/TwkUtil/SystemInfo.cpp
@@ -18,7 +18,7 @@
 #endif
 #include <windows.h>
 #include <assert.h>
-#else
+#elif __APPLE__
 #include <sys/sysctl.h>
 #endif
 


geffrak (approved):
Is it required for MacOS?

bernie-laberge (approved):

rogernelson:
> Is it really required for macOS?

Yes.  The Mac build fails with a bunch of sysctl errors if I remove it.

PR #49, opened 2023-02-02
Fix OTIO import problem when no available_range on audio

rogernelson:
This branch fixes a problem from a client OTIO that was being imported to RV.

The problem was that there was an audio clip which had no `available_range` on an audio-only clip (which is valid OTIO), so we would throw an exception when we tried to access its `start_time` property.

To fix it, we will use `trimmed_range` if no `available_range` is available.  This will mean the media will have exactly the same number of frames as the source (ie trimmed/ no handles) in these cases.

There were also a number of Effects were generating errors, so I added a hook for the generic Effect that will just print that we do not support effect [effect_name].  Users can add their own custom handling here for their effects.

### DIFF
--- a/src/plugins/rv-packages/otio_reader/otio_reader.py
+++ b/src/plugins/rv-packages/otio_reader/otio_reader.py
@@ -391,15 +391,16 @@ def _create_collection(collection, context=None):
         return results[0]
 
 
-def _create_media(media_ref, context=None):
+def _create_media(media_ref, trimmed_range, context=None):
     context = context or {}
+    media_range = media_ref.available_range or trimmed_range
 
     if isinstance(media_ref, otio.schema.ExternalReference):
         media = [_get_media_path(str(media_ref.target_url), context)]
 
         if context.get("track_kind", None) == otio.schema.TrackKind.Audio:
             # Create blank video media to accompany audio for valid source
-            blank = _create_movieproc(media_ref.available_range)
+            blank = _create_movieproc(media_range)
             # Appending blank to media promotes name of audio file in RV
             media.append(blank)
         return media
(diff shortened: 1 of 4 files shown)


geffrak (approved):

src/plugins/rv-packages/otio_reader/otio_reader.py
```diff
@@ -391,15 +391,16 @@ def _create_collection(collection, context=None):
         return results[0]
 
 
-def _create_media(media_ref, context=None):
+def _create_media(media_ref, trimmed_range, context=None):
```
geffrak, line 394: I believe context should be a [keyword-only argument](https://peps.python.org/pep-3102/). You added the new argument in the middle, so it might break existing code that passes the context as the second positional parameter. 

The end result would be `def _create_media(media_ref, trimmed_range, * ,context=None):`

Feel free to ignore the proposition.

geffrak (approved):

bernie-laberge (approved):

jminor (approved):
This fixes all of the audio problems I encountered. Thanks!

src/plugins/rv-packages/otio_reader/otio_reader.py
rogernelson: I get your point.  The _ prefix on the function name was meant to indicate that its a function private to this file.  Is there some Python-equivalent to a static function in C++ so I can absolutely ensure that no one is calling it outside of this file?

JeanChristopheMorinPerso: No. Everything is accessible in python. But `_` is the common way to say "this is private".

geffrak: Rebels tend to call `_` functions anyways, but not `__` :p

PR #50, opened 2023-02-03
Fix switch node EDL timings

rogernelson:
This is bringing back a fix I did earlier, but after the branch off for OpenRV, so it wasn't part of the initial OpenRV code.  The issue is that when we have a switch node.  Here's my original PR description:

This branch fixes the timing when a Sequence node without autoEDL is connected to a SourceNode via a SwitchNode. Normally, the SwitchNode is computing an offset for autoEDL mode so that its frames are always starting at 1 (so it can basically unwrap the autoEDL behaviour to find the real frames). With autoEDL though, raw frame numbers are input so we do not want this. The simplest fix I can think of is to pass the autoEDL mode in the context when we metaEvaluate the graph when computing the ranges. This way if we are connected to a Sequence node with autoEDL disabled, we can ignore the offset.

### DIFF
--- a/src/lib/ip/IPBaseNodes/IPBaseNodes/SwitchIPNode.h
+++ b/src/lib/ip/IPBaseNodes/IPBaseNodes/SwitchIPNode.h
@@ -94,6 +94,7 @@ private:
     FloatProperty*              m_outputFPS;
     IntProperty*                m_outputSize;
     IntProperty*                m_autoSize;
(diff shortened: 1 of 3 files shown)


bernie-laberge (approved):

geffrak (approved):

jminor (commented):
Thanks @rogernelson ! This fixes the OTIO timing problem I was having.

Issue #44, opened 2023-01-27
Sync sessions broken due to min RV version


chadrik:
Sync sessions don't currently work because the [minimum RV version is set to "2021.0.2"](https://github.com/AcademySoftwareFoundation/OpenRV/blob/41eccbe03bf2462b399fec6511be1d11de6c6327/src/plugins/rv-packages/sync/sync.mu#L29) and the [default version of OpenRV is "1.0.0"](https://github.com/AcademySoftwareFoundation/OpenRV/blob/41eccbe03bf2462b399fec6511be1d11de6c6327/cmake/globals/rv_version.cmake#L11).  This results in the error "Remote version = 1.0.0, we require 2021.0.2 or above".

One solution to this is to change `minRVVersion` to "1.0.0", but this prevents sync sessions between OpenRV and RV -- RV rejects OpenRV.   We have licenses of RV which we'd like to keep for the added features so we'd like for licensed users to be able to sync with open source users (without requiring them to juggle two different apps).  It would therefore be great if the version of OpenRV matched the version of RV. 

For now I'm setting the version at build time using the cmake defines, but I'm just guessing at the version.

rogernelson:
Good catch.  @bernie-laberge Maybe we should version the sync itself instead of using the RV version? We would need to patch RV versions already out there though.

chadrik:
Has it already been decided that the versioning for OpenRV will differ from RV, or is that still up for debate?

rogernelson:
I don't think it's set it stone.  The version was reset to 1.0 to a) signify a fresh start for OpenRV; and b) releases won't be on a fix schedule tied to the year anymore.

But if it is causing more problems that it solves, it could be worth just putting it back to match the commercial version.  Although, they will eventually become out of sync again, so it might be just kicking the can down the road...

compaga:
We will eventually release a new version of RV that should fix the version miss match for package like Sync. This should allow Sync to work between Open RV and future RV release.

bernie-laberge:
I am closing this. issue as I have just merged this PR which addresses it : https://github.com/AcademySoftwareFoundation/OpenRV/pull/51
Thank you @chadrik !

chadrik:
> We will eventually release a new version of RV that should fix the version miss match for package like Sync. This should allow Sync to work between Open RV and future RV release.

[...]

PR #51, opened 2023-02-03
SG-29775 Fix RvSync broken in Open RV

bernie-laberge:
SG-29775: Fix RvSync sessions broken due to min RV version in Open RV

This commit fixes the following issue:
https://github.com/AcademySoftwareFoundation/OpenRV/issues/44

####Problem:
The RV Sync package has a check to validate remotely connected contacts against a minimum version of RV (eg. RV 2021.0.3). This check fails on Open RV because the Open RV version has been reset to 1.0.0.
Note that the goal here is to find a solution where participants of an RV Sync session can be a mix of Open RV and Commercial RV users.

####Solution:
Check against the internal version of the Rv Sync package instead of a minimum RV version like it was done before.

Rationale:
Even though the Commercial version of RV will eventually be built off Open RV, it is assumed that their versioning will diverge. For RV packages, a new minimal Open RV version requirement (named openrv) was added to the PACKAGE file to complement the minimal Commercial RV version requirement (named rv).  This way, even though both Commercial and Open RV will have different versions, the goal is for RV packages to be the same for both. Therefore it is the RV Sync package version that should be checked against each other to see if they are compatible or not.

In this commit, the currently minimum version of RV specified in the Sync.mu code was transferred to the Sync' PACKAGE file. The internal version of the sync package was bumped and the checkRVVersion was replaced with checkSyncVersion().

Note that I tried to find a solution that would be forward compatible so that an Open RV sync session could be joined by Commercial RV (old version <= 2022.3.1) users. However I was not able to because in the Commercial RV legacy implementation, checkRVVersion() is remoteSendEvent the remote RV to call it back with the RV version:
remoteSendEvent("remote-eval", "*", """
            require system; remoteSendEvent("remote-eval", "*", 
                "require sync; sync.checkRVVersion(\\"%s\\");" % system.getenv("TWK_APP_VERSION"));
            """);
However, the upcoming RV Commercial 2023.0 will contain the exact same sync.mu as the one in Open RV so they are going to be compatible.


rogernelson (approved):
Good work Bernie!

geffrak (approved):

PR #58, opened 2023-02-09
Allow building OpenRV as a subproject

rogernelson:
This branch allows building OpenRV as a subproject. 

The problem with using the `CMAKE_SOURCE_DIR` variable is that if you add OpenRV. as a subproject the `CMAKE_SOURCE_DIR` is the source dir of the parent project.  By using `PROJECT_SOURCE_DIR` we'll ensure all the paths are relative to the OpenRV root, whether or not it is a subproject.

Because the versioning was being added in one of the CMakeFiles in the module path, I had to move that into the root CMakeFile where the `project` call is done.


cmake/globals/rv_version.cmake
```diff
@@ -3,27 +3,10 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
-
 SET(RV_VERSION_YEAR
     "2023"
     CACHE STRING "RV's year of release."
```
geffrak, line 8: We could move all those variables to the main CMakeLists.txt. It contains one more version number and variables that are not related to the RV version... 

and also

```
string(TIMESTAMP RV_VERSION_YEAR "%Y")
SET(RV_VERSION_YEAR ${RV_VERSION} CACHE INTERNAL)
```

geffrak (approved):

cmake/globals/rv_version.cmake
rogernelson: Alright, will do.

bernie-laberge (approved):

Issue #63, opened 2023-02-17
Add libarg to repo


tjjackson:
I have been tracking down a number of libraries not called out in the CentOS7 build requirements, most have been simple enough to just install.  One that has tripped me up is there are a number of includes to "arg.h", which I was able to track down to Paul Heckbert's libarg written at Carnegie Mellon, with the latest release being from 1998:

https://www.cs.cmu.edu/afs/cs/Web/People/ph/859E/src/libarg/

As one might expect, trying to build C code written in the 1990's doesn't work out of the box with modern compilers.  I presume that Autodesk has an updated in-house version that works with gcc-9?  Would it be possible to add this to the repo?  Or is there some updated version of libarg elsewhere that we just need to install?

geffrak:
It is in the src/pub submodule. Have you pulled the code recursively with "git clone --recursive"? Otherwise you can "git submodule update --init --recursive" to pull the submodules.

tjjackson:
Well, I apparently was playing on "hard mode"!  Things seem to be working now that I cloned the other repos via the --recursive flag.

geffrak:
Closing the issue as the issue seems resolved. Please re-open if I'm wrong.

PR #84, opened 2023-02-17
Update documentation

geffrak:
### Linked issues

[Build size because missing release variant in the docs]( https://github.com/AcademySoftwareFoundation/OpenRV/discussions/62)
[Missing --prefix to the install step](https://github.com/AcademySoftwareFoundation/OpenRV/discussions/65)
[Missing --recursive flag while cloning](https://github.com/AcademySoftwareFoundation/OpenRV/issues/63)
### Summarize your change.

- Refresh the documentation to have all the full build steps in the main README.md file with the common build aliases equivalent.

- Moves doc into docs.

- Blind spots in the OpenRV documentation cause multiple issues. 
- The documentation root folder should be named `docs,` as requested by the documentation team for the future ReadTheDocs integration.

- [ ] Windows aliases
- [ ] MacOS aliases
- [ ] Linux aliases
- [ ] Documentation link changes

### Add a list of changes and note any that might need special attention during the review.

I updated the common build alias files to use `main_executable` instead of a mix of `RV` and `rv`. This needs to be tested on all OSes. 

I moved the doc folder to the docs folder. I think changed all occurrences to `doc`, but it's to be double-checked.

Main Readme Render: https://github.com/geffrak/OpenRV/tree/docs#readme
Windows setup doc Render: https://github.com/geffrak/OpenRV/blob/docs/docs/build_system/config_windows.md
macOS setup doc Render: https://github.com/geffrak/OpenRV/blob/docs/docs/build_system/config_macos.md
Linux setup doc Render: https://github.com/geffrak/OpenRV/blob/docs/docs/build_system/config_linux_centos7.md


mchesnay (commented):
Some comments, but most are on form, simplifying some sentences to clarify.

rvcmds.sh
```diff
@@ -20,21 +20,19 @@ fi
 QT_VERSION="${QT_VERSION:-5.15.2}"
 if [[ "$OSTYPE" == "linux"* ]]; then
   CMAKE_GENERATOR="${CMAKE_GENERATOR:-Ninja}"
-  QT_HOME="${QT_HOME:-$HOME/Qt${QT_VERSION}/${QT_VERSION}/gcc_64}"
-  RV_TARGET=rv
+  QT_HOME="${QT_HOME:-$HOME/Qt/${QT_VERSION}/gcc_64}"
```
geffrak, line 23: @rogernelson, I switched that to the default path that Qt provides.

README.md
```diff
@@ -2,86 +2,175 @@
...
-You invoke Open RV tests with the following command:
+
+### Build
+
+Invoke the previously specified generator tool using cmake to run the `build` step (recommended).
+
+#### Common build alias
+
+You can use the `rvbuild` common build alias to run the build step (`rvbuildd` to build in Debug).
+Use the `rvbuild` (the common build alias) to run the build step. You can also use `rvbuildd` to build in Debug.
```
mchesnay, line 110: Looks like lines 109 and 110 are duplicates. When re-reading it, I also realize that the following, a combination of your original string and my suggestion is better:
Use the `rvbuild` common build alias to run the build step. You can also use `rvbuildd` to build in Debug.

mchesnay (approved):

PR #105, opened 2023-03-29
100: Fixing MacOS build with CMake 3.26

### LINKED ISSUE #100: [Bug]: Unable to build OpenRV with cmake 3.26.0
### What happened?

New contributors cannot build OpenRV if their build toolchain is too recent. It seems like 
ADD_COMPILE_OPTIONS contains -std=${_clang_cxx_standard}, and it's illegal in C.

MacOS

Any

```shell
[...]
error: invalid argument '-std=c++17' not allowed with 'C'
[...]
```

### Environment variables

_No response_

### Extra information

Repro Step:
$ cmake --build _build/ --target resample --verbose 

### Code of Conduct

tbabiin:
### Linked issues

Fixes #100 

### Summarize your change.

Fixing MacOS build error in libresample when building with CMake 3.26.

The MacOS build is currently broken due to the faulty project being C language project and CMake Mac/darwin compiler options are mixed C/C++.

Building successfully on MacOS version Ventura 13.2.1.

I have removed standard libc++ flags as they are added by CMake correctly without adding them to the C compiler.


rogernelson (approved):
Thanks Inouk!

bernie-laberge (approved):

cmake/defaults/cxx_clang_defaults.cmake
```diff
@@ -30,18 +30,15 @@ ELSE()
...
-)
+set(CMAKE_C_FLAGS "${CMAKE_C_FLAGS} \
+  ${_verbose_invocation} \
+  -Wall \
+  -Wnonportable-include-path \
+  -msse \
+  -msse2 \
+  -msse3 \
+  -mmmx")
+set(CMAKE_CXX_FLAGS "${CMAKE_CXX_FLAGS} -stdlib=libc++ -std=${_clang_cxx_standard}")
```
rogernelson, line 41: As discussed this would probably work best here
```suggestion
 ADD_COMPILE_OPTIONS(
  ${_verbose_invocation}
  -Wall
  -Wnonportable-include-path
  -stdlib=libc++
  # not sure we need to actually set '-std' I would assume CMake does it for us.
  -std=${_clang_cxx_standard}
  -msse
  -msse2
  -msse3
  -mmmx
)
```

Then we could add the following to where we set the CXX_STANDARD variable to get the stdlib flag:
```
set(CMAKE_CXX_EXTENSIONS OFF)
```

geffrak: Should we set (CMAKE_CXX_EXTENSIONS OFF) outside of that file so it applies to all 3 OSes/Compilers?

tbabiin: @geffrak That's what I was tested with yesterday. Moreover, I'm retesting this morning without any CMAKE_CXX_FLAGS: -stdlib is supposed to be added automatically. My only concern is that I cannot test what happens with "CMAKE_CXX_EXTENSIONS OFF" on Linux and Windows.

rogernelson: I think we only want it on Mac, since Mac no longer supports libstdc++ in recent versions. On other OSes its perfectly fine to use it and its the default.  

That said, I wonder if CMake is smart enough to figure that out on its own without us having to explicitly tell it not to use libstdc++ on Mac.

tbabiin: The STD flag on MacOS, Blue Ocean console is new with this PR and Green is old without C FLAGS and CXX with stdlib and duplicating stdlib=c++17

New_vs_old_STD_flags

geffrak (approved):

cmake/defaults/cxx_clang_defaults.cmake
rogernelson: Since it doesn't appear to be passing the stdlib flag at all in the new code, we can probably just remove the `CMAKE_CXX_EXTENSIONS` completely then, which if its the case will just use the platform default.  On Mac the platform default was changed to libc++ as of 10.9, so as long as we don't need to support anything older than than that, we're fine not manually specifying the flag.

My guess is we were probably using c++11 (which is only supported in libc++ on Mac) on Mac before 10.9 and that's why we had to manually specify it then.

tbabiin: By removing CXX_EXTENSIONS (which sets the default value for "CXX_EXTENSIONS" in CMAKE), gnu++17 makes a comeback:
CMAKE_CXX_EXT_ON

Is this what we want? Feels odd to get GNU on Clang.

PR #110, opened 2023-04-11
Fix PySide2 build break with recent macOS/Xcode.app

geffrak:
### Summarize your change.

Patch the PySide2 build to not use libXSLT.

Since the last update of MacOS (or Xcode.app?), we can't build PySide2 anymore. The issue seems caused by the way PySide2 finds and uses libXSLT. It's mixing multiple versions of MacOSX.sdk.

MacOS 13.3, Xcode 14.3.

### DIFF
--- a/src/build/make_pyside.py
+++ b/src/build/make_pyside.py
@@ -125,6 +125,23 @@ def prepare() -> None:
     print(f"Installing numpy with {install_numpy_args}")
     subprocess.run(install_numpy_args).check_returncode()
 
+    cmakelist_path = os.path.join(
+        SOURCE_DIR, "sources", "shiboken2", "ApiExtractor", "CMakeLists.txt"
+    )
+    old_cmakelist_path = os.path.join(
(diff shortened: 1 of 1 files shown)


src/build/make_pyside.py
```diff
@@ -176,6 +193,7 @@ def build() -> None:
         f"--openssl={os.path.join(OPENSSL_OUTPUT_DIR, 'bin')}",
         "--verbose-build",
         f"--parallel={os.cpu_count() or 1}",
+        "--skip-docs",
```
geffrak, line 196: The docs were already not building, but explicitly telling PySide2 not to build them.

bernie-laberge (approved):
Nice Kerby. 
Thanks
As we previously discussed : with the aim of limiting that kind of suffering, we might want to seriously consider using artifacts for the Open RV dependencies.

rogernelson (approved):

PR #117, opened 2023-04-19
[116: Fix segfault in httpGet command]

mattdaw:
### Summarize your change.

Fix out of bounds memory access in httpGet command

Segfault when testing httpGet in custom package.

macOS 13.2.1

Test script below no longer crashes, prints contents of github.com to console:

`RV.app/Contents/MacOS/RV -pyeval "from rv import commands; onReplyEvent = lambda event: print(event.contents()); commands.bind('default','global','on-reply-event', onReplyEvent, 'desc'); commands.httpGet('https://github.com', [], 'on-reply-event');"`

Parameter index corrected in MuUICommands.cpp

### DIFF
--- a/src/lib/app/RvCommon/MuUICommands.cpp
+++ b/src/lib/app/RvCommon/MuUICommands.cpp
@@ -1740,7 +1740,7 @@ NODE_IMPLEMENTATION(httpGet, void)
     StringType::String* authEvent  = NODE_ARG_OBJECT(3, StringType::String);
     StringType::String* progEvent  = NODE_ARG_OBJECT(4, StringType::String);
     bool                ignore     = NODE_ARG(5, bool);
-    bool                urlIsEncoded = NODE_ARG(7, bool);
+    bool                urlIsEncoded = NODE_ARG(6, bool);
 
     RvWebManager::HeaderList headers;
 


bernie-laberge (approved):
What a great catch ! 
Thank you Matt !

geffrak (approved):

src/lib/app/RvCommon/MuUICommands.cpp
```diff
@@ -1740,7 +1740,7 @@ NODE_IMPLEMENTATION(httpGet, void)
     StringType::String* authEvent  = NODE_ARG_OBJECT(3, StringType::String);
     StringType::String* progEvent  = NODE_ARG_OBJECT(4, StringType::String);
     bool                ignore     = NODE_ARG(5, bool);
-    bool                urlIsEncoded = NODE_ARG(7, bool);
+    bool                urlIsEncoded = NODE_ARG(6, bool);
```
rogernelson, line 1743: Nice one! I guess someone removed a parameter and forgot to adjust...

rogernelson (approved):

src/lib/app/RvCommon/MuUICommands.cpp
mattdaw: Yeah, may be a copy/paste error from httpPost which is in active use

mattdaw:
Thanks @geffrak :-)

PR #128, opened 2023-04-28
Rename the MMR switch groups to something more meaningful

bernie-laberge:
Rename the MMR switch groups to something more meaningful [SG-28740]

#### Problem: 
With the new Multiple Media Representations feature, the ui names of the source nodes and switch nodes were not really meaningful. 

#### Solution: 
In this commit, the media representation name is used as the ui name by default.

The MMR switch groups and sources displayed in the Session Manager now use names that are now more meaningful. [SG-28740]

Note that this fix originates from the RV 2023 commercial release.

(https://community.shotgridsoftware.com/t/rv-2023-0-0-release-is-available/17227)

Signed-off-by: Bernard Laberge <bernard.laberge@autodesk.com>

### DIFF
--- a/src/lib/app/RvApp/CommandsModule.cpp
+++ b/src/lib/app/RvApp/CommandsModule.cpp
@@ -1516,6 +1516,46 @@ NODE_IMPLEMENTATION(sourceMediaReps, Pointer)
     NODE_RETURN(names);
 }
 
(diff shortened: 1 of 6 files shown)


tbabiin (approved):
Looks good 👍

PR #132, opened 2023-04-28
Add property to sequence and stack to disable reversed blending mode

bernie-laberge:
Add property to sequence and stack to disable reversed blending mode

### Summarize your change.

To support per-segment blending modes, the SequenceIPNode node was enhanced with a property-container to 
handle per-input blending modes. 

Since ImageRenderer::renderAllChildren did not properly support multiple 
blend modes, I had to add a flag to disable reverse-order blending. 
Reverse-order blending is a mechanism that seemed in place for shader-based 
compositing (currently disabled) and that sometimes inverses the order in 
which inputs are composited. This means that in RV the first input of a 
LayoutGroup becomes the top layer (lastly blended). This also has the side 
effect of forcing a "Replace" blend mode on the first drawn input. I would 
have prefered to completely remove this mode, but this would have required a 
lot of backward-compat code so I simply added a flag in IPImage to disable it 
when required. For now, both the SequenceIPNode and the StackIPNode can disable 
it (they each own a property for that). RV should remain untouched. 

files modifications 
------------------- 

-Both the StackIPNode and the SequenceIPNodes must be configured to yield 
IPImage nodes that don't support the reverse-order blending mode. 

src/lib/ip/IPBaseNodes/SequenceIPNode.cpp 
src/lib/ip/IPBaseNodes/SequenceIPNode.h 

-Added per-input blending modes. 
-Added a flag to toggle support of reverse-order blending. 

src/lib/ip/IPBaseNodes/StackIPNode.cpp 
src/lib/ip/IPBaseNodes/StackIPNode.h 
src/lib/ip/IPCore/IPImage.cpp 
src/lib/ip/IPCore/IPImage.h 

-Refactored a piece of code to convert blending mode names into their 
numerical values. 
-Added the new supportReversedOrderBlending flag in IPImage to force-disable 
the reverse-order blending. 

src/lib/ip/IPCore/ImageRenderer.cpp 

When importing a multi-track OTIO in RV, tracks are composited in the reverse order (bottom track on top instead of top track on top).
To support per-segment blending modes, the SequenceIPNode node was enhanced with a property-container to 
handle per-input blending modes. 

Note that this fix originates from the RV 2023 commercial release.

(https://community.shotgridsoftware.com/t/rv-2023-0-0-release-is-available/17227)

Signed-off-by: Roger Nelson <roger.nelson@autodesk.com>

### DIFF
--- a/src/lib/ip/IPCore/IPImage.cpp
+++ b/src/lib/ip/IPCore/IPImage.cpp
@@ -38,38 +38,39 @@ const char* IPImage::waveformTagValue() { return "_waveform"; }
 void
 IPImage::init()
 {
-    fb              = 0;
-    noIntermediate  = false;
-    next            = 0;
-    children        = 0;
-    info            = 0;
-    dataType        = NoInternalDataType;
-    pixelAspect     = 0.0;      // ignore if 0
-    initPixelAspect = 0.0;
-    minMax          = false;
-    blendMode       = UnspecifiedBlendMode;
-    renderType      = BlendRenderType;
-    ignore          = false;
-    missing         = false;
-    invalid         = false;
-    useBackground   = false;
-    destination     = CurrentFrameBuffer;
-    shaderExpr      = 0;
-    mergeExpr       = 0;
-    branchCount     = 0;
-    m_coordID       = size_t(-1);
-    device          = 0;
-    unpremulted     = false;
-    m_renderIDHash  = 0;
-    imageNum        = -1;
-    samplerType     = Rect2DSampler;
-    hashCount       = 0;
-    isHistogram     = false;
-    isCropped       = false;
-    cropStartX      = 0;
-    cropStartY      = 0;
-    cropEndX        = 0;
-    cropEndY        = 0;
+    fb                           = 0;
+    noIntermediate               = false;
+    next                         = 0;
+    children                     = 0;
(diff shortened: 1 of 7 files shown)


geffrak (approved):

rogernelson (approved):

src/lib/ip/IPCore/IPImage.cpp
```diff
@@ -38,38 +38,38 @@ const char* IPImage::waveformTagValue() { return "_waveform"; }
...
 IPImage::init()
 {
-    fb              = 0;
-    noIntermediate  = false;
-    next            = 0;
-    children        = 0;
-    info            = 0;
-    dataType        = NoInternalDataType;
-    pixelAspect     = 0.0;      // ignore if 0
-    initPixelAspect = 0.0;
```
rogernelson, line 48: Did initPixelAspect get dropped accientally here?

rogernelson (commented):
Waiting to approve until I know if initPixelAspect drop was intentional.

src/lib/ip/IPCore/IPImage.cpp
bernie-laberge: As discussed, although it was removed in the commercial RV commit, I believe we should restore it. 
Good catch !

PR #140, opened 2023-05-03
Rocky 9 Build Guide

rogernelson:
This provides a build guide for Rocky 9.  It's similar to the CentOS build guide, but Rocky9 is a lot more minimalistic out-of-the-box, so a number of build requirements that are pre-installed on CentOS are not on Rocky9.  Thus using the CentOS build guide is not ideal since those missing packages must be manually tracked down.  This new guide is intended to make life easier for Rocky/RHEL9 users. 

I also fixed the rvcmds to base the script home on the shell scripting language instead of the OS, since the two were being conflated. It's the shell version and not the OS that determines the right command.

Tested on a fresh install of Rocky 9.1.

### DIFF
--- a/docs/build_system/config_linux_rocky9.md
+++ b/docs/build_system/config_linux_rocky9.md
@@ -1,0 +1,69 @@
+# Building Open RV on CentOS 7
+
+## Summary
+
+1. [Install Basics](#install-basics)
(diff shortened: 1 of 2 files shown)


geffrak (approved):

bernie-laberge (approved):
Superb ! Thank you Roger

PR #142, opened 2023-05-05
Fix Mac install

rogernelson:
This branch fixes the Mac installation.

The problem is the CMake variable `PROJECT_SOURCE_DIR` is not available on the install step (only configure and build).  Since no variables involving the source dir are present at installation time, the only suitable replacement I could find was to use a path relative to the current directory `CMAKE_CURRENT_LIST_DIR`.  

Tested on Mac 13.3.1

### DIFF
--- a/cmake/install/pre_install_darwin.cmake
+++ b/cmake/install/pre_install_darwin.cmake
@@ -83,6 +83,6 @@ FUNCTION(after_copy_platform FILE_PATH)
     RETURN()
   ENDIF()
 
-  EXECUTE_PROCESS(COMMAND python3 ${PROJECT_SOURCE_DIR}/src/build/remove_absolute_rpath.py --target ${FILE_PATH} COMMAND_ERROR_IS_FATAL ANY)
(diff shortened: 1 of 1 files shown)


geffrak (approved):
Good catch. I wonder how come it was working before. Is it something that went away with a CMake update? 

You could set the PROJECT_SOURCE_DIR the same way we expose all the RV vars here: https://github.com/AcademySoftwareFoundation/OpenRV/blob/main/cmake/install/CMakeLists.txt#L19-L20

rogernelson:
> Good catch. I wonder how come it was working before. Is it something that went away with a CMake update?
> 
> You could set the PROJECT_SOURCE_DIR the same way we expose all the RV vars here: https://github.com/AcademySoftwareFoundation/OpenRV/blob/main/cmake/install/CMakeLists.txt#L19-L20

No, I previously broke it.

I replaced CMAKE_SOURCE_DIR with CMAKE_PROJECT_DIR  ... source dir doesn't work when you use it as a submodule since it takes the source dir of the parent -- project dir works as a replacement because it takes the directory where the last call to project() was made.  However, project is not called at install time, so it doesn't work on the install step.

bernie-laberge (approved):

PR #148, opened 2023-05-19
Use bullet points for all contributors

rogernelson:
Use bullet points for all contributors.

The original contributors were not listed in bullet point form like the OpenRV contributors.  This makes them both use bullet points.


src/pub
geffrak, line 1: Not sure you want to change src/pub

rogernelson: Me neither.  I just uninitialized and reinitialized my submodules and it still shows that.

geffrak (approved):

bernie-laberge (approved):

PR #150, opened 2023-05-23
OpenSSL: Re-enable assembly language routines

geffrak:
### Linked issues

https://github.com/openssl/openssl/issues/18225

### Summarize your change.

Removes the `no-asm` flag to OpenSSL's build to build with the assembly language routines.

When compiling OpenSSL using Xcode 11, disabling the assembly language routines breaks the TLS 1.2 support.

Tested on MacOS Ventura 13.3.1 with an Apple Silicon SoC.

### Add a list of changes and note any needing special attention during the review.

The change will get applied to all 3 OSes to have the same behaviour on all OSes, but I didn't test on all 3 OSes yet. The additional tests will get executed once we pull that commit into ShotGrid RV's internal CI. 

#### Before (Failed to fetch the site info):
image

#### After (Site info are fetched, failing because of dummy credentials): 
image

### DIFF
--- a/src/build/make_openssl.py
+++ b/src/build/make_openssl.py
@@ -24,6 +24,7 @@ OUTPUT_DIR = ""
 ARCH = ""
 PERL_ROOT = ""
 
+
 def get_openssl_args(root) -> List[str]:
     """
     Return the path to the openssl binary given a package root.
@@ -182,7 +183,6 @@ def configure() -> None:
     else:
         configure_args.append("linux-x86_64")
 
-    configure_args.append("no-asm")
     configure_args.append(f"--prefix={OUTPUT_DIR}")
     configure_args.append(f"--openssldir={OUTPUT_DIR}")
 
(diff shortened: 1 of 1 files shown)


src/build/make_openssl.py
```diff
@@ -182,7 +183,6 @@ def configure() -> None:
     else:
         configure_args.append("linux-x86_64")
 
-    configure_args.append("no-asm")
```
rogernelson, line 185: Does OpenSSL support arm64 asm? Just wondering if this might cause an issue with eventual native M1 support.

rogernelson (approved):
Wow, nice catch!

src/build/make_openssl.py
geffrak: I don't know. I assume that it will "no-asm" automatically in that case.

bernie-laberge (approved):

PR #151, opened 2023-05-24
Added variable to fix the annotations panel closure bug

shannonyearwood:
Added the _hideDrawPane variable to keep track of when to hide the draw pane. This will allow the closure of the annotations panel when "Select/Scrub Time" Tool is selected.

Variable has 3 stages: 
0 - Initial value/Nothing selected
1 - When the "Select/Scrub Time" Tool is selected
2 - When the "Select/Scrub Time" Tool is selected and the 'x' icon is selected

### DIFF
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -1685,6 +1685,7 @@ class: AnnotateMinorMode : MinorMode
         _syncAutoStart     = false;
         _cursorChar        = char(0x1);
         _autoSave          = true;
+        _hideDrawPane      = 0;
 
         let m = mainWindowWidget(),
             g = QActionGroup(m);
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -166,6 +166,8 @@ class: AnnotateMinorMode : MinorMode
 
     char              _cursorChar;
 
+    int              _hideDrawPane;
+
     \: colorToArray (float[]; Color c) { float[] {c.x, c.y, c.z, c.w}; }
     \: arrayToColor (Color; float[] a) { Color(a[0], a[1], a[2], a[3]); }
     method: encodedName (string; string name) { regex.replace("\.", _user, "+"); }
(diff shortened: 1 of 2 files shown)


geffrak (approved):

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -166,6 +166,8 @@ class: AnnotateMinorMode : MinorMode
 
     char              _cursorChar;
 
+    int              _hideDrawPane;
```
geffrak, line 169: Given the current usage, it could be a bool.

bernie-laberge, line 169: Best to always initialize members in the constructor.
Around line 1687.

shannonyearwood: Fixed

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1529,12 +1531,14 @@ class: AnnotateMinorMode : MinorMode
         if (_currentDrawMode eq _selectDrawMode)
         {
             _leaveUiVisible = true;
```
bernie-laberge, line 1533: Note sure I understand the logic of _leaveUiVisible to be honest.
I checked and this is the original code from 14 years ago.
When do we want to leave the UI visible ?

shannonyearwood: I actually do not think that variable is necessary anymore given the changes. 
Previously, this was used to keep the annotation panel open even when the "Select/Scrub Time" Tool is selected as otherwise it would automatically close.

The new variable now keeps track of when to keep the panel open though so I can remove this

src/plugins/rv-packages/annotate/annotate_mode.mu
geffrak: Oh right. I didn't get that it's a `> 1` and not  `>= 1`.

src/plugins/rv-packages/annotate/annotate_mode.mu
shannonyearwood: Fixed

geffrak:
Please bump the version of the package before merging :) A minor bump is enough.

bernie-laberge (approved):
Great Job ! Thanks !

PR #156, opened 2023-06-02
Added RHEL vars for Redhat distros in CMake

ThomBern:
### Summarize your change.
Added the possibility to generate a variable in CMake (on Linux only) that stores the RHEL major release, to use in if statements in case of problems between distro versions.

When running rvbootstrap and rvbuild, some libraries would be 'non-existant', even though they did exist; the problem was in the library suffix which was different with what was needed (eg: .1.so instead of .so.1). This problem would cause the commands to constantly try to rebuild the 'missing' libraries, while there were no changes brought to the files.

Rocky 9.1
The problem occurred with libgc (in gc.cmake) and libcrypto (in openssl.cmake)
- gc.cmake
- openssl.cmake
- rv_targets.cmake
Both cmake files now have an if statement to check their RHEL value, and change the suffix order accordingly.


cmake/defaults/rv_targets.cmake
```diff
@@ -98,6 +98,17 @@ ELSEIF(UNIX)
     -DPLATFORM_OPENGL=1
   )
 
+  EXECUTE_PROCESS(COMMAND cat /etc/redhat-release OUTPUT_VARIABLE RHEL_VERBOSE ERROR_QUIET)
+
+  IF(RHEL_VERBOSE) 
+    MESSAGE(STATUS "Redhat Entreprise Linux version: ${RHEL_VERBOSE}")
+    string(REGEX MATCH "([0-9]+)" RHEL_VERSION_MAJOR "${RHEL_VERBOSE}")
```
geffrak, line 105: Please run the pre-commit hooks. 
`brew install pre-commit`, `pre-commit install`, `git commit --amend -s`

geffrak (approved):

cmake/dependencies/gc.cmake
```diff
@@ -83,6 +83,20 @@ ELSEIF(RV_TARGET_LINUX)
...
       ${_lib_dir}/${CMAKE_SHARED_LIBRARY_PREFIX}cord${CMAKE_SHARED_LIBRARY_SUFFIX}
   )
+  IF(RV_TARGET_IS_RHEL9)
+    SET(_gc_lib_name
+        ${CMAKE_SHARED_LIBRARY_PREFIX}gc${CMAKE_SHARED_LIBRARY_SUFFIX}.1
+    )
+    SET(_cord_lib_name
+        ${CMAKE_SHARED_LIBRARY_PREFIX}cord${CMAKE_SHARED_LIBRARY_SUFFIX}.1
+    )
+    SET(_gc_lib
```
tbabiin, line 93: SET(_gc_lib
        ${_lib_dir}/${CMAKE_SHARED_LIBRARY_PREFIX}gc${CMAKE_SHARED_LIBRARY_SUFFIX}.1
    )
=>
    SET(_gc_lib
        ${_lib_dir}/${(_gc_lib_name}
    )

---- 
Can be done for other libs in gc and openssl

tbabiin:
I have the same situation on Rocky8 hence we should add it to the condition.
Screenshot from 2023-06-05 12-48-32
Screenshot from 2023-06-05 12-49-51

geffrak:
I would do the RHEL8/RHEL9 way by default and RHEL7 as a special case.

PR #157, opened 2023-06-02
Babiin/ocio v2 2 fix ocio freezes and dbl deletes

tbabiin:
### Summarize your change.

Fixing freezes by unlock Mutex when a exception happens
Using QMutex to easily unlock when an exception happen by using Scope with QMutexLocker
Fixing double/multi deletes of Shader::Function when a Config is not found.

Fixing major issues using OCIO

Tested exception robustness when OCIO name isn't the same as RV
Tested many OCIO CS configs to test underscore fix

    Removing duplicate function object addresses in retire vector before
    destructing objects
    Using QMutex instead of pthread_mutex to take advantage of obj scope
    versus Exceptions leaking from Shader::Function ctor
    Fixing unfound OCIO Colour Spaces by removing consecutive underscores as
    OCIO removes them in their code


tbabiin:
By target destination

PR #161, opened 2023-06-05
Automatically retrigger rvpkg on package changes

geffrak:
### Summarize your change.

Adds a dependency to the RV Packages to the step that installs the package to improve the developer experience.

If someone is working on a package and modifying the files, re-building will build a new rvpkg and install it in the compiled RV.

The developer experience was bad. We had to install the package manually.

It's working on MacOS Ventura, but it's pretty platform-independent.


bernie-laberge (approved):
Beautiful ! Thanks !

rogernelson (approved):

PR #163, opened 2023-06-05
Fix rvio Error encoding audio frame

bernie-laberge:
### Summarize your change.

Properly initialize the AVFrame associated with the audio to be encoded prior to calling the FFmpeg method : avcodec_fill_audio_frame()

While testing the rvio software implementation, we realize that the Open RV rvio (hardware) implementation had an issue with the audio encoding: it was returning the following error:
ERROR: Error encoding audio frame: Invalid argument

This error was NOT reproducible on the commercial version of RV.

After investigation, it was found that Open RV uses a more recent version of FFmpeg than the commercial version : 4.4.3 for Open RV, 4.2.1 for RV Commercial. The newer FFmpeg version requires the audio AVFrame passed to avcodec_fill_audio_frame() to be fully initialized. The older FFmpeg version worked without it.
I double checked that all avcodec_fill_audio_frame() calls are done with a fully initialized AVFrame.

Signed-off-by: Bernard Laberge [bernard.laberge@autodesk.com](mailto:bernard.laberge@autodesk.com)

### DIFF
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -4619,6 +4619,9 @@ MovieFFMpegWriter::fillAudio(Movie* inMovie, double overflow, bool lastPass)
     //
 
     track->audioFrame->nb_samples = nsamps;
+    track->audioFrame->format = audioFormat;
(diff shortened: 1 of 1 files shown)


geffrak (approved):

rogernelson (approved):
Nice find! You think they'd provide some kind of ctor/init method for this...

PR #164, opened 2023-06-05
Added RHEL vars for Redhat distros in CMake

ThomBern:
### Summarize your change.
Added the possibility to generate a variable in CMake (on Linux only) that stores the RHEL major release, to use in if statements in case of problems between distro versions.

When running rvbootstrap and rvbuild, some libraries would be 'non-existant', even though they did exist; the problem was in the library suffix which was different with what was needed (eg: .1.so instead of .so.1). This problem would cause the commands to constantly try to rebuild the 'missing' libraries, while there were no changes brought to the files.

Rocky 9.1
The problem occurred with libgc (in gc.cmake) and libcrypto (in openssl.cmake)
- gc.cmake
- openssl.cmake
Both cmake files now have an if statement to check their RHEL value, and change the suffix order accordingly.
- rv_targets.cmake


cmake/dependencies/gc.cmake
```diff
@@ -83,6 +83,20 @@ ELSEIF(RV_TARGET_LINUX)
   SET(_cord_lib
       ${_lib_dir}/${CMAKE_SHARED_LIBRARY_PREFIX}cord${CMAKE_SHARED_LIBRARY_SUFFIX}
   )
+  IF(RV_TARGET_IS_RHEL9)
```
geffrak, line 86: Can you do the other way around, and do this by default and the previous way for RHEL7?

geffrak (approved):

bernie-laberge (approved):

PR #166, opened 2023-06-06
120: Fix gltext memory leak (and other mem leaks)

### LINKED ISSUE #118: Typo in addSourcesVerbose
Simple one-character error; should be `string[]` vs `sring[]` 

https://github.com/AcademySoftwareFoundation/OpenRV/blob/41eccbe03bf2462b399fec6511be1d11de6c6327/src/lib/app/mu_rvui/commands.mud#L754

### LINKED ISSUE #120: [Bug]: gltext memory leak
Using gltext seems to cause a memory leak issue.  
I provided an example mode that creates a watermark using gltext that shows the issue, however it seems to also happen with the Color Inspecter tool shipped with Openrv (uses gltext).  Memory usage of RV will continue to increase causing RV to slow down and or eventually crash.

Repo:
copy provided ttf file to /var/tmp/
install provided RV package
Launch RV
Open provided image sequence 'smaple_image.1-10#.jpg'
Play 10 frames in loop playback
'top' cmd shows RES stable around 1121M
hit 'o' hot key to toggle 'memoryleak' mode
'top' cmd shows RES steadily increases over time.
[memory_leak_example.zip](https://github.com/AcademySoftwareFoundation/OpenRV/files/11324764/memory_leak_example.zip)

Centos7

Quadro P5000

_No response_

### Environment variables

_No response_

### Extra information

_No response_

### Code of Conduct

bernie-laberge:
### Linked issues
Fixes #120 
Fixes #118 

### Summarize your change.

Fixed the gltext memory leak by fixing the PyUnicode_AsUTF8String() which was used without its associated Py_XDECREF().

Note that this memory leak occurred not only in gltext but also whenever an Python RV package was calling an RV cmd with text parameters.

This fix addresses the following Open RV reported issue:
https://github.com/AcademySoftwareFoundation/OpenRV/issues/120

#### Problem:
Using gltext seems to cause a memory leak issue.
The reporter provided an example mode that creates a watermark using gltext that shows the issue, however it seems to also happen with the Color Inspecter tool shipped with Openrv (uses gltext). Memory usage of RV will continue to increase causing RV to slow down and or eventually crash.

#### Cause:
I was able to identify the source of the memory leak using XCode/instruments.
The leak is in the Python->Mu code (PyModule::py2mu()) which is reponsible of converting Python objects to Mu objects.
More specifically when converting a Python string to Mu.
PyUnicode_AsUTF8String() was used without its associated Py_XDECREF().

#### Solution:
Added Py_XDECREF().

#### Also in this commit:
A mechanics was put in place 4 years ago to prevent a playback during a review session from being interrupted by the garbage collector which was causing some playback hiccups ie skipped frames. We had decided to disable the garbage collector during the playback which resolved the Marvel’s issue. Now because of this mechanism, the Color Inspector tool (or any mu code being executed) will 'seem' to leak memory during playback but as soon as the playback is stopped then the garbage collector will do its job and makes all the ‘allocated memory’ during the playback, available again to the RV process. However, the side effect is that RV will have a bigger memory footprint : proportional to whatever memory is allocated in Mu scripts from the start of the playback until the playback is stopped.
As decided this morning at the RV dailies, I have added an environment variable to disable this mechanism if desired. The no garbage collector cleanup up during playback would be the default behaviour and would guarantee that no hiccups occur during playback.

To re-enable the garbage collector from operating during a playback:
export RV_DISABLE_GARBAGE_COLLECTION_DURING_PLAYBACK=0

#### Also in this commit:
Fixed a Typo in addSourcesVerbose where sring was used instead of string.
(Most probably my bad!)

#### Testing Protocole:

Please validate repro steps.
I recommend using the following command which uses 60 fps to speed up the memory leakage:
./rv.exe smptebars,start=1,end=100,fps=60.movieproc

First set the following env var to 0 to make sure that garbage collection takes place during playback
export RV_DISABLE_GARBAGE_COLLECTION_DURING_PLAYBACK=0

Then you can validate without the env var.

Signed-off-by: Bernard Laberge [bernard.laberge@autodesk.com](mailto:bernard.laberge@autodesk.com)

### DIFF
--- a/src/lib/ip/IPCore/Session.cpp
+++ b/src/lib/ip/IPCore/Session.cpp
@@ -70,6 +70,11 @@
 // Flush the audio cache when stopping the playback when ON (default=OFF)
 static ENVVAR_BOOL( evFlushAudioCacheWhenStoppingPlayback, "RV_FLUSH_AUDIO_CACHE_WHEN_STOPPING_PLAYBACK", false );
 
+// Disable automatic garbage collection during playback to make sure the
+// playback doesn't get interrupted which could cause skipped frames.
+// The garbage collection will resume as soon as the playback is stopped.
+static ENVVAR_BOOL( evDisableGarbageCollectionDuringPlayback, "RV_DISABLE_GARBAGE_COLLECTION_DURING_PLAYBACK", true );
+
 template <typename T>
 inline T mod( const T &a, const T &by )
 {
@@ -1880,9 +1885,12 @@ Session::play_v2(string &eventData)
     userGenericEvent("before-play-start", eventData);
     m_beforePlayStartSignal(eventData);
 
-    // Disable automatic garbage collection during playback to make sure it
-    // doesn't get interrupted.
-    Mu::GarbageCollector::disable();
+    // Disable automatic garbage collection during playback to make sure the
+    // playback doesn't get interrupted which could cause skipped frames.
+    if (evDisableGarbageCollectionDuringPlayback.getValue())
+    {
+        Mu::GarbageCollector::disable();
+    }
 
     m_syncLastTime = 0.0;
     m_timer.start();
@@ -1967,9 +1975,12 @@ Session::play_v1(string &eventData)
     userGenericEvent("before-play-start", eventData);
     m_beforePlayStartSignal(eventData);
 
-    // Disable automatic garbage collection during playback to make sure it
-    // doesn't get interrupted.
-    Mu::GarbageCollector::disable();
+    // Disable automatic garbage collection during playback to make sure the
+    // playback doesn't get interrupted which could cause skipped frames.
+    if (evDisableGarbageCollectionDuringPlayback.getValue())
+    {
+        Mu::GarbageCollector::disable();
+    }
 
     m_syncLastTime = 0.0;
     m_timer.start();
@@ -2093,7 +2104,10 @@ Session::stop(string eventData)
     m_playStopSignal(eventData);
 
     // Re-enable automatic garbage collection after playback.
-    Mu::GarbageCollector::enable();
+    if (evDisableGarbageCollectionDuringPlayback.getValue())
+    {
+        Mu::GarbageCollector::enable();
+    }
 
     if (debugPlayback)
     {
--- a/src/lib/mu/MuPy/PyModule.cpp
+++ b/src/lib/mu/MuPy/PyModule.cpp
@@ -529,8 +529,9 @@ namespace Mu
       if( PyUnicode_Check( pobj ) )
       {
         PyObject* utf8 = PyUnicode_AsUTF8String( pobj );
-        return Value(
-            Pointer( c->stringType()->allocate( PyBytes_AsString( utf8 ) ) ) );
+        Value ret( Pointer( c->stringType()->allocate( PyBytes_AsString( utf8 ) ) ) );
+        Py_XDECREF( utf8 );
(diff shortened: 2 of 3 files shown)


src/lib/ip/IPCore/Session.cpp
```diff
@@ -1880,9 +1885,12 @@ Session::play_v2(string &eventData)
     userGenericEvent("before-play-start", eventData);
     m_beforePlayStartSignal(eventData);
 
-    // Disable automatic garbage collection during playback to make sure it
-    // doesn't get interrupted.
-    Mu::GarbageCollector::disable();
+    // Disable automatic garbage collection during playback to make sure the
+    // playback doesn't get interrupted which could cause skipped frames.
+    if (evDisableGarbageCollectionDuringPlayback.getValue())
```
rogernelson, line 1890: Where does this one get re-enabled?

src/lib/mu/MuPy/PyModule.cpp
```diff
@@ -529,8 +529,9 @@ namespace Mu
       if( PyUnicode_Check( pobj ) )
       {
         PyObject* utf8 = PyUnicode_AsUTF8String( pobj );
-        return Value(
-            Pointer( c->stringType()->allocate( PyBytes_AsString( utf8 ) ) ) );
+        Value ret( Pointer( c->stringType()->allocate( PyBytes_AsString( utf8 ) ) ) );
+        Py_XDECREF( utf8 );
```
rogernelson, line 533: Nice catch!

src/lib/ip/IPCore/Session.cpp
rogernelson: Oh nevermind, I see now.  The start and stop are separate methods.

rogernelson (approved):

src/lib/ip/IPCore/Session.cpp
```diff
@@ -2093,7 +2104,10 @@ Session::stop(string eventData)
     m_playStopSignal(eventData);
 
     // Re-enable automatic garbage collection after playback.
-    Mu::GarbageCollector::enable();
+    if (evDisableGarbageCollectionDuringPlayback.getValue())
```
rogernelson, line 2107: If the envvar is true but GC was disabled for some other reason we would turn it back on here.  Do we want to set a flag if we disabled when we started playback?

rogernelson: Something like:

```
if (evDisableGarbageCollectionDuringPlayback.getValue()) {
   m_enableGC = Mu::GargabeCollector::isEnabled();
   Mu::GarbageColector::disable();
}
```

Then to re-enable
```
if (m_enableGC) {
   m_enableGC = false;
   Mu::GarbageColector::enable();
}
```

geffrak: It would avoid a situation when someone would try to make a plugin to control that.

bernie-laberge: Good point Roger !
I think we're good without any external handling because the Garbage Collector is ref counted:
https://github.com/ivmai/bdwgc/blob/5520e1c8379615bab14ac0ee0a8e8ff57ce586e2/include/gc/gc.h#L909

rogernelson: GC is actually smarter that I though.  The calls to enable/disable are refcounted.

Issue #162, opened 2023-06-05
[Feature]: Update OCIO to v2.2


tbabiin:
### What happened?

Update OCIO v2.2.1 and update any OCIO dependencies which OCIO needs.

### List all the operating systems versions where this is happening

Windows 10, Windows 11, CentOS 8, Rocky 8, Rocky 9, MacOS 13.2.1

### On what computer hardware is this happening?

All

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

rogernelson:
Feature branch created: https://github.com/AcademySoftwareFoundation/OpenRV/tree/feature/162/ocio-update-to-v2

dbr:
Hello! Just wondering what the status of this is - the linked PR above is merged, but it is in a feature branch, not the main one?

Having RV with OCIO v2 would be very useful, as it's the last application we are using that is still on v1

compaga:
Hi dbr, we are still stabilizing the branch. We currently have confirmed the OCIO v1 config file load and render fine on all three platforms but but we are finding issue with OCIO v2 config file. Specifically on Mac, there are shader error that we need to investigate. Since you seem very interested, if you want to build this branch and help testing it, your help would be more than welcome.

herronelou:
Hello.
It seems like the PR has now been merged in main, I was wondering if there are plans to modify the `ocio_source_setup.py` rv package.
OCIO v2, in particular the newer versions, introduced a lot of changes in regards to displays and views, as well as new logic for picking the correct colorspace for different sources via file rules, none of which seems to be taken into account in the OCIO source setup package.
RV in itself has always been a bit awkward to setup for proper OCIO management, particularly for multi-shot timelines (where we tend to need to load a different CDL per shot), so this might be an opportunity to make some improvements, especially since now both RV and OCIO are Autodesk products.

geffrak:
@herronelou, the first step was to update the OCIO library being used (keeping the same functionality). The following steps are about using it properly and improving RV with it.

BrianHanke:
Happy new year all! I wanted to check in on the status of OCIO. I haven't built RV from source in quite a while, but it sounds from the above discussion that OCIO 2 is at least partially implemented now?

compaga:
This feature is completed.

PR #172, opened 2023-06-06
Issue#162: Update OCIO to v2.2 on all supported platforms

tbabiin:
### Linked issues

Fixes SG-29427
Fixes SG-30792
Fixes SG-31071
Fixes SG-31074
Fixes SG-31094

### Summarize your change.

Updating OCIO to v2.2 and updating all OCIO's dependencies including OpenImageIO.

Updating OCIO v2.2 is a requested feature which will allow for enhanced Colour Spaces display.

Tested on Linux Rocky 8 and Rocky 9
Tested on MacOS 13.2.1
Tested on Windows 11.

Adding CMake macros to alleviate boilerplate code when creating a new CMake file for a new dependency
JPEG2000 is now handled by OpenImageIO
Updating many dependencies and/or they are now downloaded instead of using src/pub: TIFF, OpenJPEG, PNG, TurboJPEG, OCIO, OIIO, WEBP, OpenEXR, and more.
Adding New dependency only for Windows: VCPKG for OCIO.
Refactoring a some code section for TIFF due to the new version
Refactoring OCIOIPNode code for the new OCIO major version (GPU pipeline)

**IMPORTANT**: Pending changes: 
[ ] update src/pub submodule to remove old dependencies that are now downloaded.
[ ] Decide if we want more verbose build logs (same as before) or less. If we decide to be Verbose, implement it in a cross-platform way.
[x] update Notice.txt and **LICENSE** especially, but limited to, for the OCIO version. ([OCIO license paragraph](https://github.com/AcademySoftwareFoundation/OpenColorIO/blob/main/LICENSE))
[x] (Unnecessary, only new is Vcpkg and it is a MIT license) collect any new licenses' changes and transfer them to legal team. To fix before merging to `main`

 **Pending Testing**: the above bug reports for freezes haven't been validated yet.
[ ] **Test fail**: MacOS: Display LUT is rendered as full black. Cause: the shader compilation is failing on GLSL 1.2.


tbabiin:
❗ IMPORTANT ❗ OpenRV-pub branch to be enable to build **this branch** : tbabiin/issue-162/remove-packages-for-ocio-v2 (only on my OpenRV-pub fork as of 06/07/2023)

cmake/defaults/rv_options.cmake
```diff
@@ -11,6 +11,9 @@ OPTION(RV_SHOW_ALL_VARIABLES "Displays all build variables." ON)
 
 #
 # General build options
+
+OPTION(RV_USE_OCIO_YAML_CPP "Uses OpenColorIO's own copy of Yaml-Cpp." ON)  # IF OFF: the user must provide another yaml lib
```
geffrak, line 15: Why do we want to give an option?

src/lib/image/IOjp2/IOjp2.cpp
```diff
@@ -98,7 +99,9 @@ string
 IOjp2::about() const
 { 
     ostringstream str;
+#if 0
```
geffrak, line 102: Will you remove this #if 0 before merging?

src/lib/image/IOjp2/IOjp2.cpp
```diff
@@ -117,6 +120,7 @@ IOjp2::throwIfError(const std::string& filename) const
 void
 IOjp2::getImageInfo(const std::string& filename, FBInfo& fbi) const
 {
+#if 0
```
geffrak, line 123: Will you remove this #if 0 before merging?

src/lib/image/IOjp2/IOjp2/IOjp2.h
```diff
@@ -37,7 +37,8 @@ class IOjp2 : public FrameBufferIO
     mutable bool m_error;
 
 private:
-    mutable opj_event_mgr_t m_eventMgr;
+    
+    // mutable opj_event_mgr_t m_eventMgr;
```
geffrak, line 41: Intended?

src/lib/image/IOoiio/IOoiio.cpp
```diff
@@ -55,10 +55,10 @@ IOoiio::IOoiio() : FrameBufferIO("IOoiio", "n") // OIIO after any defaults of ou
     //  
     // addType("tif", "TIFF Image", rw, codecs);
```
geffrak, line 56: Why is the other ones commented?

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -218,9 +218,26 @@ shaderLegal(const string& s)
 
     transform(s.begin(), s.end(), ns.begin(), op_shaderLegal);
 
+    // for any double underscore, OCIO will replace it with just one; do the same so our names align. (See: OCIO:GPUShaderDesc::setFunctionName)
```
geffrak, line 221: Sadly, the function is not exposed. It looks like implementation details can change.

src/lib/image/IOoiio/IOoiio.cpp
tbabiin: It's a decision per file format. We made a decision for JP2000 but other file formats are outside if this PR work. If needed, we will revisit at another time.

src/lib/ip/OCIONodes/OCIOIPNode.cpp
tbabiin: I'm double-checking this. A getter function that contains the proper value might be exposed: getFunctionName in GPUShaderDesc.

tbabiin:
As per [Issue 162](https://github.com/AcademySoftwareFoundation/OpenRV/issues/162), I have moved the destination branch to [Feature branch 162](https://github.com/AcademySoftwareFoundation/OpenRV/tree/feature/162/ocio-update-to-v2).

cmake/defaults/rv_options.cmake
```diff
@@ -11,6 +11,10 @@ OPTION(RV_SHOW_ALL_VARIABLES "Displays all build variables." ON)
 
 #
 # General build options
+
+OPTION(RV_USE_OCIO_OPENEXR "Uses OpenColorIO's own copy of OpenEXR." OFF)   # Mac & Windows do not work with OCIO's copy. Moreover, next release OCIO will remove OpenExr on their side.
```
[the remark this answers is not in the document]
tbabiin: Option is removed.

cmake/defaults/rv_options.cmake
tbabiin: Not too sure (I'm not the original owner) I suspect it was an investigation tool to be able to switch between versions. However since we want to remove any libs in `src/pub` that we can, I'll remove the option.

cmake/dependencies/CMakeLists.txt
```diff
@@ -49,16 +56,35 @@ IF(RV_TARGET_DARWIN
 ENDIF()
 INCLUDE(imath)
 INCLUDE(zlib)
-INCLUDE(openexr)
 INCLUDE(ffmpeg)
 INCLUDE(doctest)
+INCLUDE(jpegturbo) # jpegturbo huge advantage over jpeg is that it does come with CMake support!
+INCLUDE(png)
+INCLUDE(tiff)  # depends on jpeg, jpegturbo
+ADD_SUBDIRECTORY(../../src/pub/lcms src/pub/lcms) # required by raw
```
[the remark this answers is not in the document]
tbabiin: I think this is better even though it's weird. The other way of loading 'pub' first in the normal way would be to pull 'src' fully first and this is a chicken & egg conundrum: src needs cmake/dependencies so we cannot load src first. Finally, src loads pub as how usually CMake works. But let me know if you prefer that I remove loading pub from src and then we can load src/pub before cmake/dependencies.

cmake/dependencies/ocio.cmake
```diff
@@ -0,0 +1,301 @@
...
+IF(RV_TARGET_WINDOWS)
+  SET(_byproducts "")   # Empty out the List as it has the wrong DLL name: it doesn't have the version suffix
+  SET(_ocio_win_sharedlibname OpenColorIO_2_2.dll)
+  SET(_ocio_win_sharedlib_path ${_bin_dir}/${_ocio_win_sharedlibname})
+  LIST(APPEND _byproducts ${_ocio_win_sharedlib_path})
+ENDIF()
+# 
+# Also add OpenExr libraries to _byproduct
+#
+SET(_iex_name "Iex-3_1")
```
rogernelson, line 46: I am confused by this part.  It seems to be so we can specify the OpenEXR libs as byproducts of the OCIO build, but we are also providing it with our build of the EXR libs.  So why is it generating its own?

cmake/dependencies/tiff.cmake
```diff
@@ -0,0 +1,207 @@
...
+  # prepare directories for copy. <cmake> -E copy expects the directories to exist otherwise copy won't work correctly. WARNING: CMAKE 3.26: if directories do
+  # not exist, <cmake> -E copy will copy the FILE as a FILE that looks like a directory; no real directory will be created
+  FILE(MAKE_DIRECTORY ${_lib_dir})
+  FILE(MAKE_DIRECTORY ${_bin_dir})
+  # Building with nmake isn't providing a nice install target, we need to do it manually
+  ADD_CUSTOM_COMMAND(
+    TARGET ${_target}
+    POST_BUILD
+    COMMENT "Installing ${_target}'s libs into ${RV_STAGE_LIB_DIR}"
+    COMMAND ${CMAKE_COMMAND} -E copy ${_base_dir}/src/libtiff/libtiff.lib ${_lib_dir}
```
rogernelson, line 84: Is it possible to get all these files with a glob command?  Seems we would be in trouble if we updated the version and the list of files changed.

cmake/packman/ocio_vcpkg.json
```diff
@@ -0,0 +1,7 @@
+{
+  "dependencies": [
+    "cxxopts",
```
rogernelson, line 3: Should we add cxxopts and and freeflut to the docs for windows (I think glew is already there).

cmake/dependencies/CMakeLists.txt
```diff
@@ -49,16 +56,35 @@ IF(RV_TARGET_DARWIN
...
+INCLUDE(openjpeg) # depends on zlib, tiff, png
+INCLUDE(webp)
+IF(NOT RV_USE_OCIO_OPENEXR)
+  INCLUDE(openexr)
+ENDIF()
+INCLUDE(ocio)
+IF(RV_USE_OCIO_OPENEXR)
+  INCLUDE(openexr)  # depends on ocio
+ENDIF()
+ADD_SUBDIRECTORY(../../src/pub/freetype src/pub/freetype) # required by oiio
```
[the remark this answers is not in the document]
bernie-laberge: It would be really nice to fix this chick and egg dependency issue.
I wonder if we could ADD_SUBDIRECTORY(src/pub) in the main CMakeLists.txt at the root:
```
ADD_SUBDIRECTORY(src/pub)
ADD_SUBDIRECTORY(cmake/dependencies)
```

(it might not be possible though so what we have now might be the best we can do until src/pub is completely eradicated)

bernie-laberge (approved):

cmake/dependencies/CMakeLists.txt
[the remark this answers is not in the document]
tbabiin: I don't think it's possible unless it's possible to do:
ADD_SUBDIRECTORY(src/pub)                 # New
ADD_SUBDIRECTORY(cmake/dependencies)

# RV's main code
ADD_SUBDIRECTORY(src)                       # Already exists and re-pulls src/pub. I'm pretty sure we will get duplicated targets/external projects.

cmake/dependencies/CMakeLists.txt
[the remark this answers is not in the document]
tbabiin: I'll put a note. Removing pub will take some time to do. As for now, Qt5 and OpenSSL are used by deps in src/pub. For Qt5, QtColorTriangle is using Qt5.

cmake/dependencies/CMakeLists.txt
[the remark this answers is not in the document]
tbabiin: As per above, Qt5 in cmake/dependencies needs to be loaded before src/pub/QtColorTriangle.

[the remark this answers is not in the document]
geffrak: We can move src/pub into pub, then load cmake/dependencies, pub, src, ...

cmake/defaults/rv_options.cmake
tbabiin: As I said, we are moving away from src/pub eventually. yaml_cpp was in src/pub and is used by `rvls`, OCIO pulls it for us and we reuse it. Eventually, I suspect that we might want our own copy and pass it to OCIO if possible. For now, I'll remove the option and keep OCIO's yaml_cpp and remove yaml_cpp from src/pub. Note(FYI): RvCommon uses the target 'yaml' for what seems to be preferences.

cmake/dependencies/tiff.cmake
tbabiin: We'll pull more than we need so to save space when installing RV we do not glob. I know that updating packages is not fun but it's better to investigate every package update and test.

cmake/packman/ocio_vcpkg.json
tbabiin: Those are the dependencies that VCPKG **pulls** for us. They aren't new dependencies.

cmake/dependencies/ocio.cmake
tbabiin: I'll remove them, they are indeed remnants of when we were trying to use OCIO's OpenExr. 1) OpenExr in OCIO is going away, 2) it doesn't build well on one of the platform, probably one of the reason for 1) and 3) we might not want RV to use the exact same version as OCIO eventually.

FYI I read on ByProducts for CMake and they are used as an indication for when to rebuild; it's not to tell a package _what_ to build; it is used in part for _when_ to build.

cmake/dependencies/tiff.cmake
tbabiin: It turns out that we cannot GLOB or use Wildcards for POST_BUILD. For POST_BUILD, you have to have a way to know the files in advance which defeats the purpose of Glob'ing for new files in the next update. We could eventually refactor it for most of the files: the DLL and LIB and all EXEs, we know the names ahead of time so we could create a list and loop and call add_custom_command for each: not a big gain however.

tbabiin: All EXE commands are now removed and we just copy the folder and remove a few files. All .C and .OBJ files are excluded via pre-install.

cmake/install/pre_install_windows.cmake
```diff
@@ -15,6 +15,16 @@ FUNCTION(before_copy_platform FILE_PATH RET_VAL)
     ENDIF()
   ENDIF()
 
+  # For <TIFF>/install/bin
+  IF(FILE_PATH MATCHES "\\.obj$"
```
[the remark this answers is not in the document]
tbabiin: Thanks, moving to the main pre_install.cmake and retesting.

PR #179, opened 2023-06-16
[BUG] Fix the color inspector final color mode

ThomBern:
### Summarize your change.
The color inspector will now receive a 'graph-state-change' event whenever the pixel color mode changes (the event already existed, it was just about binding it). We use that event to trigger the correct code inside the inspector.mu to update the inspector.

I also found another bug while working on this one, about the average color pixel; instead of taking the pixel color from the current mode, it would take the source pixel color no matter what.

Also fixed some janky inspector resetting event.
When in Final Color mode in the Color Inspector, a change in the graph would not always update the inspector. You'd need to move the inspector manually to update it.
Mac 
- inspector.mu

#### Repro steps
1. Download [color_inspector_bug.rv.zip](https://github.com/AcademySoftwareFoundation/OpenRV/files/11772217/color_inspector_bug.rv.zip)
2. Load it up in RV of your choice.
3. Press F5 to bring up color inspector and in the little drop-down triangle change menu to Final Rendered Color.
4. Go to View -> Rec709.
5. Note that inspector doesn't change.
6. Jiggle inspector, it should update.

### DIFF
--- a/src/lib/app/mu_rvui/inspector.mu
+++ b/src/lib/app/mu_rvui/inspector.mu
@@ -32,7 +32,10 @@ class: Inspector : Widget
     bool            _xFlipped;
     bool            _yFlipped;
     bool            _inputsChanged;
+    bool            _graphStateChanged;
     bool            _glReadPending;
+    bool            _updateAverageColor;
+    bool            _skipNextSampleAtPoint;
     bool            _nearTriangle;
     int             _widW;
 
@@ -111,9 +114,13 @@ class: Inspector : Widget
         if (_showFBColor)
         {
             _glReadPending = true;
-            //  float[] pixels;
-            //  glReadPixels(ip.x, ip.y, 1, 1, GL_RGBA, pixels);
-            //  _fbColor = Color(pixels[0], pixels[1], pixels[2], pixels[3]);
+            _averageColor    = Color(0, 0, 0);
+            _colorSamples    = 0;
+        }
+        else
+        {
+            _averageColor    = sp.v;
+            _colorSamples    = 1;
         }
 
         _colorSampling   = true;
@@ -122,8 +129,6 @@ class: Inspector : Widget
         _colorPointEvent = ip;
         _colorPointImage = eventToImageSpace(sName, ip);
         _colorSourceInfo = pinfo;
-        _averageColor    = sp.v;
-        _colorSamples    = 1;
         _colorFrame      = frame();
         redraw();
     }
@@ -144,9 +149,13 @@ class: Inspector : Widget
         if (_showFBColor)
         {
             _glReadPending = true;
-            //  float[] pixels;
-            //  glReadPixels(ip.x, ip.y, 1, 1, GL_RGBA, pixels);
-            //  _fbColor = Color(pixels[0], pixels[1], pixels[2], pixels[3]);
+            _averageColor    = Color(0, 0, 0);
+            _colorSamples    = 0;
+        }
+        else
+        {
+            _averageColor    = sp.v;
+            _colorSamples    = 1;
         }
 
         _colorSampling   = true;
@@ -155,14 +164,28 @@ class: Inspector : Widget
         _colorPointEvent = ip;
         _colorPointImage = eventToImageSpace(sName, ip);
         _colorSourceInfo = pinfo;
-        _averageColor    = sp.v;
-        _colorSamples    = 1;
         _colorFrame      = frame();
+
+        _skipNextSampleAtPoint = true;
+
         redraw();
     }
 
     \: sampleAtPoint (void; Inspector this, Event event)
     {
+        // Whenever we shift-click to reset the _averageColor value, it calls 'pointer-1--shift--push' then
+        // 'pointer-1--shift--drag'. This behaviour makes it so when we reset, we'd get 2 samples inside the
+        // _averageColor var, which is not what we want, because we're resetting that value! 
+        // (it should be either 0 or 1, for the value under the current position). 
+        // This if statement corrects that behaviour, by making sure that we skip the first time we enter
+        // sampleAtPoint() because beginSampling() was called at the same time.
+
+        if (_skipNextSampleAtPoint)
+        {
+            _skipNextSampleAtPoint = false;
+            return;
+        }
+
         State state = data();
         recordPixelInfo(event);
         if (state.pixelInfo eq nil || state.pixelInfo.empty()) return;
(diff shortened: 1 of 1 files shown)


src/lib/app/mu_rvui/inspector.mu
```diff
@@ -252,7 +250,9 @@ class: Inspector : Widget
 
     \: optFinalColor (void; Inspector i, bool val, Event event)
     {
+        //TODO: Reset average color value? Maybe?
```
bernie-laberge, line 253: Yes it makes sense: please reset the average color value when switching mode.

src/lib/app/mu_rvui/inspector.mu
```diff
@@ -253,6 +251,8 @@ class: Inspector : Widget
     \: optFinalColor (void; Inspector i, bool val, Event event)
     {
         i._showFBColor = val;
+        i._glReadPending = true;
```
bernie-laberge, line 254: How about simply resetting the average like this:
i._averageColor=0;
i._colorSamples=0;

src/lib/app/mu_rvui/inspector.mu
```diff
@@ -511,19 +520,28 @@ class: Inspector : Widget
...
             {
                 _currentColor  = sp.v;
                 c = _currentColor;
+                // If we reset the average color (w/ shift+click) or if we initialize the inspector with Final Color mode toggled on
+                if (_colorSamples == 1 && _glReadPending) {
+                    _averageColor = _fbColor;
+                }
             }
 
             if (!_glReadPending)
```
bernie-laberge, line 547: I would remove this if(!_glReadPending) and always execute the content of the if

src/lib/app/mu_rvui/inspector.mu
```diff
@@ -532,6 +550,7 @@ class: Inspector : Widget
                 _colorSamples += 1;
                 _colorFrame    = frame();
                 _inputsChanged = false;
+                _graphStateChanged = false;
```
bernie-laberge, line 553: This way, if the _averageColor was reset earlier then the colorSamples ==0 and it initializes the _averageColor.

src/lib/app/mu_rvui/inspector.mu
```diff
@@ -511,19 +520,28 @@ class: Inspector : Widget
...
+                if (_colorSamples == 1 && _glReadPending) {
+                    _averageColor = _fbColor;
+                }
             }
             else
             {
                 _currentColor  = sp.v;
                 c = _currentColor;
+                // If we reset the average color (w/ shift+click) or if we initialize the inspector with Final Color mode toggled on
+                if (_colorSamples == 1 && _glReadPending) {
```
bernie-laberge, line 542: I would remove this weird if and reset the _averageColor+_colorSamples to 0 when we want to reset the average.

src/lib/app/mu_rvui/inspector.mu
ThomBern: The reason this if is necessary (or was used) is because it prevents the update to the averageColor WHILE we drag the inspector (prevents it any time _glReadPending is true). If we get rid of it, we'd still need another type of bool to prevent that from happening.

ThomBern: I found a fix, since we don't want the drag to update the average color, let's just make sure that we don't enter that big if statement if we're dragging.

src/lib/app/mu_rvui/inspector.mu
```diff
@@ -122,8 +121,8 @@ class: Inspector : Widget
         _colorPointEvent = ip;
         _colorPointImage = eventToImageSpace(sName, ip);
         _colorSourceInfo = pinfo;
-        _averageColor    = sp.v;
-        _colorSamples    = 1;
+        _averageColor    = Color(0, 0, 0);
```
bernie-laberge, line 124: If we are not in FinalColorMode (_showFBColor==false) then the original code was correctly initializing the average.
So perhaps we should do something like:
```
if (_showFBColor)
{
   _averageColor = Color(0, 0, 0);
   _colorSamples = 0;
}
else
{
   _averageColor = sp.v;
   _colorSamples    = 1;
}
```

src/lib/app/mu_rvui/inspector.mu
```diff
@@ -175,17 +179,14 @@ class: Inspector : Widget
...
-            //  _fbColor = Color(pixels[0], pixels[1], pixels[2], pixels[3]);
         }
 
         _currentColor     = sp.v;
         _colorPointEvent  = ip;
         _colorPointPixel  = Point(pinfo.px, pinfo.py);
         _colorPointImage  = eventToImageSpace(sName, ip);
         _colorSourceInfo  = pinfo;
-        _averageColor    += sp.v;
+        _averageColor    += if _showFBColor then _fbColor else _currentColor;
```
bernie-laberge, line 189: In final color mode (showFBColor==true), we cannot really update the average color here because the new value will be read at the next render (_flReadPending==true).
I think this code has to be conditional:
```
if (!_showFBColor)
{
   _averageColor+=_currentColor;
   _colorSamples+=1;
}
```

src/lib/app/mu_rvui/inspector.mu
```diff
@@ -225,11 +226,11 @@ class: Inspector : Widget
 
             if (_showFBColor)
             {
-                _glReadPending = true;
-                //  float[] pixels;
-                //  glReadPixels(offpt.x, offpt.y, 1, 1, GL_RGBA, pixels);
-                //  _fbColor = Color(pixels[0], pixels[1], pixels[2], pixels[3]);
+                float[] pixels;
```
bernie-laberge, line 229: I think that if we real the pixel here, we we are going to re-introduce what Alan tried to solve 10 years ago
https://git.autodesk.com/media-and-entertainment/shotgun-rv/commit/b7832a9bbafd6fcb23f9834e39121a879bb7175b

src/lib/app/mu_rvui/inspector.mu
```diff
@@ -155,14 +151,22 @@ class: Inspector : Widget
         _colorPointEvent = ip;
         _colorPointImage = eventToImageSpace(sName, ip);
         _colorSourceInfo = pinfo;
-        _averageColor    = sp.v;
-        _colorSamples    = 1;
+        _averageColor    = Color(0, 0, 0);
+        _colorSamples    = 0;
         _colorFrame      = frame();
+
+        _skipNextSamplingAtPoint = true;
```
bernie-laberge, line 158: Do we absolutely need this new _skipNextSamplingAtPoint ?

src/lib/app/mu_rvui/inspector.mu
```diff
@@ -111,9 +113,8 @@ class: Inspector : Widget
         if (_showFBColor)
         {
             _glReadPending = true;
-            //  float[] pixels;
-            //  glReadPixels(ip.x, ip.y, 1, 1, GL_RGBA, pixels);
-            //  _fbColor = Color(pixels[0], pixels[1], pixels[2], pixels[3]);
+        _averageColor    = Color(0, 0, 0);
```
bernie-laberge, line 116: _averageColor and _colorSamples are not initialized when _showFBColor==false ?

bernie-laberge: missing indentation

geffrak (approved):

PR #183, opened 2023-06-20
Add Support for Directories in RV Packages SupportPath

shannonyearwood:
### Summarize your change.
When creating a package you can list folders with files that should be installed in the SupportPath Plugins folder rather than listing each file individually

It should be easier/quicker to create packages

MacOS

- Packages are zipped with directory structure
- All files in package and its directories are zipped
- The app will traverse the PACKAGE file and see the list of folders specified. Based on this, all files under the folder and its subdirectories will be added to location provided.

N.B. When creating PACKAGE file if you want to use new behaviour it takes the following format:
folders:
- folder: [FolderName]
   location: [LocationName]

### DIFF
--- a/cmake/macros/rv_stage.cmake
+++ b/cmake/macros/rv_stage.cmake
@@ -456,7 +456,7 @@ FUNCTION(rv_stage)
       COMMENT "Creating ${_package_filename} ..."
       OUTPUT ${_package_filename}
       DEPENDS ${_files} ${_package_file}
-      COMMAND zip -v -J ${_package_filename} ${_files} ${_package_file}
+      COMMAND zip -v -j ${_package_filename} ${_files} ${_package_file}
       WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
     )
 
--- a/cmake/macros/rv_stage.cmake
+++ b/cmake/macros/rv_stage.cmake
@@ -4,16 +4,6 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-# Empty those lists at the begining of a configure stage
-SET(RV_PACKAGE_LIST
-    ""
-    CACHE INTERNAL ""
-)
-SET(INSTALLED_RV_PACKAGE_LIST
-    ""
-    CACHE INTERNAL ""
-)
-
 #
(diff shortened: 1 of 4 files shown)


cmake/macros/rv_stage.cmake
```diff
@@ -456,7 +456,7 @@ FUNCTION(rv_stage)
       COMMENT "Creating ${_package_filename} ..."
       OUTPUT ${_package_filename}
       DEPENDS ${_files} ${_package_file}
-      COMMAND zip -v -j ${_package_filename} ${_files} ${_package_file}
+      COMMAND zip -v -J ${_package_filename} ${_files} ${_package_file}
```
geffrak, line 459: I think you can remove the `-J` as it was a typo for `-j` and you implemented the required to not required `-j` anymore.

src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -1070,6 +1110,34 @@ PackageManager::loadPackageInfo(const QString& infileNonCanonical)
 				yaml_event_delete(&input_event);
 				yaml_parser_delete(&parser);
 
+        // Standardize the file and directory paths
+        for (auto& fileName : package.files) {
+          fileName.replace("\\", "/");
```
geffrak, line 1115: There's no native Qt function for this operation?

geffrak (approved):
Looks great. Does it work for both install and uninstall?

geffrak:
What happens if your packages contain a folder but don't have any `folder` directive to handle it?

src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -910,7 +923,14 @@ PackageManager::loadPackageInfo(const QString& infileNonCanonical)
                 temp[255] = 0;
                 files.push_back(temp);
                 QString f = temp;
-                if (f != "PACKAGE") m_packages.back().files.push_back(f);
+                // Check if it's a directory by checking if it ends with a '/'
+                if (f.endsWith("/"))
```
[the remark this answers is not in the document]
geffrak: There might indeed be a Qt native call to check if it's a directory

rogernelson (approved):

PR #185, opened 2023-06-20
Add dmg packager for Mac

ThomBern:
### Summarize your change.
Adds a command call in the post_install_darwin.cmake file that creates a .dmg with the result from the rvinst command.

We need to bring back the Mac dmg packaging from 2023 in Open RV.

Mac (changes for Mac only)
- post_install_darwin.cmake now contains something
- updated config_macos.md and .gitignore
- fixed typo in the root CMakeLists.txt


cmake/install/post_install_darwin.cmake
```diff
@@ -3,3 +3,9 @@
 #
 # SPDX-License-Identifier: Apache-2.0
 #
+
+# Create the .dmg
+EXECUTE_PROCESS(
+  COMMAND create-dmg --volname OpenRV --volicon "${CMAKE_SOURCE_DIR}/src/bin/nsapps/RV/RV.icns" --eula "${CMAKE_SOURCE_DIR}/LICENSE" OpenRV.dmg ./_install
```
geffrak, line 9: Please do not hardcode OpenRV. Use RV_INTERNAL_APPLICATION_NAME instead.

geffrak (changes requested):

geffrak (approved):

rogernelson (approved):

bernie-laberge (approved):

geffrak (changes requested):
Not required anymore given that we won't be making DMGs on OpenRV

PR #194, opened 2023-07-04
Fix "Fit Window to First Media Loaded" for EXRs

ThomBern:
### Summarize your change.
Added an if statement in Session.cpp to prevent the resizing of the window before the first media is loaded (which would happen with progressive loading enabled). It now resizes after it's loaded, when receiving the 'after-progressive-loading' event.
Whenever the RV_PROGRESSIVE_SOURCE_LOADING env and "Fit Window to First Media Loaded" preference was set to true, some media would not get fitted correctly, or wouldn't at all. This PR fixes this by bringing parity to the RV_PROGRESSIVE_SOURCE_LOADING behaviours (true and false).

Mac Rocky 9
- window_mode.mu
Before fix
Before the fix, on RV launch with image in args
After fix
After the fix, on RV launch with image in args


src/lib/app/mu_rvui/window_mode.mu
```diff
@@ -6,6 +6,7 @@
 module: window_mode {
 use rvtypes;
 use commands;
+use rvui;
```
bernie-laberge, line 9: We cannot include rvui here. There is a comment on line 17: 
<<NOTE: this can't be dependent on rvui since its required by it>>

src/lib/app/mu_rvui/window_mode.mu
```diff
@@ -95,6 +96,7 @@ class: WindowMinorMode : MinorMode
         let newPos = qt.QPoint(mainWin.x() + deltaX, mainWin.y() + deltaY);
         if (deltaX != 0 || deltaY != 0) mainWin.setPos(newPos);
 
+        centerResizeFit();
```
bernie-laberge, line 99: I don't think it is the intended purpose of this method. 
The centerFit() ('W' key stroke) does that.
'w' is not supposed to center.

src/lib/app/mu_rvui/window_mode.mu
```diff
@@ -109,6 +111,18 @@ class: WindowMinorMode : MinorMode
         if isConsoleVisible() then CheckedMenuState else UncheckedMenuState;
     }
 
+    method: afterProgressiveLoading (void; Event event)
+    {
+	    event.reject();
```
bernie-laberge, line 116: code not aligned with the line under it.

src/lib/app/mu_rvui/window_mode.mu
```diff
@@ -109,6 +111,18 @@ class: WindowMinorMode : MinorMode
...
+        if(progressiveSourceLoading()) {
+            let ns = computePixelRelativeScale(1.0, false);
```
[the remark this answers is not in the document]
bernie-laberge: I think that the real fix should be here:
https://github.com/AcademySoftwareFoundation/OpenRV/blob/605ee39fae539581d397ea0ed5ba0ebd5ca94793/src/lib/app/RvCommon/GLView.cpp#L334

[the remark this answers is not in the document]
bernie-laberge: I believe m_postFirstNonEmptyRender is initialized with the inverse of 'Fit Window to First Media Loaded'.

bernie-laberge (approved):

geffrak (approved):

PR #196, opened 2023-07-06
Directory support for rv packages

shannonyearwood:
### Summarize your change.
When creating a package you can list folders with files that should be installed in the SupportPath Plugins folder rather than listing each file individually

It should be easier/quicker to create packages

MacOS, Windows, Linux

Add a list of changes, and note any that might need special attention during the review.
Packages are zipped with directory structure
All files in package and its directories are zipped
The app will traverse the PACKAGE file and see the list of folders specified. Based on this, all files under the folder and its subdirectories will be added to location provided.
N.B. When creating PACKAGE file if you want to use new behaviour it takes the following format:
folders:

folder: [FolderName]
location: [LocationName]

### DIFF
--- a/src/lib/app/RvPackage/PackageManager.cpp
+++ b/src/lib/app/RvPackage/PackageManager.cpp
@@ -1112,10 +1112,10 @@ PackageManager::loadPackageInfo(const QString& infileNonCanonical)
 
         // Standardize the file and directory paths
         for (auto& fileName : package.files) {
-          fileName.replace("\\", "/");
+          fileName = QDir::toNativeSeparators(fileName);
         }
         for (auto& auxFolder : package.auxFolders) {
-          auxFolder.folder.replace("\\", "/");
+          auxFolder.folder = QDir::toNativeSeparators(auxFolder.folder);
         }
 
         for (auto& auxFolder : package.auxFolders) {
--- a/src/lib/app/RvPackage/PackageManager.cpp
+++ b/src/lib/app/RvPackage/PackageManager.cpp
@@ -1,9 +1,9 @@
 //
-//  Copyright (c) 2010 Tweak Software. 
+//  Copyright (c) 2010 Tweak Software.
 //  All rights reserved.
-//  
+//
 //  SPDX-License-Identifier: Apache-2.0
-//  
+//
 //
 #include <RvPackage/PackageManager.h>
 #include <TwkDeploy/Deploy.h>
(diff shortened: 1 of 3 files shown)


geffrak:
Also, can you pull the ASWF OpenRV main branch in your branch (using --rebase) to eliminate the already merged commits?

src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -1070,6 +1106,26 @@ PackageManager::loadPackageInfo(const QString& infileNonCanonical)
 				yaml_event_delete(&input_event);
 				yaml_parser_delete(&parser);
 
+        for (auto& auxFolder : package.auxFolders) {
```
shannonyearwood, line 1109: @geffrak This code block is the difference. Removed the use of the QT native separators

geffrak: Ok great!

src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -948,9 +964,11 @@ PackageManager::loadPackageInfo(const QString& infileNonCanonical)
 				package.openrvversion = "1.0.0";
 				bool modeState = false;
 				bool auxFileState = false;
+        bool auxFolderState = false;
```
bernie-laberge, line 967: Missing indentation ?

shannonyearwood: Very strange, the indentation is there on my local machine

bernie-laberge: That's because this file is filled with tabs (not your fault : legacy !)

shannonyearwood: Should I reformat the file to only use one (tabs or spaces)?

bernie-laberge: Are you sure you set your visual studio code to the right number of spaces per tab ?
It should be:
    "editor.insertSpaces": true,
    "editor.tabSize": 4,
https://code.visualstudio.com/docs/editor/codebasics#_indentation

shannonyearwood: I'm using CLion but the tab size is already set to 4

bernie-laberge: If you could that would be great : We prefer spaces. NOT tabs.

shannonyearwood: Fixed

bernie-laberge: Thank you !!!

geffrak (approved):

bernie-laberge (approved):

PR #197, opened 2023-07-06
Disable OpenSSL asm runtimes on Windows

geffrak:
### Summarize your change.

Reverts #150 on Windows.

There are reports in the field that there are build failures related to nasm on the OpenSSL build. It seems to work fine in my test environment, but not everywhere. It's safe to disable it on Windows since it was disabled at first. It only got enabled because it was required on MacOS and I wanted to have the same configuration on all 3 platforms.

It's a revert. It's the configuration that is being used in the Commercial RV build.

N/A

N/A


tbabiin (approved):
LGTM and tested on my Windows 11 system.

PR #198, opened 2023-07-06
Rewrite the rvinstall custom command

geffrak:
Moves the rvpkg installation commmand inside of a python script so the package list doesn't have to be known when the custom command get defined.

### Summarize your change.

Rewrite the step that installs the rvpkg packages so it always install all the packages that are built into the stage/packages folder.

Before this change, the package targets that get defined after the rvinstall target are not automatically installed into OpenRV. It's the case when the OpenRV build is used as a sub-project, like when Autodesk RV is built.

The script is platform agnostic, and tested on MacOS.

N/A

N/A


savoiei (approved):

bernie-laberge (approved):
Superb !
GG

PR #204, opened 2023-07-18
Ocio v2 feature branch update with main

tbabiin:
### Summarize your change.

Sync'ing the OCIO v2 feature branch with main.

The branches must stay aligned in order to easily merge in main in the future.

n/a


tbabiin:
I think this is a mistake, squashing will cause issues.

PR #208, opened 2023-07-21
Update documentation for Windows and Linux and build fix for RHEL8

tbabiin:
### Summarize your change.

Updating Windows build instructions & Fixing Linux build issue

Using MSYS to build on Windows has many non obvious choices such as how the PATH is set which is used to select App version; if the wrong version of an App is used, the build will break.

Win 10 & Win 11 and Linux Rocky 8

Most of the updates are for build docs but for Rocky8 there's a tweak on the build code; I can remove it if it isn't wanted or incorrect.t
Re-adding the Py7zr workaround for PIP
Adding new file for Rocky8; it's the same as Rocky 9 except the OpenSSL section which is new and only for Rocky8.


cmake/dependencies/openssl.cmake
```diff
@@ -166,6 +166,11 @@ IF(RV_TARGET_WINDOWS)
     ${_target}-stage-target ALL
     DEPENDS ${RV_STAGE_BIN_DIR}/${_crypto_lib_name} ${RV_STAGE_BIN_DIR}/${_ssl_lib_name}
   )
+ELSEIF(RV_TARGET_IS_RHEL8)
```
tbabiin, line 169: Reason for this change: Don't use RV OpenSSL for Rocky8.

docs/build_system/build_errors.md
```diff
@@ -146,6 +146,30 @@ Ref.: [Resolve "DSO missing from command line" error](https://zhangboyi.gitlab.i
...
+      note: This is an issue with the package mentioned above, not pip.
+      hint: See above for details.
+      [end of output]
+
+  note: This error originates from a subprocess, and is likely not a problem with pip.
+error: subprocess-exited-with-error
+
+× pip subprocess to install build dependencies did not run successfully.
+```
+Note: There's an issue with the latest version of MSYS2 and Python3.xx versions. Enter the following command from the same MSYS2-MinGW64 shell if you encounter an error when installing the **py7zr** python requirement:
```
tbabiin, line 167: Re-adding Bernie's fix for Python on Windows. I had the issue on both Win10 and Win11.

docs/build_system/config_windows.md
```diff
@@ -73,10 +73,41 @@ pacman -S --needed \
         zip
 ```
 
+While installing the MSYS packages, review the list for any missing package. Some packages might not be installed after the first command.
+
+Note: To confirm which version/location of any tool used inside the MSYS shell, `where` can be used e.g. `where python`. If there's more than one path return, the top one will be used.
+
+### Setting the PATH
+
+The path to mingw's bin must be added e.g. `msys64/mingw64/bin`.
```
tbabiin, line 82: MSYS PATH is so bad. The only way to make it work is to use Python from MinGW.

bernie-laberge (approved):

geffrak (approved):

PR #209, opened 2023-07-24
Un-hardcoding RV Icon path

ThomBern:
### Summarize your change.
Added a variable in CMake that contains the path to the Open RV Icon. Will be used to set the RV Commercial path.

Needed for the RV Commercial About Box.

MacOS

 - RvCommon's CMakeLists
 - rv_git.cmake


src/lib/app/RvCommon/CMakeLists.txt
```diff
@@ -230,6 +230,11 @@ IF(RV_TARGET_WINDOWS)
   )
 ENDIF()
 
+SET(RV_ICON
+    ":/images/RV_icon.png"
```
geffrak, line 234: How does it work? This is a path to a QRC resource. Did you find a way to add an extra QRC from another project to the OpenRV build? It's possible but won't magically work unless you know it's working. 

An easier way would be to copy the content of the qrc folder next to this file into $CMAKE_BINARY_DIR and then override the thumbnail if there's an RV_ICON override set.

bernie-laberge: You should run Kerby's  script Thomas so that you can experiment and make sure it works both in Open RV and https://git.autodesk.com/media-and-entertainment/rv:
https://o02-fcia.jenkins.autodesk.com/view/Shotgun/job/media-and-entertainment/job/rv-cicd-scripts/job/sync-external-branch/build?delay=0sec

geffrak:
It doesn't allow to change the app Icon, is it expected?

src/lib/app/RvCommon/CMakeLists.txt
ThomBern: I took the string that was in QIcon(...) and put it in a the RV_ICON cmake var, I didn't do anything really wonky, but there obviously should be better ways of handling this.

geffrak (approved):

bernie-laberge (approved):

PR #212, opened 2023-07-25
Added RV Icon as a cmake var

ThomBern:
### Summarize your change.
Added a variable in CMake that contains the path to the Open RV Icon. Will be used to set the RV Commercial path.

Needed for the RV Commercial About Box and the Application icon.

MacOS

 - RvCommon's CMakeLists
 - rv_global.cmake (to expose the ressource directory for RV Commercial)
 - RvApplication.cpp and RvDocument.cpp now use the CMake Icon var.


bernie-laberge (approved):

geffrak:
How do you update the app icon of the .app with this pull request? Or it's for another PR?

ThomBern:
> How do you update the app icon of the .app with this pull request? Or it's for another PR?

The rest of the changes would be in RV Commercial.

geffrak (approved):

PR #215, opened 2023-07-25
Add Video Output API plugin support (SG-29837)

bernie-laberge:
### Add Video Output API plugin support (SG-29837)

This commit adds the Video Output plugins capability to Open RV.

### Summarize your change.

The video output devices are no longer directly instantiated in the RV main.cpp (the createDevices() and showDevices())
They are dynamically loaded instead provided that a dynamic library (located under the plugins/output sub-directory by default but configurable via the TWK_OUTPUT_PLUGIN_PATH environment variable) implements the two following exported functions:
//      extern "C" TwkApp::VideoModule* output_module_create(float output_plugin_version, int output_module_index);
//      extern "C" void output_module_destroy(TwkApp::VideoModule*);
Note that the TwkApp::VideoModule is basically a factory for TwkApp::VideoDevice objects, where each instance of TwkApp::VideoDevice can be used as an RV presentation device (a broadcast or an output device if you prefer)
rv_prefs_video_output

Some dead code was removed which I discovered was not used (not even in the commercial version of RV): 
src/lib/app/OutputVideoDevices/...

Prior to this commit, the Video Output (AJA+Blackmagic) were directly instantiated in the commercial version of RV and could not be extended to other third party video output implementations. This commit opens the door to third party video output implementations such as NDI for example.

Tested on Mac using the commercial RV AJA implementation (which leverages the exact same new video output plugin architecture because it is based on Open RV as a submodule).


src/lib/app/TwkApp/OutputPlugins.cpp
```diff
@@ -0,0 +1,107 @@
...
+    {
+        std::string pluginPath;
+
+        if (getenv(pathVar.c_str()))
+        {
+            pluginPath = getenv(pathVar.c_str());
+        }
+        else
+        {
+            pluginPath = ".";
```
geffrak, line 26: Loading plugins from $pwd? Do we want that? it is already done elsewhere?

geffrak (approved):

src/lib/app/TwkApp/OutputPlugins.cpp
bernie-laberge: Yes. This is what is done for both the Movie and the Image plugins:
https://github.com/AcademySoftwareFoundation/OpenRV/blob/3b509ab0c07c803aac4b9fbd406043d0adc21535/src/lib/image/TwkMovie/MovieIO.cpp#L334
https://github.com/AcademySoftwareFoundation/OpenRV/blob/3b509ab0c07c803aac4b9fbd406043d0adc21535/src/lib/image/TwkFB/IO.cpp#L327

bernie-laberge: But only if the request is done without an env var but we always pass an env var.

src/lib/app/TwkApp/Application.cpp
```diff
@@ -7,8 +7,11 @@
 //******************************************************************************
 
 #include <TwkApp/Application.h>
-#include <assert.h>
+#include <TwkApp/OutputPlugin.h>
+#include <TwkApp/VideoDevice.h>
+#include <TwkApp/VideoModule.h>
 #include <stl_ext/stl_ext_algo.h>
+#include <assert.h>
```
rogernelson, line 14: Why assert.h? I don't see any asserts.

src/lib/app/TwkApp/OutputPlugin.cpp
```diff
@@ -0,0 +1,92 @@
...
+    {
+        output_module_create_t*  plugCreate  = reinterpret_cast<output_module_create_t*>(load_plugin_fct(file, handle, "output_module_create"));
+        output_module_destroy_t* plugDestroy = reinterpret_cast<output_module_destroy_t*>(load_plugin_fct(file, handle, "output_module_destroy"));
+        if (!plugCreate || !plugDestroy)
+        {
+            dlclose( handle );
+            return nullptr;
+        }
+
+        return new TwkApp::OutputPlugin(file, handle, plugCreate, plugDestroy);
```
rogernelson, line 46: Return raw pointers is pretty prone to introducing memory leaks.  Perhaps consider returning a `std::shared_ptr` so the transfer of ownership is obvious?

src/lib/app/TwkApp/TwkApp/OutputPlugins.h
```diff
@@ -0,0 +1,57 @@
...
+//
+
+class OutputPlugins
+{
+  public:
+
+    OutputPlugins();
+    virtual ~OutputPlugins();
+
+    typedef std::unordered_set<OutputPlugin*> Plugins;
```
rogernelson, line 29: The modern way to write this is
`using Plugins = std::unordered_set<OutputPlugin*>;`

They're equivalent though, I just find the new way more readable, but feel free to use what you like.

As I also noted it above, you could avoid having to manual memory management with:

`using Plugins = std::unordered_set<std::shared_ptr<<OutputPlugin>>;`

rogernelson (approved):
This looks really good Bernie!

src/lib/app/TwkApp/OutputPlugin.cpp
bernie-laberge: Totally. I will make that change.

src/lib/app/TwkApp/TwkApp/OutputPlugins.h
bernie-laberge: Totally. Much nicer. I'll do it. Thanks!

PR #216, opened 2023-07-26
Updating SHA1 cmake vars

ThomBern:
### Summarize your change.
Changed the way we set the RV_GIT_COMMIT_SHORT_HASH cmake var, so we can append the Commercial RV hash to the Open RV one.
We want to be able to show both the Commercial and Open RV sha1 in the about box and shell.
MacOS, but all changes were in CMake, so should be good on all platforms.
- Deleting the CMakeCache might be necessary.


geffrak:
I don't think this pull request is required. We could do the same thing directly from the Commercial RV project.

PR #219, opened 2023-08-04
Add Event-based MediaLibrary

rogernelson:
[Create Event-Based MediaLibrary] 

This branch creates an event-based MediaLibrary, which allows for users to add custom http headers and/or cookies to any request to read media.

A new class called EventMediaLibrary has been derived from the base Library class to accomplish this.  For the time being it will only generate events to get cookies and headers with default implementations for the rest of the virtual methods, but if needed, the remaining methods could also generate events.  Library node types are broadly divided into media and non-media types, so I just added an EventRootNode and a EventMediaNode, since we only really care about media nodes in this context.

These virtual methods are triggered whenever we attempt to read media through the FileSource node, which will ask any MediaLibrary instances if they own the path/url about to be read.  The EventMediaLibrary added by this branch will dispatch an event called `media-library-is-url` when this occurs, allowing packages also act as MediaLibraries if they respond with `"true"` to the event.

If the path/url is owned by the MediaLibrary it can optionally add cookies to the ffmpeg request used to read the media, which is useful if the url is located on a server that requires authorization. Another event has been added for this purposed called `media-library-get-http-cookies` which can return cookies in the format `domain;path;name;value` (where `;` is the delimiter for the various parts and a newline is the delimiter for multiple cookies).

Lastly, since authorization can also be done with headers (like bearer tokens), I also added support for this to the FileSourceNode, and a corresponding event (`media-library-get-http-headers`) to return cookies in the format `name;value` (where `;` is the delimiter and a newline is the delimiter for multiple headers).

This has been tested on Mac OS X.

The following test code  in a dummy package worked successfully and could be used as a template for creating authorized requests:
```
from rv import rvtypes
from urllib.parse import urlparse

class ShotgridMediaLibrary(rvtypes.MinorMode):
    def __init__(self, server):
        super(ShotgridMediaLibrary, self).__init__()
        self.init(
            "shotgrid_media_library",
            [
                (
                    "media-library-is-url",
                    self.is_url,
                    "Determine if this is a shotgrid url",
                ),
                (
                    "media-library-get-mime-type",
                    self.get_mime_type,
                    "Get mime type for a shotgrid url",
                ),

                (
                    "media-library-get-http-cookies",
                    self.get_auth_cookie,
                    "Get the authorization cookie",
                ),
            ],
            None,
        )
        self.server = server
    
    def is_our_url(self, event):
        url = urlparse(event.contents())
        return (url.scheme == "http" or url.scheme == "https") and url.netloc == self.server

    def is_url(self, event):
        event.reject()
        if self.is_our_url(event):
            event.setReturnContent("true")

    def get_mime_type(self, event):
        event.reject()
        if self.is_our_url(event):
            event.setReturnContent("video/mp4")
 
    def get_auth_cookie(self, event):
        event.reject() 
        if self.is_our_url(event):
            domain = self.server
            path = "/"
            name = "_session_id"
            value = "dummy" # TODO: get shotgrid auth cookie
            event.setReturnContent(
                "{};{};{};{}".format(domain, path, name, value)
            )

def createMode():
    server = "shotgrid.com" 
    return ShotgridMediaLibrary(server)
```


src/lib/ip/IPBaseNodes/FileSourceIPNode.cpp
```diff
@@ -2572,6 +2572,16 @@ FileSourceIPNode::openMovieTask(const string& filename,
                               << "; domain=" << cookies[c].domain;
                 }
                 m_inparams.push_back(StringPair("cookies", cookieStm.str()));
+
+                ostringstream headersStm;
+                TwkMediaLibrary::HTTPHeaderVector headers = api->httpHeaders();
```
rogernelson, line 2577: @mattdaw This part may be of interest to you.

bernie-laberge (approved):
Wow !
Beautiful work Roger !

geffrak (approved):

PR #230, opened 2023-08-09
Enable optional FFmpeg and mio_ffmpeg customization

bernie-laberge:
[ Enable optional FFmpeg and mio_ffmpeg customization ]

These small changes are required so that one can customize both FFmpeg and mio_ffmpeg if say one project was to consume Open RV as a submodule.

### Summarize your change.

-  Only add cmake custom target 'dependencies' but only if it hasn't been created already (ADD_CUSTOM_TARGET(dependencies))-To allow a project consuming Open RV as a sub-module to add the target first.
-  FFmpeg.cmake : Add an optional post_configure_step that is added only if the RV_FFMPEG_POST_CONFIGURE_STEP cmake variable is defined-To allow applying patches/customization to FFmpeg.
- mio_ffmpeg : Skip building of Open RV mio_ffmpeg if cmake var OPEN_RV_SKIP_MIO_FFMPEG is set (optional) - To enable a project consuming Open RV as a sub-module to build its own customized version.
- MovieFFMpeg : Now set the cmake variable MOVIE_FFMPEG_SOURCE_DIR so that a project consuming Open RV as a sub-module can build its own customized version.
- MovieFFMpeg.cpp : Added  a mechanism to register custom codecs when MOVIE_FFMPEG_IO_REGISTER_CUSTOM_CODEC is defined.

Successfully built on all platforms and tested on macOS

### DIFF
--- a/cmake/dependencies/ffmpeg.cmake
+++ b/cmake/dependencies/ffmpeg.cmake
@@ -212,6 +212,15 @@ IF(RV_TARGET_WINDOWS)
   )
 ENDIF()
 
+IF(RV_FFMPEG_POST_CONFIGURE_STEP)
+  EXTERNALPROJECT_ADD_STEP(
+    ${_target} post_configure_step
+    ${RV_FFMPEG_POST_CONFIGURE_STEP}
+    DEPENDEES configure
+    DEPENDERS build
+  )
+ENDIF()
+
 FILE(MAKE_DIRECTORY ${_include_dir})
 
 FOREACH(
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -5449,7 +5449,8 @@ MovieFFMpegIO::MovieFFMpegIO(CodecFilterFunction codecFilter,
                              bool bruteForce,
                              int codecThreads,
                              string language,
-                             double defaultFPS)
+                             double defaultFPS,
+                             void (*registerCustomCodecs)() /*=nullptr*/)
     : MovieIO("MovieFFMpeg","v2"),
       m_codecFilter(codecFilter)
 {
@@ -5464,6 +5465,9 @@ MovieFFMpegIO::MovieFFMpegIO(CodecFilterFunction codecFilter,
 
     // Store a default frame rate just in case
     setDoubleAttribute("defaultFPS", defaultFPS);
+
+    // Register custom codecs if requested (optional)
+    if (registerCustomCodecs) registerCustomCodecs();
 
     // Init av
     avformat_network_init();
--- a/src/lib/image/mio_ffmpeg/CMakeLists.txt
+++ b/src/lib/image/mio_ffmpeg/CMakeLists.txt
@@ -4,29 +4,39 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-INCLUDE(cxx_defaults)
-
-SET(_target
-    "mio_ffmpeg"
+# Note: This cache variable can be overriden to FALSE in a super-project if a custom mio_ffmpeg is built as part of that super-project.
+SET(RV_BUILD_MIO_FFMPEG
+    TRUE
+    CACHE BOOL "Whether or not to build the Open RV mio_ffmpeg movie plugin"
 )
 
-LIST(APPEND _sources init.cpp)
+IF(RV_BUILD_MIO_FFMPEG)
 
-ADD_LIBRARY(
-  ${_target} SHARED
-  ${_sources}
-)
+  INCLUDE(cxx_defaults)
 
-TARGET_LINK_LIBRARIES(
-  ${_target}
-  PRIVATE MovieFFMpeg Boost::program_options TwkFB
-)
-
-IF(RV_TARGET_WINDOWS)
-  TARGET_COMPILE_OPTIONS(
-    ${_target}
-    PRIVATE "-D__STDC_CONSTANT_MACROS"
+  SET(_target
+      "mio_ffmpeg"
   )
-ENDIF()
 
-RV_STAGE(TYPE "MOVIE_FORMAT" TARGET ${_target})
+  LIST(APPEND _sources init.cpp)
+
+  ADD_LIBRARY(
+    ${_target} SHARED
+    ${_sources}
+  )
+
+  TARGET_LINK_LIBRARIES(
+    ${_target}
+    PRIVATE MovieFFMpeg Boost::program_options TwkFB
+  )
+
+  IF(RV_TARGET_WINDOWS)
+    TARGET_COMPILE_OPTIONS(
+      ${_target}
+      PRIVATE "-D__STDC_CONSTANT_MACROS"
+    )
+  ENDIF()
+
+  RV_STAGE(TYPE "MOVIE_FORMAT" TARGET ${_target})
+
+ENDIF() # IF (RV_BUILD_MIO_FFMPEG)
--- a/cmake/dependencies/CMakeLists.txt
+++ b/cmake/dependencies/CMakeLists.txt
@@ -4,11 +4,16 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-ADD_CUSTOM_TARGET(dependencies)
-SET(RV_DEPS_LIST
-    ""
-    CACHE INTERNAL ""
-)
+IF(NOT TARGET dependencies)
+  ADD_CUSTOM_TARGET(dependencies)
+ENDIF()
+
+IF(NOT DEFINED RV_DEPS_LIST)
+  SET(RV_DEPS_LIST
+      ""
+      CACHE INTERNAL ""
+  )
+ENDIF()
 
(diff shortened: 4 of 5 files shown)


cmake/dependencies/ffmpeg.cmake
```diff
@@ -212,6 +212,15 @@ IF(RV_TARGET_WINDOWS)
   )
 ENDIF()
 
+IF(RV_FFMPEG_POST_CONFIGURE_STEP)
```
geffrak, line 215: Do we need to define this? It seems like we can define this on the commercial RV repo.

geffrak (approved):

src/lib/image/mio_ffmpeg/CMakeLists.txt
```diff
@@ -4,29 +4,33 @@
 # SPDX-License-Identifier: Apache-2.0
 #
 
-INCLUDE(cxx_defaults)
+IF(NOT OPEN_RV_SKIP_MIO_FFMPEG)
```
rogernelson, line 7: Minor point: I thin BUILD_MIO_FFMPEG would be a better name -- just to avoid double negatives like this (ie NOT + SKIP).

Also another potential way of doing this might be to use [EXCLUDE_FROM_ALL](https://cmake.org/cmake/help/latest/prop_tgt/EXCLUDE_FROM_ALL.html) in the commercial RV CMake so that OpenRV does not need to be modified at all.  I haven't tested it though, so I don't know 100% if it would work.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -5465,6 +5465,11 @@ MovieFFMpegIO::MovieFFMpegIO(CodecFilterFunction codecFilter,
     // Store a default frame rate just in case
     setDoubleAttribute("defaultFPS", defaultFPS);
 
+    // Mechanism to register custom codecs when defined
+#ifdef MOVIE_FFMPEG_IO_REGISTER_CUSTOM_CODEC
```
rogernelson, line 5469: Are there any cases where you would want `MOVIE_FFMPEG_IO_REGISTER_CUSTOM_CODEC`  and `NOT OPEN_RV_SKIP_MIO_FFMPEG` (below) to have different values? If so, perhaps we should just have one variable to make it more clear that if you want custom codecs you need to provide mio_ffmpeg as well?

rogernelson (approved):

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
rogernelson: Actually on second thought, you could probably avoid the define here entirely.

You could just add a call to method called `registerCustomCodecs()` here (or something like that) and in the default mio_ffmpeg code its a noop.  Then anyone that wants to register custom codecs just does so by implementing that method in their own mio_ffmpeg plugin.

cmake/dependencies/ffmpeg.cmake
bernie-laberge: FFmpeg is not built in the super project. It is built as part of the OpenRV submodule.
So I think that we do need to define this optional post_configure_step here.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
bernie-laberge: Both points are excellent points. I am making the suggested changes. Thank you Roger !

cmake/dependencies/ffmpeg.cmake
geffrak: So it's not possible to add this snippet in the super project with the right target name after it includes OpenRV?

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
bernie-laberge: About the first comment: 
The MOVIE_FFMPEG_IO_REGISTER_CUSTOM_CODEC and OPEN_RV_SKIP_MIO_FFMPEG affect two completely different libs and are two completely different mechanism. I could see a situation where someone would want to build its own version of mio_ffmpeg having say the AC3 audio codec enabled and not in the disallowed list of codec as it is currently. In this scenario there would be no nead to use the FFmpeg dynamic codec registration to introduce a new codec.

cmake/dependencies/ffmpeg.cmake
bernie-laberge: OK yes, I get it now, it should work !
But it would make the FFmpeg customization by the super project awkward no ?
Currently we have :
`ADD_SUBDIRECTORY(cmake/extra_dependencies)
ADD_SUBDIRECTORY(OpenRV)`
But if we really want to hide this optional post_configure_step we would have:
`ADD_SUBDIRECTORY(cmake/extra_dependencies)
ADD_SUBDIRECTORY(OpenRV)
IF(RV_FFMPEG_POST_CONFIGURE_STEP)
  EXTERNALPROJECT_ADD_STEP(
    RV_DEPS_FFMPEG post_configure_step
    ${RV_FFMPEG_POST_CONFIGURE_STEP}
    DEPENDEES configure
    DEPENDERS build
  )
ENDIF()`
Awkward but it should work. You are right (as usual).
Wouldn't you prefer to have the FFmpeg build intricacies all defined within the same file ? (OpenRV/cmake/dependencies/ffmpeg.cmake)

geffrak: This or we add a dependencies customization folder with all of these if we end up with multiple customization.

bernie-laberge: Another argument to defend the proposed implementation :-) 
At first I was using the already existing ExternalProject_Add/PATCH_COMMAND.
Now to leverage it, I didn't have to add an extra step (EXTERNALPROJECT_ADD_STEP), I just needed to use it in the ffmpeg.cmake/EXTERNALPROJECT_ADD. I was using a cmake var to do it like this:

```
IF(RV_FFMPEG_CUSTOM_PATCH_COMMAND)
  SET(_patch_command
      ${RV_FFMPEG_CUSTOM_PATCH_COMMAND}
  )
ENDIF()

EXTERNALPROJECT_ADD(
  ${_target}
  DEPENDS dav1d::dav1d RV_DEPS_OPENSSL
  DOWNLOAD_NAME ${_target}_${_version}.zip
  DOWNLOAD_DIR ${RV_DEPS_DOWNLOAD_DIR}
  DOWNLOAD_EXTRACT_TIMESTAMP TRUE
  INSTALL_DIR ${_install_dir}
  URL ${_download_url}
  URL_MD5 ${_download_hash}
  SOURCE_DIR ${RV_DEPS_BASE_DIR}/${_target}/src
  ${_patch_command}
  CONFIGURE_COMMAND ...
  BUILD_COMMAND ...
  INSTALL_COMMAND ...
  BUILD_IN_SOURCE TRUE
  BUILD_ALWAYS FALSE
  BUILD_BYPRODUCTS ${_build_byproducts}
  USES_TERMINAL_BUILD TRUE
)

bernie-laberge: We are going to end up with multiple customizations. Thomas is about to customize the _disable_decoders flag in ffmpeg.cmake.
This way, the superproject's ffmpeg customization could all be dictated from the same file like this one for example : https://git.autodesk.com/media-and-entertainment/rv/blob/external/bernie-laberge/apple_pro_res/cmake/extra_dependencies/ffmpeg_customization.cmake

bernie-laberge: Let's discuss it at the RV dailies today. I will make the changes required once we come up with a decision. No worries.

bernie-laberge: I have made the changes proposed by Kerby and the end result is actually cleaner in mu opinion ! Thank you Kerby !

src/lib/image/mio_ffmpeg/CMakeLists.txt
bernie-laberge: EXCLUDE_FROM_ALL is what I tried first but I was not able to make it work to my needs unfortunately.
Good point for BUILD_MIO_FFMPEG: I'll make the change.

cmake/dependencies/ffmpeg.cmake
bernie-laberge: Unfortunately this scheme doesn't work :(
EXTERNALPROJECT_ADD_STEP() is not working when called from a different directory than the EXTERNALPROJECT_ADD(). So I will have to revert my changes.

geffrak (approved):

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
bernie-laberge: Note that I had created a brand new class deriving from 
But I had forgotten that virtual methods are not working in constructors 🤦 
So I am now passing an optional  void (*registerCustomCodecs)()=nullptr to the constructor to achieve the same end.

rogernelson (approved):

PR #233, opened 2023-08-14
Added Configurable FFMpeg build

ThomBern:
### Summarize your change.
The FFMpeg external build now uses mostly CMake list vars for its configure step. 
We need to easily add new flags to the FFMpeg build, so that we can modify the latter depending on our needs (in our case, commercial RV, but the same mechanism could also be used for any super project consuming Open RV as a git submodule).
MacOS


rogernelson (approved):

geffrak (approved):

bernie-laberge (approved):
Beautiful Thomas !
Thanks

PR #241, opened 2023-08-22
Set the OSX/MacOS deployment target to all deps

geffrak:
### Summarize your change.

Makes sure that all OpenRV dependencies know about the required minimum deployment target. 

Before this pull request, CMAKE_OSX_DEPLOYMENT_TARGET was not properly applied to all third parties, so MacOS versions between CMAKE_OSX_DEPLOYMENT_TARGET and the host MacOS version could have unexpected crashes.

Example: std::filesystem requires a CMAKE_OSX_DEPLOYMENT_TARGET of 10.15. If someone were to introduce a third party that uses it but wants to target MacOS versions lower than 10.15, it would build but crash at runtime. 

With this change, it will either not build or take any alternative path (using experimental). 

Tested on MacOS Ventura
(Will update once I did the test on the other OS

I don't think I remove any functionality or compile flag, but a second pair of eye would be great. 

N/A


geffrak:
#242 is better.

PR #250, opened 2023-09-01
feature/vfx2023 update to vs2022 and boost to 1 80

tbabiin:
### Linked issues

Fixes #SG-31056

### Summarize your change.

Updating to VS2022 and the compiler toolchain V143

VFX 2023 platform update

Windows 11: RV boot start (on Debug)
Both Windows Build Configs built correctly

- PySide2 in a new Win11 machine with only VS2022 was able to build. RV also built


tbabiin:
❗ Note this will break the branch since OCIO v1 doesn't build with VS2022.

geffrak (approved):

bernie-laberge (approved):

PR #251, opened 2023-09-05
Fix Windows debug build caused by spdlog

bernie-laberge:

This commit fixes the windows debug build 

### Summarize your change.

When spdlog is built in debug, the 'd' suffix is added to the resulting static lib by the spdlog build system.
This 'd' suffix was added to the spdlog.cmake and I took the opportunity to do some cleanup.
Also, the CMAKE_BUILD_TYPE was NOT passed to the spdlog build on Mac+Linux. I fixed that too.

Tested on all platforms in both debug and release

### DIFF
--- a/cmake/dependencies/spdlog.cmake
+++ b/cmake/dependencies/spdlog.cmake
@@ -41,8 +41,14 @@ ELSE()
   )
 ENDIF()
 
+IF(CMAKE_BUILD_TYPE MATCHES "^Debug$")
+  SET(RV_SPDLOG_DEBUG_POSTFIX
+      "d"
+  )
+ENDIF()
+
 SET(_spdlog_lib_name
-    ${CMAKE_STATIC_LIBRARY_PREFIX}spdlog${CMAKE_STATIC_LIBRARY_SUFFIX}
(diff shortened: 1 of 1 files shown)


tbabiin:
(feeling the pain) DCO is failing. 😢 FYI @bernie-laberge

cmake/dependencies/spdlog.cmake
```diff
@@ -41,8 +41,14 @@ ELSE()
   )
 ENDIF()
 
+IF(${CMAKE_BUILD_TYPE} STREQUAL "Debug")
```
tbabiin, line 44: IF(CMAKE_BUILD_TYPE MATCHES "^Debug$") feels better to me. We have both style for this exact comparison (debug). Which one we should use everywhere?

tbabiin (approved):
LGTM! Only a code style comment.

cmake/dependencies/spdlog.cmake
tbabiin: Note that CMAKE_BUILD_TYPE doesn't need to interpolated here when using MATCHES.

rogernelson (approved):

cmake/dependencies/spdlog.cmake
bernie-laberge: Good catch ! 
Consider it done.
Thanks !

tbabiin: Thanks!

PR #253, opened 2023-09-11
Fix examples in doc not working because of outdated syntax

bernie-laberge:

A user reported that the fourth chapter of the RV reference manual contained examples that were using outdated syntax and therefore could not be executed.

### Summarize your change.

I have tested all the examples in the fourth chapter of the RV reference manual and two of them had outdated syntax which was corrected in this commit.

Additionally, a note was added to the Mu documentation to inform the user that the recommended RV scripting language is Python (not Mu).

### DIFF
--- a/docs/rv-manuals/rv-mu-programming.md
+++ b/docs/rv-manuals/rv-mu-programming.md
@@ -2,6 +2,8 @@
 
 ## 1. Overview ‌ ‌
 
+You can use Python3 in RV in conjunction with Mu or in place of it. It's even possible to call Python commands from Mu and vice versa. Python is a full peer to Mu as far as RV is concerned. So in answer to the question: which language should I use to customize RV? At this point, we recommend using Python.
+
 Mu is principally targeted towards computer graphics applications. The original (usable) version appeared at Tweak Films around 2001-2002. Over the years its syntax and runtime has been refactored and evolved from a simple shading language to a nearly full featured general language. However, Mu is still probably best suited for computer graphics than other computing tasks.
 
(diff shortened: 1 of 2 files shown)


geffrak (approved):

docs/rv-manuals/rv-mu-programming.md
```diff
@@ -2,6 +2,8 @@
 
 ## 1. Overview ‌ ‌
 
+You can use Python3 in RV in conjunction with Mu or in place of it. It's even possible to call Python commands from Mu and vice versa. Python is a full peer to Mu as far as RV is concerned. So in answer to the question: which language should I use to customize RV? The answer is whichever you like. At this point we recommend using Python.
```
mchesnay, line 5:
```suggestion
You can use Python3 in RV in conjunction with Mu or in place of it. It's even possible to call Python commands from Mu and vice versa. Python is a full peer to Mu as far as RV is concerned. So in answer to the question: which language should I use to customize RV? At this point, we recommend using Python.
```
I would do away with "The answer is whichever you like." We should take a stand, and recommend Python.

mchesnay (changes requested):

docs/rv-manuals/rv-mu-programming.md
bernie-laberge: Good point. I'll make the change here and also in the Python doc. Thanks!

mchesnay (approved):

compaga (commented):
Thanks

PR #257, opened 2023-09-14
Fix -noPrefs cmd line arg that was resetting prefs

brossee1:
### Summarize your change.

The settings are now loaded with the appropriate QSettings depending on if RV was started with _-noPrefs_ or not.

When starting RV with _-noPrefs_, the current preferences were ignored, but they were overwritten with the default values when closing RV.

The bug was tested on MacOS.

1. Launch the latest RV build
2. Go into _OpenRV / Preferences / General_ and change the font size to 15
3. Restart RV and make sure the font size is still 15
4. Close RV and restart it with _-noPrefs_
5. The font size is 13, which is the default value
6. Close RV and restart it again without any argument
7. The font size is now back to 15 like expected instead of 13

- [X] getQSettings() create a new QSettings with _"RVALT"_ as the application name if RV was started with _-noPrefs_ or with _INTERNAL_APPLICATION_NAME_ otherwise.
- [X] Remove QCoreApplication::setApplicationName( "RVALT" ); in RVSettings()
- [X] Run clang-format

### DIFF
--- a/src/lib/app/RvPackage/PackageManager.cpp
+++ b/src/lib/app/RvPackage/PackageManager.cpp
@@ -29,7 +29,7 @@ namespace Rv
   {
     QString zipfileCanonical = QFileInfo( zipfile ).canonicalFilePath();
     QString pattern = zipfile + "-.*\\.rvpkg";
-    QRegExp rx(pattern);
+    QRegExp rx( pattern );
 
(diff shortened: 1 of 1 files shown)


geffrak (approved):

bernie-laberge (approved):
Great catch and beautiful PR description Éloïse !
Thank you for the fix !

PR #259, opened 2023-09-18
Issue-162: OCIO v2 configs don't work on Mac; shader compiler fail

tbabiin:
### Linked issues
N/A

### Summarize your change.

The version comparison doesn't even check the GLSL version, maybe it should. For higher than GLSL 150, we will keep the current code. If the version is less than 150, we will use "version 120" as the header and remove the extension GL_ARB_texture header.

Mac fails to build with the current GLSL header.

Tested on Mac Ventura 13.2.1, Rocky 8 Dbg config.

Removing OCIO dep from OIIO as our OIIO config doesn't need OCIO 
Fixing byproducts for WebP on Mac and Linux for iterative builds 
Tweaking OCIO pipeline code in OCIOIPNode and setting the GPUProc as OPT_DEFAULT instead of OPT_NONE. ocioconvert uses OPT_DEFAULT for legacy 
Fixing bad GLSL Version comparison for Mac. We may need to iterate on this fix 
Fixing the Annonate Output when GLSL fails as it wasn't showing the exact GLSL that failed

FixMarciNotShadedV2Mac

### DIFF
--- a/src/lib/ip/IPCore/ShaderFunction.cpp
+++ b/src/lib/ip/IPCore/ShaderFunction.cpp
@@ -851,6 +851,27 @@ Function::replaceTextureCalls()
 }
 
 namespace {
+    // param: pass the GL_VERSION, not GLSL -- ex: glGetString(GL_VERSION)
+    const char* const get_glsl_header(const char* const glVersion)
+    {
+        if(nullptr == glVersion)
+        {
+            std::cerr << "get_glsl_header error: glVersion is null." << std::endl;
+            return global_glsl;
+        }
+
+        const char* glsl_header = global_glsl;
+        if(Shader::Function::isGLSLVersionLessThan150())
+        {
+            glsl_header = global_glsl_lt_150;
+        }
+        else
+        {
+            glsl_header = glVersion[0] <= '2' ? global_glsl_gl2 : global_glsl;
+        }
+
+        return glsl_header;
+    }
 void
 compileGLSL(const string& source, GLuint& shaderID, int& status, vector<char>& log)
 {
@@ -860,14 +881,7 @@ compileGLSL(const string& source, GLuint& shaderID, int& status, vector<char>& l
     int logsize;
 
     const char* src[2];
-    if(Shader::Function::isGLSLVersionLessThan150())
-    {
-        src[0] = global_glsl_lt_150;
-    }
-    else
-    {
-        src[0] = glVersion[0] <= '2' ? global_glsl_gl2 : global_glsl;
-    }
+    src[0] = get_glsl_header(glVersion);
     src[1] = source.c_str();
 
     glShaderSource(shaderID, 2, src, 0);
@@ -907,6 +921,8 @@ Function::compile() const
     }
     
     compileGLSL(m_sourceCode, m_state->shader, status, buffer);
+    const char* glVersion = (const char*)glGetString(GL_VERSION);
+    const std::string glsl_header(get_glsl_header(glVersion));
 
     if (status != GL_TRUE)
     {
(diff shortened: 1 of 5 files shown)


src/lib/ip/IPCore/ShaderFunction.cpp
```diff
@@ -111,6 +112,22 @@ Function::useShadingLanguageVersion(const char* glVersion)
...
+{
+    if(glslMajor > 1)
+    {
+        return false;
+    }
+    else
+    {
+        // Two edge case checks: check for 0 and check that minor versions are multiples of 10. Kronos spec is that minor is always a mult of 10.
+        // If not a multiple of 10; we abort and keep the number as-is.
+        unsigned int minor = glslMinor == 0 || glslMinor < 10 ? glslMinor : glslMinor % 10 != 0 ? glslMinor : glslMinor / 10;
```
geffrak, line 125: nested ternary are hard to read, can you break it into multiple `if`?

src/lib/ip/IPCore/ShaderFunction.cpp
```diff
@@ -826,7 +843,7 @@ compileGLSL(const string& source, GLuint& shaderID, int& status, vector<char>& l
     int logsize;
 
     const char* src[2];
-    src[0] = glVersion[0] <= '2' ? global_glsl_gl2 : global_glsl;
+    src[0] = Shader::Function::isGLSLVersionLessThan150() ? global_glsl_lt_150  : glVersion[0] <= '2' ? global_glsl_gl2 : global_glsl;
```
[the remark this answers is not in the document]
tbabiin: Done... also fixes some coding errors: bad casting on return and non-existent variable.

src/lib/ip/IPCore/ShaderFunction.cpp
```diff
@@ -826,7 +860,14 @@ compileGLSL(const string& source, GLuint& shaderID, int& status, vector<char>& l
     int logsize;
 
     const char* src[2];
-    src[0] = glVersion[0] <= '2' ? global_glsl_gl2 : global_glsl;
+    if(Shader::Function::isGLSLVersionLessThan150())
+    {
+        src[0] = global_glsl_lt_150;
```
bernie-laberge, line 865: With the aim of getting rid of the duplicated code, perhaps we could put all this logic in a helper function which would return the appropriate glsl code header : 
const char* get_glsl_header(); (get_glsl_header() would return one of the 3 possible globals)
src[0] = get_glsl_header();
It could be reused below in the debug output code outputAnnotatedCode().

src/lib/ip/IPCore/ShaderFunction.cpp
```diff
@@ -29,6 +29,7 @@ namespace {
 const char* global_glsl = "#version 150\n"
     "#extension GL_ARB_texture_rectangle : require\n";
 const char* global_glsl_gl2 = "#extension GL_ARB_texture_rectangle : require\n";
+const char* global_glsl_lt_150 = "#version 120\n";
```
bernie-laberge, line 32: Perhaps we should add for the global_glsl_lt_150 case just to be safe because it also means that we are dealing with an older version of OpenGL:
"#extension GL_ARB_texture_rectangle : require\n"; 
But on second thought, I don't think it is required because:

GL_ARB_texture_rectangle is an extension, and it made it into core GL with version 2.1, so there is no need to activate this extension "manually" in modern programmable GL.
Note that on mac, which is the lowest common denominator, we have :  OpenGL 2.1 + GLSL 1.2.

tbabiin:
Please merge this branch when you approve since I am not authorized to merge.

geffrak (approved):

bernie-laberge (approved):

PR #261, opened 2023-09-19
Fix rvpush not merging sessions when tag name has underscores

eloisebrosseau:
### Summarize your change.

The rvpush command can now be run to merge sessions with a tag name containing underscores.

Two sessions were being created instead of merging them into the same one when the tag name had underscores it in.

The bug was tested on MacOS.

```
// Run this command
./rvpush -tag "tag_name" merge "Movie1.mov"
// And this one
./rvpush -tag tag_name merge "Movie2.mov"
```

You should now have both video files inside the same RV session.

- [X] Get only the name of the file and compare it to the tag name in RvPusher::buildFileList()
- [X] Fix initializer order that didn't match the declaration order
- [X] Run Clang-Format

### DIFF
--- a/src/bin/apps/rvpush/RvPusher.cpp
+++ b/src/bin/apps/rvpush/RvPusher.cpp
@@ -31,11 +31,11 @@ using namespace std;
 RvPusher::RvPusher(QCoreApplication &app, string tag, string cmd, vector<string> argv, QObject *parent) :
         QObject(parent),
         m_app(app),
-        m_tag(tag),
+        m_client(0),
         m_command(cmd),
+        m_tag(tag),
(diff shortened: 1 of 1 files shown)


geffrak (approved):
Thank you for the fix, and thank you for putting the clang-format run in an isolated commit :D 

Please make sure it's properly tested, I can't tell that everything is fixed just by looking at the code.

bernie-laberge (approved):
LGTM
Note that I tried (unsuccessfully) to break your fix by using multiple underscores to the tag name like
./rvpush -tag "tag_name_1" merge "clip1.mov"
But your fix is solid. 
Thank you Éloïse !

PR #266, opened 2023-09-23
Code to make sure when using rv in windows, that the app does not block

### LINKED ISSUE #265: Shell launching with rv.exe
### Discussed in https://github.com/AcademySoftwareFoundation/OpenRV/discussions/247

<div type='discussions-op-text'>

<sup>Originally posted by **dkeraudren** August 29, 2023</sup>
When launching open rv, a shell window is launched along it ?
Is there a way to prevent that from happening ?</div>

richardssam:
### Cmake tweak to make sure that windows shell does not open when rv opens.

This is a minor change to cmake to treat the resulting application as a windows application, so that it does not block the console output.

### Linked issues

Fixes #265

### Summarize your change.
As above.

Getting a console window for every launch is really annoying.

Only tested on windows, but the cmake script shouldnt run on other operating systems.


geffrak (approved):
Thanks for the PR @richardssam the fix makes sense to me.

src/bin/apps/rv/CMakeLists.txt
```diff
@@ -45,6 +45,13 @@ ADD_EXECUTABLE(
   ${SOURCES} ${_rv_qrc}
 )
 
+IF( RV_TARGET_WINDOWS)
+  if (CMAKE_BUILD_TYPE STREQUAL "Release")
```
bernie-laberge, line 49: I guess you could even remove this Debug/Release condition and set it no matter what. 
As you wish.
Thanks!

bernie-laberge (approved):
Good catch !
Thank you for the fix !

src/bin/apps/rv/CMakeLists.txt
richardssam: I wasnt sure if having it enabled for debug mode would stop it from reliably running unit tests, or being run in a debug sesson.

tbabiin:
Hello, thank you for the PR! Could you run cmake-format in your workspace? There's a few issues in your formatting for the code that you added. The formatting apps are installed via 'rvsetup' or 'python -m pip install -r requirements.txt'. In requirements.txt, it's the pre-commit app that must be installed. Note: In .pre-commit-config.yaml, there's a hardcoded python version, if you have a mismatch version, you could disable the pre-commit hook and run cmake-format manually.

src/bin/apps/rv/CMakeLists.txt
bernie-laberge: Good point ! I just tested it and when launched from an MSYS2 shell, the reporting of the info and error messages is delayed until RV exits with your fix. So let's keep your original code then (with the Release check).
Thanks!

PR #267, opened 2023-09-24
Fix download_name for python3

paulbarton90:
### Summarize your change.
Updates `DOWNLOAD_NAME` to use `_python3_version` as `_version` isn't declared in this file.

The downloaded package ended up with the wrong version string. In initial testing this was was the version from OpenSSL.

Tested that it builds in a virtual machine for Centos 7


bernie-laberge (approved):
Great catch !
Thank you for the fix !

geffrak (approved):

PR #271, opened 2023-09-27
Fix segmentation fault if OCIO is not set

eloisebrosseau:
### Summarize your change.

The idea is to only call _OpenColorIO_v2_2::ColorSpace::getName()_ if the OCIO environment variable is set.

Using the new OCIOv2 system, if a user uses rvio to export a .rv session file in which some OCIO configs were applied clips, everything works as expected as long as the OCIO environment variable is detected in the terminal in which rvio is used. However, if the OCIO environment variable is not set, the tool will reach a segmentation fault. This is a regression. Using a past release of RV (i.e 2023.0.1), the tool would export the content from the .rv session file but would just not apply the OCIO changes if it couldn't find the environment variable.

The bug was tested on MacOS.

1. Download the attached zip file containing OCIO configs from [here](https://jira.autodesk.com/browse/SG-32970)
2. Install the latest version of RV using the OCIOv2 system
3. Open a terminal and make sure the OCIO environment variable is set and points on a valid OCIO config (i.e: ../aces_1.0.3/config.ocio)
4. Launch RV, then import some clips and make sure to apply an OCIO LUT on them
5. File / Save Session As...: Save the media and context into a .rv session file
6. Close RV, then open another terminal and go into the "bin" folder of the RV installed on step 2
7. Make sure to remove the OCIO environment variable from the terminal
8. Use the rvio tool to export the content of the .rv session file saved on step 5 (i.e: ./rvio /mnt/nvme/myOCIOSession.rv -o /mnt/nvme/myExport.mov)

You should now get the same behaviour as before (i.e. the rvio tool wil export the content in that context, but without applying OCIO).

- [X] Call  _OpenColorIO_v2_2::ColorSpace::getName()_ if the variable is set, output an error message otherwise.

### DIFF
--- a/src/lib/ip/OCIONodes/OCIOIPNode.cpp
+++ b/src/lib/ip/OCIONodes/OCIOIPNode.cpp
@@ -14,6 +14,7 @@
 #include <OpenColorIO/OpenColorIO.h>
 #include <boost/functional/hash.hpp>
 #include <algorithm>
+#include <cstdlib>
 #include <regex>
 
 namespace OCIO = OCIO_NAMESPACE;
@@ -169,7 +170,13 @@ OCIOIPNode::updateConfig()
 
     m_state->display  = m_state->config->getDefaultDisplay();
     m_state->view     = m_state->config->getDefaultView(m_state->display.c_str());
-    m_state->linear   = m_state->config->getColorSpace(OCIO::ROLE_SCENE_LINEAR)->getName();
+    
+    if (getenv("OCIO")) {
+        m_state->linear   = m_state->config->getColorSpace(OCIO::ROLE_SCENE_LINEAR)->getName();
+    } else {
+        std::cerr << "ERROR: OCIO environment variable not set" << '\n';
+    }
+    
     m_state->shaderID = "";
     m_state->lutID    = "";
 
--- a/src/lib/ip/OCIONodes/OCIOIPNode.cpp
+++ b/src/lib/ip/OCIONodes/OCIOIPNode.cpp
@@ -171,11 +171,14 @@ OCIOIPNode::updateConfig()
     m_state->display  = m_state->config->getDefaultDisplay();
     m_state->view     = m_state->config->getDefaultView(m_state->display.c_str());
     
-    if (getenv("OCIO")) {
+    if (getenv("OCIO"))
+    {
         m_state->linear   = m_state->config->getColorSpace(OCIO::ROLE_SCENE_LINEAR)->getName();
-    } else {
+    }
+    else
+    {
         m_state->linear   = "";
-        std::cerr << "ERROR: OCIO environment variable not set" << '\n';
+        std::cerr << "ERROR: OCIO environment variable not set" << std::endl;
     }
     
     m_state->shaderID = "";
--- a/src/lib/ip/OCIONodes/OCIOIPNode.cpp
+++ b/src/lib/ip/OCIONodes/OCIOIPNode.cpp
@@ -174,6 +174,7 @@ OCIOIPNode::updateConfig()
     if (getenv("OCIO")) {
         m_state->linear   = m_state->config->getColorSpace(OCIO::ROLE_SCENE_LINEAR)->getName();
     } else {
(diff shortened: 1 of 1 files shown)


geffrak (approved):

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -169,7 +170,13 @@ OCIOIPNode::updateConfig()
 
     m_state->display  = m_state->config->getDefaultDisplay();
     m_state->view     = m_state->config->getDefaultView(m_state->display.c_str());
-    m_state->linear   = m_state->config->getColorSpace(OCIO::ROLE_SCENE_LINEAR)->getName();
+    
+    if (getenv("OCIO")) {
+        m_state->linear   = m_state->config->getColorSpace(OCIO::ROLE_SCENE_LINEAR)->getName();
+    } else {
+        std::cerr << "ERROR: OCIO environment variable not set" << '\n';
```
bernie-laberge, line 177: Detail: I think it would be safer to initialize m_state->linear = "" here when the OCIO env var could not be found just in case.
I checked the original OCIOv1 code and it was returning an empty string ("") in that case.

eloisebrosseau: Done!

bernie-laberge: Beautiful ! Thanks !

bernie-laberge (approved):
GG

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -169,7 +170,14 @@ OCIOIPNode::updateConfig()
 
     m_state->display  = m_state->config->getDefaultDisplay();
     m_state->view     = m_state->config->getDefaultView(m_state->display.c_str());
-    m_state->linear   = m_state->config->getColorSpace(OCIO::ROLE_SCENE_LINEAR)->getName();
+    
+    if (getenv("OCIO")) {
+        m_state->linear   = m_state->config->getColorSpace(OCIO::ROLE_SCENE_LINEAR)->getName();
+    } else {
+        m_state->linear   = "";
+        std::cerr << "ERROR: OCIO environment variable not set" << '\n';
```
tbabiin, line 178: Minor: replace '\n' by std::endl.

tbabiin (approved):

src/lib/ip/OCIONodes/OCIOIPNode.cpp
eloisebrosseau: Is this a convention the team agreed on? Or do you prefer it that way because that's how you did it elsewhere in the file? I don't really get why we would want to flush the buffer on each output instead of letting the buffer manage himself.

tbabiin: Using the standard is preferable in terms of maintenance. Also in terms of maintenance, most of the code I've seen in RV uses std::endl when outputting to cout/cerr hence using the same style makes maintenance easy. 66% of our work as Devs - probably 80% as a contractor for myself - is maintenance instead of new code; thus making our life easier for maintenance is great.
About flushing every time: that would be a optimization decision and for outputting here, I don't think this is a optimization tuning situation.
And for code review: When I say 'minor', it means that it is optional and I don't want to hold off the PR if you need to merge quickly. (hence why I approved)

bernie-laberge: If I may: Since that this is an error, I would vote for flushing right away : std::endl
So I second Inouk's motion. :-)

bernie-laberge: Thank you !

eloisebrosseau: From my understanding, the buffer should flush anyway with \n since it's a std::cerr. You're just letting it decide to flush instead of telling it to flush knowing that it will flush anyway, but the result should be the same:)

It's fixed!

PR #272, opened 2023-09-27
Automatically install rebuilt packages

geffrak:
### Linked issues

N/A

### Summarize your change.

When rebuilding a rvpkg, clear the rvinstall file so the builds know that the rvpkg install step is dirty.

I lost time expecting the newly built rvpkg content to be automatically added to the staged app.

I tested it on MacOSX, but it's not platform-specific.

N/A

N/A

### DIFF
--- a/cmake/macros/rv_stage.cmake
+++ b/cmake/macros/rv_stage.cmake
@@ -462,6 +462,7 @@ FUNCTION(rv_stage)
       OUTPUT ${_package_filename}
       DEPENDS ${_temp_file} ${_files} ${_package_file}
       COMMAND ${CMAKE_COMMAND} -E tar "cfv" ${_package_filename} --format=zip --files-from=${_temp_file}
+      COMMAND ${CMAKE_COMMAND} -E remove ${RV_STAGE_PLUGINS_PACKAGES_DIR}/rvinstall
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):

cmake/macros/rv_stage.cmake
```diff
@@ -462,6 +462,7 @@ FUNCTION(rv_stage)
       OUTPUT ${_package_filename}
       DEPENDS ${_temp_file} ${_files} ${_package_file}
       COMMAND ${CMAKE_COMMAND} -E tar "cfv" ${_package_filename} --format=zip --files-from=${_temp_file}
+      COMMAND ${CMAKE_COMMAND} -E remove ${RV_STAGE_PLUGINS_PACKAGES_DIR}/rvinstall
```
tbabiin, line 465: remove is deprecated since 3.17. Use 'rm' instead and if you do not want to fail if the file doesn't exist (let's say on a clean build), use rm force aka 'rm -f'.
CMAKE Doc for -E commands: https://cmake.org/cmake/help/latest/manual/cmake.1.html

tbabiin (approved):
Looks good but don't use deprecated code.

PR #274, opened 2023-09-28
VFX 2023: Updating Python short version number

tbabiin:
### Linked issues
N/A

### Summarize your change.
Fixing the build since OCIO looks for python3.9

The build is broken

Windows, Mac & Linux

- Refactoring the code to avoid code duplication and sync'ing the short number with all other version numbers in this file

### DIFF
--- a/cmake/dependencies/python3.cmake
+++ b/cmake/dependencies/python3.cmake
@@ -16,15 +16,27 @@ SET(_pyside2_target
     "RV_DEPS_PYSIDE2"
 )
 
+SET(PYTHON_VERSION_MAJOR
+    3
+)
+
+SET(PYTHON_VERSION_MINOR
+    10
+)
+
+SET(PYTHON_VERSION_PATCH
+    12
+)
+
 SET(_python3_version
-    "3.10.12"
+    "${PYTHON_VERSION_MAJOR}.${PYTHON_VERSION_MINOR}.${PYTHON_VERSION_PATCH}"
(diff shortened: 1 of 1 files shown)


tbabiin:
Note: This is a suggestion on how to fix, we can discuss other ways to remove some duplication or, I can just change the short variable to "3.10". Let me know which idea you like the most.

geffrak:
How does this fixes OCIO?

tbabiin:
> How does this fixes OCIO?

OCIO uses Python for bindings, Python API. We pass the python EXE path and the library (DLL/SO) to the OCIO ExternalPrj_Add build. Those are hint points for OCIO's CMake to "Find Python" (FindPython.cmake). Passing a bad path to a unexisting file will make OCIO fail to configure. Hence, Python39 isn't found hence the fix.

geffrak:
> > How does this fixes OCIO?
> 
> 
> 
> OCIO uses Python for bindings, Python API. We pass the python EXE path and the library (DLL/SO) to the OCIO ExternalPrj_Add build. Those are hint points for OCIO's CMake to "Find Python" (FindPython.cmake). Passing a bad path to a unexisting file will make OCIO fail to configure. Hence, Python39 isn't found hence the fix.

Oh! I missed the remaining 3.9 from the merge of main (and the OCIO feature branch) into the vfx2023 feature branch.

geffrak (approved):

PR #276, opened 2023-10-02
Fix OIIO download hash

bernie-laberge:
### Summarize your change.

Changed the download hash for the OpenImageIO's code.
We believe this change was caused by the fact that the tagged oiio sources are now hosted by ASWF.
From:
https://github.com/OpenImageIO/oiio/archive/refs/tags/v2.4.6.0.tar.gz
To:
https://github.com/AcademySoftwareFoundation/OpenImageIO/archive/refs/tags/v2.4.6.0.tar.gz

The OIIO download hash was changed under our feet for the exact same OpenImageIO version.

Successfully built tested on Mac OS


geffrak (approved):

tbabiin (approved):
Approved

PR #281, opened 2023-10-04
Improve error handling and debugging

rogernelson:
### Summarize your change.
This branch does three things when the user attempts to open media in OpenRV.  First, we will now attempt to push as many errors as possible back to the UI instead of just displaying the generic "Unsupported codec" error.  Often this error is inaccurate and something else is the actual problem (file doesn't exist, user doesn't have permission to read it, etc).  

We do this by throwing exceptions with `TWK_THROW_EXC_STREAM` whenever we can catch the error in the FileSourceIPNode or in codecs themselves.  The UI will catch this exception and display it to the user.  We will also run additional existence and permission checks prior to opening it in the FileSourceNode instead of assuming that anything past this point is a codec error.

While doing this, I discovered that `QtFileInfo::isWritable` and `QtFileInfo::isReadable` always return true on Windows unless special actions are taken.  From their [API docs](https://doc.qt.io/qt-6/qfileinfo.html#ntfs-permissions):

> On NTFS file systems, ownership and permissions checking is disabled by default for performance reasons. 

To work around this fact, I originally used the access system call often used on *nix systems, but it also has a pitfall on windows.  From the [Windows docs]:

> When used with directories, _access determines only whether the specified directory exists

So to workaround both these restrictions, we will use access except when we are testing directory permissions on Windows.  In that case, we will use Qt, setting and unsetting the special global variable used to test permissions and mimic the return values of access.

I also found many other places in the code where we were using `QtFileInfo::isWritable` and `QtFileInfo::isReadable`, so I created a new `TwkUtil::access` method with `isReadable` and `isWritable` helper functions and redirected all the calls to use those to solve the problem everywhere.

Last, but not least, it is useful to know that there is a `plugins` debugging category when debugging these problems, but there's no way know that exists without reading the code.  Because of this, I added the debugging categories to the help menu so the user can know what debugging options they have.

To help the user debug problems when opening media.

Linux Rocky 9, Mac OS 13.3, Windows 10


geffrak (approved):

bernie-laberge (approved):
This is Beautiful Roger.
Thank you !

PR #282, opened 2023-10-05
Fix build on Apple silicon Macs

bernie-laberge:
### Summarize your change.

- Added the CMAKE_OSX_ARCHITECTURES to the spdlog.cmake configure which was missing.
- Instructed the OpenImageIO build NOT to use libfreetype: It is NOT used in commercial RV 2023 and early and using it makes the apple silicon build tries to link with the system (arm) libfreetype thus breaking the build. 
- Removed redundant DCMAKE_INSTALL_PREFIX in some cmake files.

To fix the build on Apple silicon Macs which had been recently broken.

Successfully build tested on Mac M1


cmake/macros/rv_stage.cmake
```diff
@@ -96,16 +96,18 @@ FUNCTION(rv_stage)
...
-            dep_file_path
-            TARGET ${dep}
-            PROPERTY LOCATION
-          )
-          GET_FILENAME_COMPONENT(dep_file_name ${dep_file_path} NAME)
-          ADD_CUSTOM_COMMAND(
-            COMMENT "Fixing ${dep_file_name}'s rpath in ${arg_TARGET}" TARGET ${arg_TARGET} POST_BUILD
-            COMMAND ${CMAKE_INSTALL_NAME_TOOL} -change "${dep_file_path}" "@rpath/${dep_file_name}" "$<TARGET_FILE:${arg_TARGET}>"
-          )
+          IF(TARGET ${dep})
```
geffrak, line 99: Ah! Interesting! I wonder what started to cause this issue.

geffrak (approved):

cmake/macros/rv_stage.cmake
bernie-laberge: Exactly !

bernie-laberge: CMake gives an error because some targets do not exist yet.

geffrak: Is it when we add/remove targets by modifying the code or pulling the main branch?

rogernelson (approved):

cmake/dependencies/oiio.cmake
```diff
@@ -46,6 +46,7 @@ LIST(APPEND _configure_options "-DBUILD_THIRDPARTY=1")
 LIST(APPEND _configure_options "-DBUILD_TESTING=OFF")
 LIST(APPEND _configure_options "-DUSE_PYTHON=0") # this on would requireextra pybind11 package
 LIST(APPEND _configure_options "-DUSE_OCIO=0")
+LIST(APPEND _configure_options "-DUSE_FREETYPE=0")
```
tbabiin, line 49: Are you sure this is taken in account by the OIIO CMake Build? I know 100% that USE_OCIO=0 is unused by the OIIO CMake build.
EDIT: Either way: if it is merged, I will try for sure to see if it is used.

cmake/macros/rv_stage.cmake
tbabiin: > Is it when we add/remove targets by modifying the code or pulling the main branch?

It's this code that causes the 'doesn't exist yet error in OpenRV' in cmake/dependencies/CMakeLists.txt:

```
IF(NOT TARGET dependencies)
  ADD_CUSTOM_TARGET(dependencies)
ENDIF()

IF(NOT DEFINED RV_DEPS_LIST)
  SET(RV_DEPS_LIST
      ""
      CACHE INTERNAL ""
  )
ENDIF()
```
If you remove both IFs, it works. I don't understand why.

tbabiin (approved):
Looks good, I just added some comments for future discussions.

cmake/dependencies/oiio.cmake
bernie-laberge: Yes I can confirm it is taken into account as it solved the following build issue on apple silicon mac:
Linking CXX shared library ../../lib/libOpenImageIO.dylib-ld: warning: ignoring file /opt/homebrew/lib/libfreetype.dylib, building for macOS-x86_64 but attempting to link with file built for macOS-arm64

cmake/macros/rv_stage.cmake
bernie-laberge: I believe RV_DEPS_LIST needs to be a cache cmake variable because it can be set by a super project to list additional dependencies.

geffrak: Yeah, I don't think it should be a cache one. It should not cause any issue with the super project defining it first since it's in the same environment.

bernie-laberge: OK roger that: I'll make the change both here and in the super project.

bernie-laberge: As discussed  with geffrak, we propose to merge this PR as is and to replace the RV_DEPS_LIST's CACHE with PARENT_SCOPE in another PR.

PR #291, opened 2023-10-12
Create OpenEXR patch to fill invalid lines and tiles with black

rogernelson:
### Summarize your change.

This branch add a patch to OpenEXR after we download it.  The patch will fill any scanlines or tiles that it fails to read with black.

This was a feature that was lost when migrating from RV to OpenRV.  During the migration dependencies were removed directly from the repo and instead fetched via CMake's `ExternalProject_Add` functionality.  However, when replacing the OpenEXR code we were not aware it had been modified. 

The correct fix for this is to do it in the API directly (or use OIIO to read OpenEXR which already does exactly that). But until that can be done, we'll just take the changes from the old system and create a patch file from them which can be integrated into CMake's `PATCH_COMMAND` step with `ExternalProject_Add.

Mac M1 Ventura 13.1

### DIFF
--- a/cmake/dependencies/openexr.cmake
+++ b/cmake/dependencies/openexr.cmake
@@ -146,6 +146,10 @@ IF(RV_TARGET_WINDOWS)
   LIST(APPEND _openexr_byproducts ${_openexr_implib} ${_ilmthread_implib} ${_iex_implib})
 ENDIF()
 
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):
LGTM

geffrak (approved):

PR #293, opened 2023-10-13
Extra fixes on the RvConsoleWindow

geffrak:
### Linked issues

N/A

### Summarize your change.

Misc tweaks on the RvConsoleWindow so we don't lose the first characters of a line without type (ex: INFO: )

While debugging an issue I had working on a Mu package, I was having the first letters of my backtrace lines stripped. It's because they don't have the logger type prefix in their lines.

Tested on MacOS. 

N/A

Before:

image

After: 

image


src/lib/app/RvCommon/RvConsoleWindow.cpp
```diff
@@ -269,7 +269,7 @@ namespace Rv
     // filter this
     if( qtimerwarning || qfilesystemwatcher ) return false;
 
-    spdlog::level::level_enum lineLogLevel = spdlog::level::info;
+    static spdlog::level::level_enum lineLogLevel = spdlog::level::info;
```
geffrak, line 272: Setting it static allows it to stay for all the lines of a backstrace. 

A line without the log type prefix will use the same one as the one before.

src/lib/app/RvCommon/RvConsoleWindow.cpp
```diff
@@ -308,19 +308,19 @@ namespace Rv
 
     if( lineLogLevel == spdlog::level::err )
     {
-      html += "<font color=red>ERROR:</font> ";
+      html += "<font color=red>ERROR: </font> ";
```
geffrak, line 311: Formatting, because we are removing the space if there's any.

src/lib/app/RvCommon/RvConsoleWindow.cpp
```diff
@@ -355,49 +355,51 @@ namespace Rv
     // filter this
     if( qtimerwarning || qfilesystemwatcher ) return false;
 
-    spdlog::level::level_enum lineLogLevel = spdlog::level::info;
-
-    if( line.find( "ERROR:" ) == 0 )
```
geffrak, line 360: Merged with the code below.

bernie-laberge (approved):

tbabiin (approved):
:+1:

PR #294, opened 2023-10-13
VFX2023 merge to main

tbabiin:
### Linked issues
N/A

### Summarize your change.
Merging VFX2023 to main and adding a commit for fixing py-interp

RV is updating their third-party toolchains

Mac, Windows and Linux Rocky 8 and Rocky 9

- Added a commit to fix py-interp: removing Py_Initialize as it is called by Py_Main. Removing PySys_SetArgvEx as it doesn't work when removing Py_Initialize but the new PyConfig works to set Py[...]SetArgv.

### DIFF
--- a/src/build/make_pyside.py
+++ b/src/build/make_pyside.py
@@ -193,7 +193,7 @@ def remove_broken_shortcuts(python_home: str) -> None:
             if filename not in [
                 "python",
                 "python3",
-                "python3.9",
+                "python3.10",
             ]:
                 print(f"Removing {filepath}...")
                 os.remove(filepath)
--- a/src/build/make_pyside.py
+++ b/src/build/make_pyside.py
@@ -99,7 +99,7 @@ def prepare() -> None:
     elif system == "Linux":
         clang_filename_suffix = "80-based-linux-Rhel7.2-gcc5.3-x86_64.7z"
     elif system == "Windows":
-        clang_filename_suffix = "80-based-windows-vs2017_64.7z"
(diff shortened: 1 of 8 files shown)


tbabiin:
Only [46b8289](https://github.com/AcademySoftwareFoundation/OpenRV/pull/294/commits/46b82895bda4ba5338f8f6b1790944508ab80277) needs to be reviewed, the rest is a branch merge.

geffrak (approved):

src/build/make_pyside.py
```diff
@@ -222,7 +222,7 @@ def build() -> None:
     # PySide2 v5.15.2.1 builds with errors on Windows using Visual Studio 2019.
     # We force Visual Studio 2017 here to make it build without errors.
     if platform.system() == "Windows":
-        source_widows_msvc_env("2017")
+        source_widows_msvc_env("2022")
```
geffrak, line 225: Do we still need that now that it's the same version as the rest of the build? Not asking for a change here.

tbabiin: Maybe we could test after merging. The scenario I'm not sure about is what happens with a computer that has: (VS2017) + VS2019 + VS2022; if we don't source, which one does PySide decides to pick. 🤔

bernie-laberge (approved):

rogernelson (approved):

tbabiin:
Forced-pushed again for another rebase to fix a conflict in make_pyside: it wasn't a real conflict, only that one of the line that I changed had a remote change for the line just before thus, a conflict for a remote change on the previous line of my change. 🤷‍♂️

PR #297, opened 2023-10-16
Add patch tool to windows doc

rogernelson:
Fixes #295 

### Summarize your change.

This is a minor update to the Windows documentation to add patch to the list of pacman packages needed in mingw. We started using patch after generating the docs so we overlooked adding it when doing so.  

Make building on Windows less painful.

Windows 11


geffrak (approved):

docs/build_system/config_windows.md
```diff
@@ -69,6 +69,7 @@ pacman -S --needed \
         git \
         nasm \
         p7zip \
+        patch \
```
bernie-laberge, line 72: We also need to add the patch package to the linux documentation and remove it from commercial rv.

rogernelson: This is just the docs, it doesn't execute the command.  I will add it to Linux though.

bernie-laberge (approved):

PR #302, opened 2023-10-23
Add script variant of windows reg files for rvlink and rvpush

geffrak:
### Linked issues

N/A

### Summarize your change.

Provide a script that sets the required registry entries for rvpush and rvlink where the paths to RV tools are relative to the script.

The current .reg files doesn't work because we can't guess the values at build time. Without a proper installer, we can't update the files at install time. The solution I propose is to use a batch script that computes the path to rv and to rvpush. 

Tested on Windows 10 by dropping the new files in the etc folder of a pre-compiled OpenRV.

Please double check the entries. I *think* I did it right, but it's really error prone. 


geffrak:
@mchesnay, I added you so you can have a look at the changes to the README.

bernie-laberge (approved):

packages/rv/README.regfiles
```diff
@@ -1,18 +1,20 @@
+Each provided .reg have a REPLACE_ME_FOR_RV_HOME placeholder that have to be replaced for the final location of the app.
+We provide a .bat variant of the .reg files that sets the registry entries wile automatically fixing the REPLACE_ME_FOR_RV_HOME part.
```
mchesnay, line 2:
```suggestion
Also included is a .bat version of each .reg files. Each .bat file sets the registry entry and automatically updates REPLACE_ME_FOR_RV_HOME with the location of the application.
```

tbabiin (approved):
Looks good to me, the only minor thing is: blank lines are missing at the end of the batch files.

PR #303, opened 2023-10-24
Download Qt artifacts from the  University of California mirror

geffrak:
### Linked issues

N/A

### Summarize your change.

Avoid using Qt's main mirror since it's often really slow and unreliable.

I got the link from https://download.qt.io/static/mirrorlist/

CI and local builds are failing because of the unreliability of the official sources.

Tested on MacOS 

N/A

N/A

### DIFF
--- a/src/build/make_pyside.py
+++ b/src/build/make_pyside.py
@@ -37,7 +37,8 @@ QT_OUTPUT_DIR = ""
 PYTHON_OUTPUT_DIR = ""
 OPENSSL_OUTPUT_DIR = ""
 
-LIBCLANG_URL_BASE = "https://download.qt.io/development_releases/prebuilt/libclang/libclang-release_"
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):

src/build/make_pyside.py
```diff
@@ -37,7 +37,8 @@
 PYTHON_OUTPUT_DIR = ""
 OPENSSL_OUTPUT_DIR = ""
 
-LIBCLANG_URL_BASE = "https://download.qt.io/development_releases/prebuilt/libclang/libclang-release_"
+LIBCLANG_URL_BASE = "https://mirrors.ocf.berkeley.edu/qt/development_releases/prebuilt/libclang/libclang-release_"
```
rogernelson, line 40: Should we list both URLs here? Then if one fails, try the other?

rogernelson (approved):

src/build/make_pyside.py
geffrak: We could. Let's see how reliable this mirror is before.

PR #304, opened 2023-10-24
Use all cpu cores when building OpenRV

geffrak:
### Linked issues

N/A

### Summarize your change.

It uses python to get the current core count to decide how many build tasks should run in parallel.

As for the other variables, someone can set RV_BUILD_PARALLELISM before loading rvcmds.sh to override the value.

The previous number was hardcoded to 8. On my computer, the number should be 10 (M1 Pro). It allows us to build slightly faster by allowing 25% more parallelism. 

Tested on MacOS.

N/A

```
# geffrak @ ADSKGV4VL2YQGP in ~/code/git/github.com/org-geffrak/OpenRV on git:fast-build o [16:38:32] 
$ source rvcmds.sh
Please ensure you have installed any required dependencies from doc/build_system/config_[os]

CMake parameters:
RV_BUILD_PARALLELISM is 10
RV_HOME is /Users/geffrak/code/git/github.com/org-geffrak/OpenRV
RV_BUILD is /Users/geffrak/code/git/github.com/org-geffrak/OpenRV/_build
RV_INST is /Users/geffrak/code/git/github.com/org-geffrak/OpenRV/_install
CMAKE_GENERATOR is Ninja
QT_HOME is /Users/geffrak/Qt/5.15.14/macos
To override any of them do unset [name]; export [name]=value; source rvcmds.sh

If this is your first time building RV try rvbootstrap (release) or rvbootstrapd (debug)
To build quickly after bootstraping try rvmk (release) or rvmkd (debug)
```

```
# geffrak @ ADSKGV4VL2YQGP in ~/code/git/github.com/org-geffrak/OpenRV on git:fast-build o [16:39:16] 
$ rvbuild
Change Dir: '/Users/geffrak/code/git/github.com/org-geffrak/OpenRV/_build'

Run Build Command(s): /opt/homebrew/bin/ninja -v -j 10 main_executable
[...]
```

### DIFF
--- a/rvcmds.sh
+++ b/rvcmds.sh
@@ -43,14 +43,15 @@ fi
 RV_HOME="${RV_HOME:-$SCRIPT_HOME}"
 RV_BUILD="${RV_BUILD:-${RV_HOME}/_build}"
 RV_INST="${RV_INST:-${RV_HOME}/_install}"
+RV_BUILD_PARALLELISM="${RV_BUILD_PARALLELISM:-$(python3 -c 'import os; print(os.cpu_count())')}"
 
 # ALIASES: Basic commands
 
 alias rvsetup="python3 -m pip install --user --upgrade -r ${RV_HOME}/requirements.txt"
 alias rvcfg="cmake -B ${RV_BUILD} -G \"${CMAKE_GENERATOR}\" ${CMAKE_WIN_ARCH} -DCMAKE_BUILD_TYPE=Release -DRV_DEPS_QT5_LOCATION=${QT_HOME} -DRV_DEPS_WIN_PERL_ROOT=${WIN_PERL}"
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):

rvcmds.sh
```diff
@@ -43,14 +43,15 @@ fi
 RV_HOME="${RV_HOME:-$SCRIPT_HOME}"
 RV_BUILD="${RV_BUILD:-${RV_HOME}/_build}"
 RV_INST="${RV_INST:-${RV_HOME}/_install}"
+RV_BUILD_PARALLELISM="${RV_BUILD_PARALLELISM:-$(python3 -c 'import os; print(os.cpu_count())')}"
```
rogernelson, line 46: Why not just use `nproc --all`? Might be better too since it gives the number of CPUs available for the current process.

geffrak: Is `nproc` available on all platforms? I have it on my Mac, but it comes from Homebrew, so it's not native. I didn't checked on Windows/MSYS.

rogernelson (approved):

PR #308, opened 2023-10-27
VFX2023: Updating the documentation

tbabiin:
### Summarize your change.

Updating install Visual Studio instructions to build on Windows

Documentation points to VS2019 & 2017.

Build machines: RHEL7 and Rocky9, Windows, Mac

- Make_pyside: removing source call to 2022 since only 2022 needs to be installed on build machines.

### DIFF
--- a/docs/build_system/config_windows.md
+++ b/docs/build_system/config_windows.md
@@ -109,10 +109,10 @@ Symptoms of using Windows' (and not mingw64) CMake:
 
 ### NOTE: Before starting a build
 
-Ensure that:
-- Your codebase is fully updated to the latest version (or the version you want to use) via a command like `git pull`.
-- Fix all conflicts due to updating the code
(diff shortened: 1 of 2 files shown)


tbabiin:
I still don't know what happens if you build with VS2022 and you still have VS2019 installed (like for other projects). I tested on Jenkins but since it's the first run, I don't see the parameters but I'm pretty sure it was the 2022 agent.

mchesnay (commented):
Just a few suggestions to improve the flow of the bullet list

tbabiin:
@mchesnay I've updated the text, resolve the conversations if you are ok with my fixes. Thanks for your help!

mchesnay (approved):

bernie-laberge (approved):
Beautiful. Thanks!

PR #310, opened 2023-11-01
Fix OpenGL errors on rvio exit

rogernelson:
### Summarize your change.
This branch rvio errors regarding unknown sync objects.  It turned out the cause was that the MovieRV frame buffer objects used by `rvio` contain their own video device that contains its own OpenGL context, and `rvio` also sets up its own OpenGL device, which contains the current OpenGL context on shutdown.  When the FBOs are destructed they will attempt to delete but the wrong context is set causing the errors.

To fix this, we'll add an uninit method to the MovieRV objects (one for hardware rendering and one for software) that will set its device to the current context and then explicitly uninit any FBOs it allocated during write.

During the investigation I tried a few other things that turned out not to be the cause of the error but I kept them anyway since the are good practice.  This includes protecting the FBOs against move self-assignment, a memory leak if move-assigning an already allocated FBO, and explicitly deleting copy operations (its not strictly necessary, but I find it makes the intention more clear).

Lastly, there was a bug in uninitializing the FBO pool where it was setting the wrong boolean to false.  It wouldn't really matter unless the FBO was reinitialized (which is not the case anywhere in our code), but it doesn't hurt to fix it now in case it does happen in the future.

Note that we do not have this problem in OpenRV itself since the MovieRV objects are used exclusively by rvio.

Even though the error was on shutdown and the output would be successfully generated, any tool wrapping rvio and relying on its exit code to tell whether or not it successfully completed its rendering will abort assuming error due to the bad exit code.

Tested with `rvio [image sequence] -o /tmp/out.mov`

Mac M1 and Linux Rocky 9.


src/lib/graphics/TwkGLF/GLSyncObject.cpp
```diff
@@ -401,17 +407,21 @@ GLSyncObject::GLSyncObject()
...
 
 //------------------------------------------------------------------------------
 //
 GLSyncObject& GLSyncObject::operator=(GLSyncObject&& rhs)
 {
-   _imp = rhs._imp;
-   rhs._imp = nullptr;
+   if( this != &rhs )
+   {
+     delete _imp;
```
rogernelson, line 421: Fixes memory leak if _imp was non-null.

src/lib/graphics/TwkGLF/GLSyncObject.cpp
```diff
@@ -401,17 +407,21 @@ GLSyncObject::GLSyncObject()
...
    rhs._imp = nullptr;
 }
 
 //------------------------------------------------------------------------------
 //
 GLSyncObject& GLSyncObject::operator=(GLSyncObject&& rhs)
 {
-   _imp = rhs._imp;
-   rhs._imp = nullptr;
+   if( this != &rhs )
```
rogernelson, line 419: Protects against self-assignment

src/lib/graphics/TwkGLF/GLPixelBufferObjectPool.cpp
```diff
@@ -884,7 +885,7 @@ class PBOWrap
       {
          if ( gPoolFromGPUInitialized )
          {
-            gPoolToGPUInitialized = false;
```
rogernelson, line 887: Probably a cut and paste error from the above block.

src/lib/graphics/TwkGLF/TwkGLF/GLSyncObject.h
```diff
@@ -23,6 +23,9 @@ class GLSyncObject
    GLSyncObject& operator=(GLSyncObject&&);
    ~GLSyncObject();
 
+   GLSyncObject( const GLSyncObject& rhs ) = delete;
```
rogernelson, line 26: The declaration of the move operations above will implicitly delete these, but it doesn't hurt to explicitly communicate intent.

bernie-laberge (approved):

geffrak (approved):

PR #311, opened 2023-11-02
Add Info Strip Scaling Option

eloisebrosseau:
### Summarize your change.

The info strip is automatically adjusted to the display's resolution and a new info strip menu option is available to disable this new scaling and to revert to the legacy behaviour (when no scaling occurred). Note that this is both a Presentation Mode as well as a regular display issue, although the issue is more visible in Presentation Mode due to its larger resolution.

The info strip in unreadable in 4K.

This change was tested on MacOS.

1. Launch RV
2. Load 4K media
2. In RV/Preferences/Video: Select 4K video output resolution either with AJA or Blackmagic (or NDI if you are on OpenRV)
3. Enable Presentation Mode's info strip in the RV/View/PresentationSettings/InfoStrip menu
4. Enable Presentation Mode
5. Enable InfoStrip (F7)

To disable the info strip automatic scaling:
1. Right-click on the Info Strip
2. Click on "Scale with Resolution"

- [X] Set the text size based on the display's resolution by default.
- [X] Add an option to disable the info strip scaling. In that case, set the text size to 20 like before

### DIFF
--- a/src/lib/app/mu_rvui/HUD.mu
+++ b/src/lib/app/mu_rvui/HUD.mu
@@ -283,6 +283,7 @@ class: ImageInfo : Widget
 class: InfoStrip : Widget
 {
     bool _showFilename;
+    bool _scaleWithResolution;
 
     method: showFilename (EventFunc; bool useFilename)
     {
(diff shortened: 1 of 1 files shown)


geffrak (approved):

bernie-laberge (approved):
Superb !

rogernelson (approved):

PR #312, opened 2023-11-02
Fix “Streaming” Movie Media from Online Sources documentation

mchesnay:
### Linked issues
None.

### Summarize your change.
Documentation erroneously stated in [chapter thirteen of the user manual](https://aswf-openrv.readthedocs.io/en/latest/rv-manuals/rv-user-manual/rv-user-manual-chapter-thirteen.html#streaming-movie-media-from-online-sources) that HTTPS in media paths was only supported on macOS. This is wrong, since Linux and Windows also support this.

Also took care of some markdown and other formatting issues.

Improve documentation accuracy.

### DIFF
--- a/docs/rv-manuals/rv-user-manual/rv-user-manual-chapter-thirteen.md
+++ b/docs/rv-manuals/rv-user-manual/rv-user-manual-chapter-thirteen.md
@@ -10,12 +10,11 @@ The reference manual has documentation about the networking protocol and how to 
 
 While it's possible to create connections across the internet, it's a little tricky, see [13.1.4](#1314-networking-across-the-internet) .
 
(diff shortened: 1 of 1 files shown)


mchesnay:
@geffrak  I deleted the line that said HTTPS was only supported on macOS. But I'm having second thoughts now. Maybe instead I should replace that line with the following:
* “https” (SSL) URLs should be used on macOS, Linux, and Windows.
What do you think?

rogernelson (approved):

geffrak:
> @geffrak I deleted the line that said HTTPS was only supported on macOS. But I'm having second thoughts now. Maybe instead I should replace that line with the following:
> 
> * “https” (SSL) URLs should be used on macOS, Linux, and Windows.
>   What do you think?

No, we don't care about ssl or not. It's up to them.

geffrak (approved):

PR #316, opened 2023-11-08
Fix the build on Rocky8

geffrak:
### Summarize your change.

Uses the system OpenSSL on RHEL8 while building OpenRV.

RHEL8 uses the same version of OpenSSL than the current OpenRV, and they enter in conflict during the build. There were a bit of documentation about how to avoid the conflict, but this is implementing a real fix.

Tested on Rocky8, and on Autodesk's RV CI to make sure nothing breaks with this change.

With this change, we no longer package OpenSSL in RV if building on a Rocky8. CentOS7 provides a openssl11 package that gives OpenSSL 1.1.1 and Rocky9 provides a compat-openssl package.

N/A


src/lib/files/WFObj
geffrak, line 1: This is for the .gitignore bit. It wasn't working on case sensitive file systems.

cmake/dependencies/python3.cmake
geffrak, line 1: Only specify openssl-dir if homemade

docs/build_system/config_linux_rocky8.md
geffrak, line 1: No need for this hack anymore.

src/build/make_pyside.py
geffrak, line 1: make OPENSSL_OUTPUT_DIR optional

bernie-laberge (approved):
Superb !

rogernelson (approved):

PR #320, opened 2023-11-10
Disable autoEDL on RVSwitchNodes created for MMR

geffrak:
### Linked issues

N/A

### Summarize your change.

Disables the autoEDL property on the RVSwitchNode that get created as part of Multiple Media Representation. 

The default autoEDL property changes the behaviour the ImageRangeInfo() of a RVSwitchNode
https://github.com/AcademySoftwareFoundation/OpenRV/blob/main/src/lib/ip/IPBaseNodes/SwitchIPNode.cpp#L561-L572

Where the m_offset is set based on the autoEDL property https://github.com/AcademySoftwareFoundation/OpenRV/blob/main/src/lib/ip/IPBaseNodes/SwitchIPNode.cpp#L268

It's an issue for parent nodes that doesn't necessarily expects RVSwitchNodes, but RVSourceNodes instead.

For example, the result of this is really different if MMR is being used or not. 
https://github.com/AcademySoftwareFoundation/OpenRV/blob/main/src/lib/ip/IPBaseNodes/SequenceIPNode.cpp#L600

FYI, the same fix is applied in the OTIO reader. https://github.com/AcademySoftwareFoundation/OpenRV/blob/b62c9c50e4e480a818906f22fce83205fefda195/src/plugins/rv-packages/otio_reader/otio_reader.py#L487

Tested on MacOS, but this one liner is platform agnostic.

I *think* that's the only change required, but please tell me if there's another entrypoint for RVSwitchNodes created for MMR. 

*MMR OFF, the benchmark:*

``` 
reviewAppSequenceGroup_sequence : RVSequence (1)
{
    edl
    {
        int source = [ 0 1 0 ]
        int frame = [ 1 21 109 ]
        int in = [ 1001 963 0 ]
        int out = [ 1020 1050 0 ]
    }

    output
    {
        int size = [ 1920 1080 ]
        float fps = 25
        int interactiveSize = 1
        int autoSize = 1
    }

    mode
    {
        int autoEDL = 1
        int useCutInfo = 1
        int supportReversedOrderBlending = 1
    }

    composite
    {
        string inputBlendModes = [ ]
    }
}
```

*MMR ON, Without the fix:*
```
reviewAppSequenceGroup_sequence : RVSequence (1)
{
    edl
    {
        int source = [ 0 1 0 ]
        int frame = [ 1 21 109 ]
        int in = [ 1 1 0 ]
        int out = [ 20 88 0 ]
    }

    output
    {
        int size = [ 1920 1080 ]
        float fps = 25
        int interactiveSize = 1
        int autoSize = 1
    }

    mode
    {
        int autoEDL = 1
        int useCutInfo = 1
        int supportReversedOrderBlending = 1
    }

    composite
    {
        string inputBlendModes = [ ]
    }
}
``` 

*MMR ON, With the fix:*
```
reviewAppSequenceGroup_sequence : RVSequence (1)
{
    edl
    {
        int source = [ 0 1 0 ]
        int frame = [ 1 21 109 ]
        int in = [ 1001 963 0 ]
        int out = [ 1020 1050 0 ]
    }

    output
    {
        int size = [ 1920 1080 ]
        float fps = 25
        int interactiveSize = 1
        int autoSize = 1
    }

    mode
    {
        int autoEDL = 1
        int useCutInfo = 1
        int supportReversedOrderBlending = 1
    }

    composite
    {
        string inputBlendModes = [ ]
    }
}
```


src/lib/app/RvApp/RvGraph.cpp
```diff
@@ -367,6 +367,8 @@ RvGraph::setupSource(SourceIPNode* source, IPCore::SourceIPNode* mediaRepSisterS
             }
 
             switchGroup = new SwitchGroupIPNode(name+"_switchGroup", def, this);
+            switchGroup->switchNode()->setProperty<IntProperty>("mode.autoEDL", 0);
```
bernie-laberge, line 370: I am not sure we want to do this all the time. This might have nefarious effects.

For screening room , we disable autoEDL on the switch nodes when edls are specified:
https://git.autodesk.com/media-and-entertainment/rv/blob/7a8c05216ef2ad4760cbe2bfb1051599dfa8952a/src/plugins/packages/screening_room/shotgrid_review_app.mu#L1810

Question: Could you set the autoEDL only when doing this compare scenario ? 

What do you think Roger ?

geffrak: The reason I took that route is to make the RVSwitch node act like a virtual Source Group for the ranges.

rogernelson: I was actually the one that added this property because I had the same problem with OTIO import.  In that case we want to specify the in/out points according to the `media_reference` in the OTIO Clips.  Otherwise, autoEDL always applies unwanted offset to the clips.

My original intention was to try to find if there was a Sequence node input to the Switch nodes and use the autoEDL value from it. But I thought that was going at little too far for just OTIO.  But perhaps now that we have other use cases maybe its what we should do?

geffrak: > its what we should do?

The pull request content or @bernie-laberge's suggestion?

rogernelson: I meant my original intention -- if the SwitchNode is connected to a Sequence node, then we use the autoEDL property from the SequenceNode.

The point of the offset is to adjust to the same time (ie offset from the beginning of the clip) when you switch between inputs that do not all share the same source time.  So setting `autoEDL` off by default would cause sources to jump around in time (potentially before the beginning or after the end) when switching to them.

rogernelson: The reason this works in OTIO is because we are guaranteed that all clips share the same source time, since there is only one `source_range` per clip, even when it contains multiple `media_references`.

rogernelson (approved):

bernie-laberge (approved):

PR #326, opened 2023-11-22
Add a tag in the window title when not in release

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Based on the CMake's variable `RV_RELEASE_DESCRIPTION`, add a tag to the window title to identify if the current build is not a release build.

Display that the current build is not a release build in the window title.

Tested on Windows and Linux

n/a

**_Without setting RV_RELEASE_DESCRIPTION or RV_RELEASE_DESCRIPTION=RELEASE (no file loaded)_**
Untitled - Release

**_Without setting RV_RELEASE_DESCRIPTION or RV_RELEASE_DESCRIPTION=RELEASE (with a file loaded)_**
File loaded - Release

_**RV_RELEASE_DESCRIPTION=PRE-RELEASE (no file loaded)**_
Pre-release-openrv - untitled

_**RV_RELEASE_DESCRIPTION="Internal Beta" (with a file loaded)**_
Internal beta

### DIFF
--- a/src/plugins/rv-packages/window_title/window_title.mu
+++ b/src/plugins/rv-packages/window_title/window_title.mu
@@ -13,6 +13,8 @@ class: WindowTitle : MinorMode
 { 
     string _title;
     string _mystate;
+    string _titlePrefix;
+    bool _isRelease;
     qt.QTimer _playStopTimer;
     qt.QTimer _updateTimer;
 
@@ -41,15 +43,13 @@ class: WindowTitle : MinorMode
         if (_title != s) 
         {
             _title = s;
-            string releaseVariant = getReleaseVariant();
-            if (releaseVariant == "RELEASE")
+            if (_isRelease)
             {
                 setWindowTitle(_title);
             }
             else
             {
-                // Add the release variant as suffix.
-                setWindowTitle("(%s) %s" % (releaseVariant, _title));
+                setWindowTitle(_titlePrefix + _title);
             }
         }
     }
@@ -134,6 +134,14 @@ class: WindowTitle : MinorMode
 
         _title = "";
         _mystate = "unknown";
+
+        // Compute the title prefix.
+        _titlePrefix = getReleaseVariant();
+        _isRelease = _titlePrefix == "RELEASE";
+        if (!_isRelease)
+        {
+            _titlePrefix = "(%s) " % _titlePrefix;
+        }
 
         _playStopTimer = qt.QTimer(mainWindowWidget());
         _playStopTimer.setSingleShot(true);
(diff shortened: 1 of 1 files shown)


src/plugins/rv-packages/window_title/window_title.mu
```diff
@@ -40,8 +40,17 @@ class: WindowTitle : MinorMode
...
+            _title = s;
+            string releaseVariant = getReleaseVariant();
+            if (releaseVariant == "RELEASE")
+            {
+                setWindowTitle(_title);
+            }
+            else
+            {
+                // Add the release variant as suffix.
+                setWindowTitle("(%s) %s" % (releaseVariant, _title));
```
cedrik-fuoco-adsk, line 52: I don't have a strong opinion on the format. I can be changed if needed.

geffrak (approved):

src/plugins/rv-packages/window_title/window_title.mu
```diff
@@ -40,8 +40,17 @@ class: WindowTitle : MinorMode
     {
         if (_title != s) 
         {
-	    _title = s;
-	    setWindowTitle(_title);
+            _title = s;
+            string releaseVariant = getReleaseVariant();
```
bernie-laberge, line 44: This set() method is called on a per frame basis because it is hooked to the frame-changed event.
Mu strings are not super efficient. I've had to minimize their use for some performance issues in the past.
I would put the getReleaseVariant() == "RELEASE" comparison in the WindowTitle constructor below so that we call getReleaseVariant() and also do this string comparison once and set say a bool if something needs to be added to the title.

bernie-laberge: window_title_set_called_each_frame

src/plugins/rv-packages/window_title/window_title.mu
bernie-laberge: You could compute a _titlePrefix in the constructor and use something like this I think:
`setWindowTitle(_titlePrefix+_title)`

bernie-laberge: Note: Mu code is interpreted. So this is why it is not efficient like compiled code.

src/plugins/rv-packages/window_title/window_title.mu
cedrik-fuoco-adsk: Thanks for all the information! I'll make changes.

bernie-laberge (approved):
Superb ! Thanks

cedrik-fuoco-adsk:
@rogernelson Could you review this PR? It looks like I need 3 reviewers to be able to merge it.

PR #332, opened 2023-12-04
Prevent crash when exception originates from a Python plugin in -debug mu

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Using `linenum` value to assume that the sourceFileName is most likely empty or invalid. This seems to happen when the exception originates from a python plugin.

When the exception originates from a python plugin, there are no Mu::Node or Mu::AnnotatedNode that correspond to the python code because these are created when the MuLang code interprets all the Mu code. Therefore, there are no `sourceFileName`, `linenum `or `charnum` for the python code.

Tested on Windows

n/a

n/a


cedrik-fuoco-adsk:
There is one issue that I did not find the reason. The issue is that the exception string does not get printed in the console. The exception is caught [here in PyMySymbolType.cpp L375](https://github.com/AcademySoftwareFoundation/OpenRV/blob/ef9f2e1f4e18b4495027c6cea3f337924b1e7467/src/lib/app/PyTwkApp/PyMuSymbolType.cpp#L375), but there is nothing displayed in the console.

geffrak (approved):

bernie-laberge (approved):

cedrik-fuoco-adsk:
Added two related commits. The first one prints the error message that we have up until that point and the second commits rename PyOpenColorIO.pyd to PyOpenColorIO_d.pyd in debug on Windows.

PR #334, opened 2023-12-07
Add two color mgmt nodes to Open RV that used to be only available in RV

bernie-laberge:
### Linked issues
NA

A little bit of context is required here: missing until this commit in Open RV, the commercial version of RV had two special color management related nodes that some users were leveraging:

1. SYNLinearize: this node could be used to include some media file's embedded color transforms into RV's color management pipeline.
2. SYNDisplay: this node could be used to include an ICC monitor profile file into RV's color management pipeline.

However, these two nodes were implemented using an Autodesk's internally developed color management library.
With the aim of adding these two nodes to Open RV without having to open source this internally developed library, it was decided to replace their implementation with OCIOv2.
This is exactly what this commit contains.

### Summarize your change.

Note that with the aim of preventing our existing users from having to modify their existing scripts to leverage these two color management related nodes, we have maintained the exact same interface that was available in the commercial RV to leverage those nodes: this includes their node names (SYNLinearize, SYNDisplay) and their associated parameters. 

Two nodes were added to the RV nodes definitions:

- SYNLinearize
- SYNDisplay

Both nodes are instantiated using the same already existing OCIOIPNode, but with new specific functions: synlinearize and syndisplay. 

The original implementation of those nodes did not require an OCIO config to be specified by the client. 
With the aim of allowing existing users to leverage those nodes without an OCIO config, the OCIOIPNode now supports built-in OCIO config which takes into effect automatically when the OCIO environment variable hasn't been set by the user.

For the SYNLinearize node, the input transform can be specified in two ways: via a url or via a data array. When the input transform is specified via a data array, then you will notice that we store the data array in a temporary file before passing it to OCIOv2 as it is not currently trivial to make OCIOv2 load a color transform from memory. There is the OCIOv2's ConfigIOProxy that can be used for that purpose, however it proved to be a bit overkill for our needs, since we only wanted to overload the color transform, not having to allocate a special config.

Already Tested on Mac OS. But will also test on all platforms.

Note that to test those nodes, one can create an .rvrc.mu file in their home directory (do not forget the dot . at the beginning) with the following content:

[rvrc.mu.zip](https://github.com/AcademySoftwareFoundation/OpenRV/files/13603419/rvrc.mu.zip)

### DIFF
--- a/src/lib/ip/OCIONodes/CMakeLists.txt
+++ b/src/lib/ip/OCIONodes/CMakeLists.txt
@@ -10,9 +10,13 @@ SET(_target
     "OCIONodes"
 )
 
-FIND_PACKAGE(Qt5 REQUIRED COMPONENTS Core)
+FIND_PACKAGE(
+  Qt5 REQUIRED
+  COMPONENTS Core
+)
 
-FILE(GLOB _sources *.c*)
+FILE(GLOB _sources *.cpp)
+
 ADD_LIBRARY(
   ${_target} STATIC
   ${_sources}
@@ -25,8 +29,8 @@ TARGET_INCLUDE_DIRECTORIES(
 
 TARGET_LINK_LIBRARIES(
   ${_target}
-  PUBLIC IPCore TwkFB
-  PRIVATE ocio::ocio TwkUtil Qt5::Core
+  PUBLIC IPCore TwkFB ocio::ocio
+  PRIVATE TwkUtil Qt5::Core
 )
 
 IF(RV_TARGET_WINDOWS)
--- a/src/lib/ip/OCIONodes/ConfigIOProxy.cpp
+++ b/src/lib/ip/OCIONodes/ConfigIOProxy.cpp
@@ -1,0 +1,71 @@
+//
+// Copyright (C) 2023  Autodesk, Inc. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <OCIONodes/ConfigIOProxy.h>
+#include <TwkExc/Exception.h>
+#include <IPCore/IPNode.h>
+
+namespace IPCore {
+
+const std::string ConfigIOProxy::USE_IN_TRANSFORM_DATA_PROPERTY = "OCIOIPNode.inTransform.data.icc";
+
+std::string ConfigIOProxy::getConfigData() const
+{
+    std::ostringstream os;
+    OCIO::Config::CreateRaw()->serialize(os);
+    return os.str();
+}
+
+std::vector<uint8_t> ConfigIOProxy::getLutData(
+        const char * filepath) const
+{
+    std::vector<uint8_t> buffer;
+
+    // Load color transform from memory
+    if (std::string(filepath).find(ConfigIOProxy::USE_IN_TRANSFORM_DATA_PROPERTY)!=std::string::npos)
+    {
+        const TwkContainer::ByteProperty* inTransformData = m_node->property<TwkContainer::ByteProperty>("inTransform.data");
+        if (!inTransformData)
+        {
+            std::ostringstream os;
+            os << "Could not read property inTransform.data from : " << m_node->name();
+            TWK_THROW_EXC_STREAM(os.str().c_str());
+        }
+        const char* data     = static_cast<const char*>(inTransformData->rawData());
+        const size_t dataSize = inTransformData->size(); // don't assume null terminated (ICC is not text)
+        buffer.resize(dataSize);
+        std::memcpy(buffer.data(), data, dataSize);
+    }
+    // Load color transform from file
+    else
+    {
+        std::ifstream fstream(
+            filepath, 
+            std::ios_base::in | std::ios_base::binary | std::ios::ate
+        );
+        if (fstream.fail())
+        {
+            std::ostringstream os;
+            os << "Could not read LUT file : " << filepath;
+            TWK_THROW_EXC_STREAM(os.str().c_str());
+        }
+
+        const auto eofPosition = static_cast<std::streamoff>(fstream.tellg());
+        buffer.resize(eofPosition);
+        fstream.seekg(0, std::ios::beg);
+        fstream.read(reinterpret_cast<char*>(buffer.data()), eofPosition);
+    }
+
+    return buffer;
+}
+
+std::string ConfigIOProxy::getFastLutFileHash(const char * filename) const
+{
+    return std::string(filename);
+}
+
+
+} // IPCore
--- a/src/lib/ip/OCIONodes/OCIOIPNode.cpp
+++ b/src/lib/ip/OCIONodes/OCIOIPNode.cpp
@@ -6,18 +6,20 @@
 //  
 //
 #include <OCIONodes/OCIOIPNode.h>
+#include <OCIONodes/ConfigIOProxy.h>
 #include <IPCore/SessionIPNode.h>
 #include <IPCore/NodeDefinition.h>
 #include <IPCore/IPGraph.h>
 #include <IPCore/ShaderCommon.h>
+#include <TwkExc/Exception.h>
 #include <TwkUtil/EnvVar.h>
-#include <OpenColorIO/OpenColorIO.h>
+
 #include <boost/functional/hash.hpp>
+
 #include <algorithm>
 #include <cstdlib>
+#include <sstream>
 #include <regex>
-
-namespace OCIO = OCIO_NAMESPACE;
 
 namespace IPCore {
 using namespace std;
@@ -71,7 +73,8 @@ OCIOIPNode::OCIOIPNode(const string& name,
                        IPGraph* graph,
                        GroupIPNode* group)
     : IPNode(name, def, graph, group),
-      m_lutfb(0)
+      m_lutfb(0),
+      m_useRawConfig(false)
 {
     Property::Info* info = new Property::Info();
     info->setPersistent(false);
@@ -93,10 +96,20 @@ OCIOIPNode::OCIOIPNode(const string& name,
 
     declareProperty<StringProperty>("ocio_display.display", "");
     declareProperty<StringProperty>("ocio_display.view", "");
-    //declareProperty<FloatProperty>("ocio_display.exposure");
-    //declareProperty<FloatProperty>("ocio_display.gamma");
-    //declareProperty<IntProperty>("ocio_display.channels");
+
+    if (func == "synlinearize")
+    {
+        m_inTransformURL = declareProperty<StringProperty>("inTransform.url", "", info);
+        m_inTransformData = declareProperty<ByteProperty>("inTransform.data", info);
+        m_useRawConfig = true;
+    }
     
+    if (func == "syndisplay")
+    {
+        m_outTransformURL = declareProperty<StringProperty>("outTransform.url", "", info);
+        m_useRawConfig = true;
+    }
+
     //
     //  Read-only properties to report config info to the user
     //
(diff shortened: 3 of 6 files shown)


src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -10,20 +10,26 @@
 #include <IPCore/NodeDefinition.h>
 #include <IPCore/IPGraph.h>
 #include <IPCore/ShaderCommon.h>
+#include <TwkExc/Exception.h>
 #include <TwkUtil/EnvVar.h>
-#include <OpenColorIO/OpenColorIO.h>
```
geffrak, line 14: We should keep it, no?

geffrak (approved):

src/lib/ip/OCIONodes/OCIOIPNode.cpp
bernie-laberge: Yes you are totally right but I just moved it to the header because I needed it.

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -251,6 +265,16 @@ shaderAdd3DLutAsParameter(std::string& inout_glsl, const std::string& lutSampler
...
 
+const std::string& OCIOIPNode::getWorkingSpaceTransformFilePath()
+{
+    if (!m_workingSpaceTransformFile)
+    {
+        m_workingSpaceTransformFile = std::make_shared<ColorTransformTmpFile>("XYZ_to_709_pri.ctf", XYZ_to_709_pri_ctf, strlen(XYZ_to_709_pri_ctf));
+    }
+
+    return m_workingSpaceTransformFile->getFilePath();
+}
```
doug-walker, line 276: Rather than adding this extra file to the repo, you could just declare the matrix in the code as follows (and I improved the precision of the values relative to what was in the file):

```
    OCIO::MatrixTransformRcPtr xyz2rec709_matrix = OCIO::MatrixTransform::Create();
    double m44[16] = { 3.240969941905, -1.537383177570, -0.498610760293, 0, 
                      -0.969243636281, 1.875967501508, 0.041555057407, 0, 
                       0.055630079697, -0.203976958889, 1.056971514243, 0, 
                       0, 0, 0, 1 };
    xyz2rec709_matrix->setMatrix(m44);
```

Append it to the GroupTransform below (or set the inverse direction) the same way as you do with the FileTransform.

src/lib/ip/OCIONodes/CMakeLists.txt
```diff
@@ -5,14 +5,22 @@
...
 
-FIND_PACKAGE(Qt5 REQUIRED COMPONENTS Core)
+FIND_PACKAGE(
+  Qt5 REQUIRED
+  COMPONENTS Core
+)
+
+FILE(GLOB _sources *.cpp)
+
+QUOTE_FILE(_ctf_file INPUT_FILENAME "XYZ_to_709_pri_ctf.ctf" CPP_SYMBOL "XYZ_to_709_pri_ctf")
```
doug-walker, line 21: I think you could remove this, please see below.

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -10,20 +10,26 @@
...
+//  This is an embedded CTF file that goes from CIE-XYZ to 709 linear.
+//
+extern const char* XYZ_to_709_pri_ctf;
 
 namespace IPCore {
 using namespace std;
 using namespace boost;
 
 static ENVVAR_INT( evOCIOLut3DSize, "RV_OCIO_3D_LUT_SIZE", 32 );
+static ENVVAR_STRING( evOCIOBuiltInConfig, "RV_OCIO_BUILT_IN_CONFIG", "studio-config-v1.0.0_aces-v1.3_ocio-v2.1" );
```
doug-walker, line 32: In OCIO 2.3 or higher, you could use "ocio://studio-config-latest" in CreateFromFile, or "studio-config-latest" in CreateFromBuiltin.

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -71,7 +77,8 @@ OCIOIPNode::OCIOIPNode(const string& name,
                        IPGraph* graph,
                        GroupIPNode* group)
     : IPNode(name, def, graph, group),
-      m_lutfb(0)
+      m_lutfb(0),
+      m_useBuiltInConfig(!getenv("OCIO"))
```
doug-walker, line 81: This could be dangerous, please see comment below.

src/lib/ip/OCIONodes/OCIOIPNode.cpp
bernie-laberge: Fantastic ! 
Thank you so much Doug !

src/lib/ip/OCIONodes/OCIOIPNode.cpp
bernie-laberge: RV is currently using OCIO v2.2.1.
But I will add a note in the code to make sure we revisit that code the next time we update OCIO in RV.

src/lib/ip/OCIONodes/OCIOIPNode.cpp
bernie-laberge: About the studio config : Good point ! In my original implementation, I was relying on the "Linear Rec.709 (sRGB)" color space which was only in the studio config if I am not mistaken. I don't need it anymore.

About the ConfigIOProxy: Excellent, I will try what you said but exclusively for the synlinearize+syndisplay implementations. Messing up with the client's OCIO config was making me nervous. 

My PR is getting shorter and shorter. I like it. Less code, less bugs. Less is more.
Thank you Doug !

doug-walker:
Here's a rough Python example of how to load a FileTransform from a buffer. 

```
    class CIOPTest(ocio.PyConfigIOProxy):
        def getConfigData(self):
            return ocio.Config.CreateRaw().serialize()

        def getLutData(self, filepath):
            with open(filepath, 'rb') as file_t:
                # This is your buffer.
                return bytearray(file_t.read())

        def getFastLutFileHash(self, filepath):
            import os
            if os.path.isfile(filepath):
                return filepath
            else:
                return ''

    ciop = CIOPTest()
    config = ocio.Config.CreateFromConfigIOProxy(ciop)
    config.validate()

    icc_path = '/usr/local/workspace/OpenColorIO-dw3/tests/data/files/icc-test-1.icc'

    ft = ocio.FileTransform(icc_path)
    processor = config.getProcessor(ft)
    cpu = processor.getDefaultCPUProcessor()
    print(cpu.applyRGB([0.1,0.2,0.4]))
```

bernie-laberge:
> Here's a rough Python example of how to load a FileTransform from a buffer. Once you have the Processor, you may dispose of the config.
> 
> ```
>     class CIOPTest(ocio.PyConfigIOProxy):
>         def getConfigData(self):
>             return ocio.Config.CreateRaw().serialize()
> 
>         def getLutData(self, filepath):
>             with open(filepath, 'rb') as file_t:
>                 # This is your buffer.
>                 return bytearray(file_t.read())
> 
>         def getFastLutFileHash(self, filepath):
>             import os
>             if os.path.isfile(filepath):
>                 return filepath
>             else:
>                 return ''
> 
>     ciop = CIOPTest()
>     config = ocio.Config.CreateFromConfigIOProxy(ciop)
> 
>     icc_path = '/usr/local/workspace/OpenColorIO/tests/data/files/icc-test-1.icc'
> 
>     ft = ocio.FileTransform(icc_path)
>     processor = config.getProcessor(ft)
>     cpu = processor.getDefaultCPUProcessor()
>     print(cpu.applyRGB([0.1,0.2,0.4]))
> ```

Thank you for the example Doug ! Much appreciated.
I am actually working on it but the difficulty here is that I don't have a filepath but some data array in memory which I have to figure out how to connect with the OCIO::ConfigIOProxy derived class to be able to fulfill the     inline std::vector<uint8_t> getLutData(const char * filepath) const override.

src/lib/ip/OCIONodes/OCIOIPNode.cpp
bernie-laberge: It looks like OCIOv2 still considers the color transform as a real file as I am called multiple times to get the LutData unless I name the color transform with a known extension:
!!!ConfigIOProxy-getFastLutFileHash()-filename=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data
!!!ConfigIOProxy-getLutData()-filepath=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data
!!!ConfigIOProxy-getLutData()-filepath=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data
!!!ConfigIOProxy-getLutData()-filepath=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data
!!!ConfigIOProxy-getLutData()-filepath=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data
!!!ConfigIOProxy-getLutData()-filepath=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data
!!!ConfigIOProxy-getLutData()-filepath=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data
!!!ConfigIOProxy-getLutData()-filepath=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data
!!!ConfigIOProxy-getLutData()-filepath=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data
!!!ConfigIOProxy-getLutData()-filepath=/var/tmp/tmp.0.rQFbky_OCIOIPNode.inTransform.data

bernie-laberge: Cédrik believes that it might be because OCIOv2 might go through all its fill parsers.

bernie-laberge:
@doug-walker : Thanks to the ConfigIOProxy example you sent me, I was able to completely get rid of the unnecessary crutch of having to persist a temporary color transform file to disk. Thank you Doug !

src/lib/ip/OCIONodes/ConfigIOProxy.cpp
```diff
@@ -0,0 +1,71 @@
...
+        fstream.seekg(0, std::ios::beg);
+        fstream.read(reinterpret_cast<char*>(buffer.data()), eofPosition);
+    }
+
+    return buffer;
+}
+
+std::string ConfigIOProxy::getFastLutFileHash(const char * filename) const
+{
+    return m_node->name()+std::string(filename);
```
doug-walker, line 67: I'm worried about always returning the same string when in buffer mode. This could cause OCIO to use it's file transform cache rather than calling getLutData. The hash should be unique for each unique buffer of ICC data that you want to load. Note that the file cache is global for all of OCIO, it is not part of the config. (Sorry, my Python example was too simplistic.)

bernie-laberge: OK good catch ! Thanks.
Note that the hash also contains the RV nodes' name (like "sourceGroup000000_tolinPipeline_0" for example), and I beliebe that the nodes's name are never the same for an RV process. I am currently confirming that.

bernie-laberge: Oh but that's not enough. The user could potentially decides to change the data transform. In which case we need a more unique name.

bernie-laberge: SYNCOLOR::loadColorTransformFromBuffer() was really nice ! :-)

bernie-laberge: I believe I can just add a simple counter. I'll make the change.

bernie-laberge: OK I made the change and it works ! 
You correctly predicted that OCIOv2 would reuse the cached color transform by the way.
The fix is in OCIOIPNode::evaluate():
                        `// We need to provide a unique name to OCIOv2, otherwise it might use a 
                        // potentially incorrect color transform with the same name already in its cache.
                        static int uniqueCounter = 0;
                        inTransformURL = name()+"."+std::to_string(uniqueCounter++)+"."+ConfigIOProxy::USE_IN_TRANSFORM_DATA_PROPERTY;`

doug-walker (approved):

cedrik-fuoco-adsk (approved):

PR #344, opened 2023-12-14
Add Aboutbox.rtf in RV package

eloisebrosseau:
### Summarize your change.

The Aboutbox.rtf file was added to the RV package. This is the one from RV 2023 since the new one isn't ready yet.

The addition of the file was tested on MacOS.


geffrak:
Can we have it once in the repository? With CMake, it's a lot easier than the old build system.

PR #347, opened 2023-12-21
Improve copy calls

geffrak:
### Linked issues

N/A

### Summarize your change.

Changes how we copy files in the build to avoid following the symlinks.

In the final app bundle of OpenRV, we had multiple copies of the duplicate files (dynamic libraries and others) because the symlinks were being resolved when copied. 

Build tested on: (WIP)
- [ ] Windows
- [ ] Linux
- [X] MacOS

N/A

N/A


CMakeLists.txt
```diff
@@ -89,8 +89,6 @@ SET(CMAKE_BUILD_TYPE
 
 MESSAGE(STATUS "Build type: ${CMAKE_BUILD_TYPE}")
 
-MESSAGE("PROJECT DIR ${PROJECT_SOURCE_DIR}")
```
geffrak, line 92: Left over from another changelist...

cmake/dependencies/atomic_ops.cmake
```diff
@@ -115,7 +115,7 @@ LIST(APPEND RV_DEPS_LIST atomic_ops::atomic_ops)
 ADD_CUSTOM_COMMAND(
   COMMENT "Installing ${_target}'s libs into ${RV_STAGE_LIB_DIR}"
   OUTPUT ${RV_STAGE_LIB_DIR}/${_atomic_ops_lib_name}
-  COMMAND ${CMAKE_COMMAND} -E copy_directory ${_lib_dir} ${RV_STAGE_LIB_DIR}
+  COMMAND ${CMAKE_COMMAND} -E env -- cp -Rfv "${_lib_dir}/" "${RV_STAGE_LIB_DIR}"
```
geffrak, line 118: -R recusive
-f deletes the target before copying
-v prints what get's copied

cmake/dependencies/boost.cmake
```diff
@@ -205,7 +205,7 @@ EXTERNALPROJECT_ADD(
     # Ref.: https://www.boost.org/doc/libs/1_70_0/tools/build/doc/html/index.html#bbv2.builtin.features.cflags Ref.:
     # https://www.boost.org/doc/libs/1_76_0/tools/build/doc/html/index.html#bbv2.builtin.features.cflags
     ./b2 -a -q toolset=${_toolset} cxxstd=${RV_CPP_STANDARD} variant=${_boost_variant} link=shared threading=multi architecture=x86 address-model=64
-    ${_boost_with_list} ${_boost_b2_options} -j${_cpu_count} install --prefix=${_install_dir}
+    ${_boost_with_list} ${_boost_b2_options} -j${_cpu_count} install --prefix=${_install_dir} define=_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION
```
geffrak, line 208: This is required so I can build on my Mac for a moment. I'm pushing it unless it breaks the other platforms on our standard build agents.

cmake/dependencies/qt5.cmake
```diff
@@ -67,91 +64,23 @@ SET(RV_DEPS_QT5_VERSION
...
-  FOREACH(
-    _qt_plugin_dir
-    ${_qt_plugins_dirs}
-  )
-    FILE(
-      COPY ${RV_DEPS_QT5_LOCATION}/plugins/${_qt_plugin_dir}
-      DESTINATION ${RV_STAGE_PLUGINS_QT_DIR}
-    )
-  ENDFOREACH()
+  EXECUTE_PROCESS(COMMAND ${CMAKE_COMMAND} -E env -- cp -Rnv "${RV_DEPS_QT5_LOCATION}/lib/" "${RV_STAGE_LIB_DIR}" )
```
geffrak, line 67: -n does not override the file if it's already there, so the configure step can stay fast if re-executed.

If you update Qt, you will have to wipe your _build folder anyways...

cmake/globals/rv_globals.cmake
```diff
@@ -37,6 +37,9 @@ IF(RV_TARGET_DARWIN)
       ${RV_APP_ROOT}/RV.app/Contents
       CACHE STRING "RV's build install root directory."
   )
+
+  EXECUTE_PROCESS(COMMAND ${CMAKE_COMMAND} -E create_symlink "MacOS" "bin"  WORKING_DIRECTORY ${RV_STAGE_ROOT_DIR})
```
geffrak, line 41: Helper, so there's a "bin" in MacOS. If we end up installing something in bin from a RV_DEPS, it will be in MacOS.

cmake/globals/rv_globals.cmake
```diff
@@ -73,10 +76,17 @@ ENDIF()
...
 
-SET(RV_STAGE_LIB_DIR
-    ${RV_STAGE_ROOT_DIR}/lib
-)
-FILE(MAKE_DIRECTORY ${RV_STAGE_LIB_DIR})
+IF(RV_TARGET_DARWIN)
+  SET(RV_STAGE_LIB_DIR
+          ${RV_STAGE_FRAMEWORKS_DIR}
+  )
+  EXECUTE_PROCESS(COMMAND ${CMAKE_COMMAND} -E create_symlink "Frameworks" "lib"  WORKING_DIRECTORY ${RV_STAGE_ROOT_DIR})
```
geffrak, line 83: On MacOS, dylibs should be located in Frameworks.

cmake/macros/rv_stage.cmake
```diff
@@ -85,7 +85,6 @@ FUNCTION(rv_stage)
         ADD_CUSTOM_COMMAND(
           COMMENT "Fixing ${arg_TARGET}'s RPATHs" TARGET ${arg_TARGET} POST_BUILD
           COMMAND ${CMAKE_INSTALL_NAME_TOOL} -add_rpath "@executable_path/../Frameworks" "$<TARGET_FILE:${arg_TARGET}>"
-          COMMAND ${CMAKE_INSTALL_NAME_TOOL} -add_rpath "@executable_path/../lib" "$<TARGET_FILE:${arg_TARGET}>"
```
geffrak, line 88: Stuff is installed in Framework, so this is not required. It's saving one call per executable/library!

bernie-laberge (approved):

cmake/dependencies/raw.cmake
```diff
@@ -65,9 +65,9 @@ IF(RV_TARGET_WINDOWS)
     TARGET ${_target}
     POST_BUILD
     COMMENT "Installing ${_target}'s libs & files into ${_install_dir}"
-    COMMAND ${CMAKE_COMMAND} -E copy_directory ${_base_dir}/src/lib ${_lib_dir}
-    COMMAND ${CMAKE_COMMAND} -E copy_directory ${_base_dir}/src/libraw ${_include_dir}/libraw
-    COMMAND ${CMAKE_COMMAND} -E copy_directory ${_base_dir}/src/bin ${_bin_dir}
+    COMMAND ${CMAKE_COMMAND} -E env -- cp -Rfv "${_base_dir}/src/lib/" "${_lib_dir}"
+    COMMAND ${CMAKE_COMMAND} -E env -- cp -Rfv "${_base_dir}/src/libraw/" "${_include_dir}/libraw"
+    COMMAND ${CMAKE_COMMAND} -E env -- cp -Rfv "${_base_dir}/src/bin/" "${_bin_dir}"
```
cedrik-fuoco-adsk, line 70: That's the Windows-only section. Shouldn't it be in the else?

geffrak: No, I changed all of the calls to make it simpler. Otherwise, it's easy to re-introduce copy_directory in a UNIX code path after refactoring.

geffrak:
cp directory/ destination doesn't behave the same on MacOS and on RHEL, closing the PR until I implement a fix

PR #350, opened 2023-12-22
Remove unwanted files in the install directory

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
I disabled some tools and apps that were built by a few dependencies and a few fixes in the variables that we use in our deps Cmake modules. (**_all platform_**)

I've also added more file filters when we copy files over the install directory for Qt. I filtered out the following extensions: `pdb`, `lib` and `exe` (except QtWebEngineProcess.exe because is it needed). (**_Windows only_**)

I used `windeployqt.exe` to determine which executables was needed, and only QtWebEngineProcess.exe showed up in the list.

The goal was to lighten the Windows package. These changes removed about 200MB from the zip file and about 800MB for the unzip folder (numbers are very rough).

Tested the Qt part on Windows because the Qt changes only affect Windows.
Tested the build on Rocky 8.

n/a

n/a


cmake/dependencies/oiio.cmake
```diff
@@ -67,7 +66,7 @@ ENDIF()
 GET_TARGET_PROPERTY(_png_library PNG::PNG IMPORTED_LOCATION)
 GET_TARGET_PROPERTY(_png_include_dir PNG::PNG INTERFACE_INCLUDE_DIRECTORIES)
 LIST(APPEND _configure_options "-DPNG_LIBRARY=${_png_library}")
-LIST(APPEND _configure_options "-DPNG_INCLUDE_DIR=${_png_include_dir}")
+LIST(APPEND _configure_options "-DPNG_PNG_INCLUDE_DIR=${_png_include_dir}")
```
cedrik-fuoco-adsk, line 69: OIIO was not able to find PNG because of the wrong include variable.

cmake/dependencies/oiio.cmake
```diff
@@ -89,14 +88,7 @@ GET_TARGET_PROPERTY(_openjpeg_include_dir OpenJpeg::OpenJpeg INTERFACE_INCLUDE_D
...
 
-GET_TARGET_PROPERTY(_tiff_library Tiff::Tiff IMPORTED_LOCATION)
-GET_TARGET_PROPERTY(_tiff_include_dir Tiff::Tiff INTERFACE_INCLUDE_DIRECTORIES)
-IF(NOT RV_TARGET_WINDOWS)
-  LIST(APPEND _configure_options "-DTIFF_LIBRARY=${_tiff_library}")
-ELSE()
-  LIST(APPEND _configure_options "-DTIFF_ROOT=${RV_DEPS_TIFF_ROOT_DIR}")
-ENDIF()
-LIST(APPEND _configure_options "-DTIFF_INCLUDE_DIR=${_tiff_include_dir}")
+LIST(APPEND _configure_options "-DTIFF_ROOT=${RV_DEPS_TIFF_ROOT_DIR}")
```
cedrik-fuoco-adsk, line 91: TIFF_ROOT works great here and is simpler on all platforms.

cmake/dependencies/openjpeg.cmake
cedrik-fuoco-adsk, line 1: Added some notes here because we are passing variables for ZLIB, PNG and TIFF but they are ignored by OpenJPEG. Because `BUILD_THIRDPARTY=1`, OpenJPEG builds all of them from the source. The versions used don't match what we are using outside of OpenJPEG.

We could set `BUILD_THIRDPARTY=0`, but that would require adding a new lcms2.cmake module and pass the variable to OpenJPEG. OpenJPEG build fine without LCMS2 (some minimal code is disabled because of that in OpenJPEG), but I'm not sure if it is critical to OIIO or not as OpenJPEG is only used by OIIO right now.

cmake/dependencies/webp.cmake
```diff
@@ -37,15 +37,19 @@ ELSE()
...
+ENDIF()
+GET_TARGET_PROPERTY(zlib_include_dir ZLIB::ZLIB INTERFACE_INCLUDE_DIRECTORIES)
+LIST(APPEND _configure_options "-DZLIB_INCLUDE_DIR=${zlib_include_dir}")
+LIST(APPEND _configure_options "-DZLIB_LIBRARY=${zlib_library}")
 
 GET_TARGET_PROPERTY(_png_library PNG::PNG IMPORTED_LOCATION)
 GET_TARGET_PROPERTY(_png_include_dir PNG::PNG INTERFACE_INCLUDE_DIRECTORIES)
 LIST(APPEND _configure_options "-DPNG_LIBRARY=${_png_library}")
-LIST(APPEND _configure_options "-DPNG_INCLUDE_DIR=${_png_include_dir}")
+LIST(APPEND _configure_options "-DPNG_PNG_INCLUDE_DIR=${_png_include_dir}")
```
cedrik-fuoco-adsk, line 52: Change to the correct PNG include variable

cmake/dependencies/webp.cmake
```diff
@@ -37,15 +37,19 @@ ELSE()
...
-LIST(APPEND _configure_options "-DZLIB_LIBRARY=${_zlib_library}")
-LIST(APPEND _configure_options "-DZLIB_INCLUDE_DIR=${_zlib_include_dir}")
+IF(RV_TARGET_WINDOWS)
+  GET_TARGET_PROPERTY(zlib_library ZLIB::ZLIB IMPORTED_IMPLIB)
+ELSE()
+  GET_TARGET_PROPERTY(zlib_library ZLIB::ZLIB IMPORTED_LOCATION)
+ENDIF()
+GET_TARGET_PROPERTY(zlib_include_dir ZLIB::ZLIB INTERFACE_INCLUDE_DIRECTORIES)
+LIST(APPEND _configure_options "-DZLIB_INCLUDE_DIR=${zlib_include_dir}")
+LIST(APPEND _configure_options "-DZLIB_LIBRARY=${zlib_library}")
```
cedrik-fuoco-adsk, line 47: Some method that we use for other dependencies on Windows.

cmake/dependencies/webp.cmake
```diff
@@ -89,8 +102,18 @@ EXTERNALPROJECT_ADD(
...
+  COMMENT "Installing ${_target}'s libs into ${RV_STAGE_LIB_DIR}"
+  OUTPUT ${RV_STAGE_LIB_DIR}/${_libname}
+  COMMAND ${CMAKE_COMMAND} -E copy_directory ${_lib_dir} ${RV_STAGE_LIB_DIR}
+  DEPENDS ${_target}
+)
+
+ADD_CUSTOM_TARGET(
+  ${_target}-stage-target ALL
+  DEPENDS ${RV_STAGE_LIB_DIR}/${_libname}
+)
```
cedrik-fuoco-adsk, line 116: Had to remove RV_COPY_LIB_BIN_FOLDERS() because there are no `bin` directory anymore because we do not build the tools and we link webp as a static library.

rvcmds.sh
```diff
@@ -55,7 +55,8 @@ alias rvbuildtd="cmake --build ${RV_BUILD} --config Debug -v --parallel=${RV_BUI
 alias rvbuild="rvbuildt main_executable"
 alias rvbuildd="rvbuildtd main_executable"
 alias rvtest="ctest --test-dir ${RV_BUILD} --extra=verbose"
-alias rvinst="cmake --install ${RV_BUILD} --prefix ${RV_INST}"
+alias rvinst="cmake --install ${RV_BUILD} --prefix ${RV_INST} --config Release"
+alias rvinstd="cmake --install ${RV_BUILD} --prefix ${RV_INST} --config Debug"
```
cedrik-fuoco-adsk, line 59: Had issue with using the `rvinst` alias because we weren't passing the build type here.

cmake/dependencies/oiio.cmake
```diff
@@ -119,6 +111,7 @@ LIST(APPEND _configure_options "-DFFMPEG_LIBAVFORMAT=${_ffmpeg_libavformat}")
 LIST(APPEND _configure_options "-DFFMPEG_LIBAVUTIL=${_ffmpeg_libavutil}")
 LIST(APPEND _configure_options "-DFFMPEG_LIBSWSCALE=${_ffmpeg_libswscale}")
 
+# Question: Do we need freetype at all? It is disabled above with USE_FREETYPE=0.
```
bernie-laberge, line 114: Good point ! 
In the OpenRV/cmake/dependencies/CMakeLists.txt:
`ADD_SUBDIRECTORY(../../src/pub/freetype src/pub/freetype) # required by oiio`
Since it is no longer required by OIIO, then perhaps there is some hope.
It would be really nice to remove it if it works.

bernie-laberge (approved):

cmake/dependencies/openjpeg.cmake
geffrak: Let's try with build_thirdparty=0

cmake/install/pre_install_windows.cmake
```diff
@@ -15,6 +15,28 @@ FUNCTION(before_copy_platform FILE_PATH RET_VAL)
     ENDIF()
   ENDIF()
 
+  # Export file contains information about symbols. Not needed for runtime.
+  IF(FILE_PATH MATCHES "\\.exp$")
+    IF(CMAKE_INSTALL_CONFIG_NAME MATCHES "^Release$")
```
geffrak, line 20: Unlike PDBs, we never want to install EXPs and LIBs.

cmake/dependencies/openjpeg.cmake
cedrik-fuoco-adsk: Without adding a cmake module to download and install `lcms2`?

geffrak: We can have the build output validated by QA

cmake/dependencies/png.cmake
```diff
@@ -33,7 +33,10 @@ ELSE()
   RV_MAKE_STANDARD_LIB_NAME("libpng16" "${_libpng_lib_version}" "SHARED" "d")
 ENDIF()
 # The '_configure_options' list gets reset and initialized in 'RV_CREATE_STANDARD_DEPS_VARIABLES'
+# Future: The main branch of libpng has deprecated 'PNG_EXECUTABLES' in favor of 'PNG_TOOLS'.
```
geffrak, line 36: Can you set PNG tools right now or it's breaking the build?

cedrik-fuoco-adsk: PNG_TOOLS is still used in the version that we are using (1.6.39) and in 1.6.40 as well. I assume that the change from PNG_TOOLS to PNG_EXECUTABLES will be in their next release.

cmake/dependencies/oiio.cmake
```diff
@@ -119,6 +111,7 @@ LIST(APPEND _configure_options "-DFFMPEG_LIBAVFORMAT=${_ffmpeg_libavformat}")
 LIST(APPEND _configure_options "-DFFMPEG_LIBAVUTIL=${_ffmpeg_libavutil}")
 LIST(APPEND _configure_options "-DFFMPEG_LIBSWSCALE=${_ffmpeg_libswscale}")
 
+# TODO: Check if RV really need freetype because it is disabled above (USE_FREETYPE=0).
```
bernie-laberge, line 114: I just checked and I think RV needs freetype (Not in oiio but in TwkGLText).
Rationale: TwkGLText depends on FTGL and FTGL depends on freetype.
So perhaps you can simply remove this TODO. 
Thanks for the nice cleanup Cédrik !

bernie-laberge (approved):

geffrak (approved):

PR #357, opened 2024-01-10
Enable Help Menu customization

bernie-laberge:
### Linked issues
NA

This commit's intent is to allow third party applications consuming Open RV as a submodule to customize the Help menu to their liking.

### Summarize your change.

1. The Help menu has been transformed into a package with the exact same content as before.
2. Additionally, a new simple capability to the RV_STAGE(TYPE "RVPKG") macro has been added to prevent an Open RV default package from being installed : RVPKG_DO_NOT_INSTALL_LIST. By appending a package name to this list, one can prevent Open RV from installing the specified package(s).

Tested on Mac.

### DIFF
--- a/cmake/macros/rv_stage.cmake
+++ b/cmake/macros/rv_stage.cmake
@@ -389,91 +389,95 @@ FUNCTION(rv_stage)
     ADD_SHARED_LIBRARY_LIST(${arg_TARGET})
 
   ELSEIF(${arg_TYPE} STREQUAL "RVPKG")
-    SET(_package_file
-        PACKAGE
-    )
-
-    IF(NOT arg_FILES)
-      FILE(
-        GLOB_RECURSE _files
-        RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
-        CONFIGURE_DEPENDS *
-      )
+    IF(${arg_TARGET} IN_LIST RVPKG_DO_NOT_INSTALL_LIST)
+      MESSAGE(STATUS "Skipping RV package '${arg_TARGET}' as requested")
     ELSE()
-      SET(_files
-          ${arg_FILES}
+      SET(_package_file
+          PACKAGE
       )
-    ENDIF()
 
-    LIST(REMOVE_ITEM _files Makefile CMakeLists.txt ${_package_file})
+      IF(NOT arg_FILES)
+        FILE(
+          GLOB_RECURSE _files
+          RELATIVE ${CMAKE_CURRENT_SOURCE_DIR}
+          CONFIGURE_DEPENDS *
+        )
+      ELSE()
+        SET(_files
+            ${arg_FILES}
+        )
+      ENDIF()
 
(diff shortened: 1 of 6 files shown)


src/plugins/rv-packages/openrv_help_menu/openrv_help_menu_mode.mu
```diff
@@ -154,7 +131,7 @@ class: HelpMinorMode : MinorMode
 
     \: inactiveState (int;) { DisabledMenuState; }
 
-    method: HelpMinorMode (HelpMinorMode;)
+    method: OpenRVHelpMenuMinorMode (OpenRVHelpMenuMinorMode;)
     {
         string menuName = "Help";
         if (runtime.build_os() == "DARWIN")
```
geffrak, line 137: Not related to your change but this can go away.

geffrak (approved):

cmake/macros/rv_stage.cmake
```diff
@@ -389,91 +389,95 @@ FUNCTION(rv_stage)
...
-    SET(_package_filename
-        "${RV_PACKAGES_DIR}/${arg_TARGET}-${_pkg_version}.rvpkg"
-    )
+      EXECUTE_PROCESS(
+        COMMAND bash -c "cat ${_package_file} | grep version: | grep --only-matching -e '[0-9.]*'"
+        RESULT_VARIABLE _result
+        OUTPUT_VARIABLE _pkg_version
+        OUTPUT_STRIP_TRAILING_WHITESPACE COMMAND_ERROR_IS_FATAL ANY
+        WORKING_DIRECTORY ${CMAKE_CURRENT_SOURCE_DIR}
+      )
```
cedrik-fuoco-adsk, line 425: Could you do it in 'pure' CMake? Using `cat` will work on Windows since we are using msys2, but that change will tie us more to msys2.

You should be able to use file(READ ...) and string(REGEX MATCH ..) from CMake to do the same.

src/plugins/rv-packages/openrv_help_menu/openrv_help_menu_mode.mu
bernie-laberge: Good catch ! OK done: I removed it.

cmake/macros/rv_stage.cmake
bernie-laberge: Good point ! 
Now, I did not change that part as you can see if you "Hide the white spaces" in the file diff: 
https://github.com/AcademySoftwareFoundation/OpenRV/pull/357/files?diff=split&w=1
I would rather keep the changes to the minimum for now if you don't mind so that we can be in a position to release ASAP. (I still have 2 big issues to investigate on my plate before we do).

cedrik-fuoco-adsk (approved):

cmake/macros/rv_stage.cmake
geffrak: It will indeed need to become a python script. It would also be more robust.

PR #358, opened 2024-01-12
Improves the way third parties are copied into OpenRV's app bundle 

geffrak:
### Summarize your change.

Moves the copy calls from make's built-in copy calls to a Python script.

* CMake's copy script resolves the symlink before copying them
* We were copying files we don't need into the app bundle (ex: .exp, .a, .lib, etc.)
* It allows us to make sure the rpaths are properly set in the stage directory (on MacOS)

Tested on: 

- [X] MacOS
- [ ] Windows
- [ ] Linux

### Add a list of changes, and note any needing special attention during the review.

N/A

N/A


cmake/dependencies/boost.cmake
```diff
@@ -205,7 +205,7 @@ EXTERNALPROJECT_ADD(
     # Ref.: https://www.boost.org/doc/libs/1_70_0/tools/build/doc/html/index.html#bbv2.builtin.features.cflags Ref.:
     # https://www.boost.org/doc/libs/1_76_0/tools/build/doc/html/index.html#bbv2.builtin.features.cflags
     ./b2 -a -q toolset=${_toolset} cxxstd=${RV_CPP_STANDARD} variant=${_boost_variant} link=shared threading=multi architecture=x86 address-model=64
-    ${_boost_with_list} ${_boost_b2_options} -j${_cpu_count} install --prefix=${_install_dir}
+    ${_boost_with_list} ${_boost_b2_options} -j${_cpu_count} install --prefix=${_install_dir} define=_LIBCPP_ENABLE_CXX17_REMOVED_UNARY_BINARY_FUNCTION
```
bernie-laberge, line 208: A fix for XCode 15 ! Nice ! Thanks

bernie-laberge (approved):
This is excellent Kerby !
Wow that was a big undertaking.
Thank you !

cmake/dependencies/boost.cmake
geffrak: Yes, I needed it to compile

PR #381, opened 2024-01-29
Copy RV_DEPS preserving symlinks

geffrak:
### Linked issues

N/A

### Summarize your change.

Moves the copy calls from make's built-in copy calls to a Python script.

* CMake's copy script resolves the symlink before copying them
* We were copying files we don't need into the app bundle (ex: .exp, .a, .lib, etc.)
* It allows us to make sure the rpaths are properly set in the stage directory (on MacOS)

Tested on: 

- [X] MacOS
- [X] Windows
- [X] Linux

### Add a list of changes, and note any needing special attention during the review.

N/A

N/A


bernie-laberge (approved):
Wow ! GG !

PR #390, opened 2024-02-06
Fixing version and release value for sphinx

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Fix the version and release value in the documentation

same as above

n/a

n/a

image

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -5,23 +5,14 @@
 #
 CMAKE_MINIMUM_REQUIRED(VERSION 3.24)
 
-SET(RV_MAJOR_VERSION
-    "2"
-    CACHE STRING "RV's version major"
-)
-SET(RV_MINOR_VERSION
-    "0"
-    CACHE STRING "RV's version minor"
-)
-SET(RV_REVISION_NUMBER
-    "0"
-    CACHE STRING "RV's revision number"
-)
-
-SET(RV_VERSION_YEAR
-    "2024"
-    CACHE STRING "RV's year of release."
-)
+# These variables are parsed by sphinx for the documentation.
+# One-line formatting facilitates parsing and readability.
+# cmake-format: off
+SET(RV_MAJOR_VERSION "2" CACHE STRING "RV's version major")
+SET(RV_MINOR_VERSION "0" CACHE STRING "RV's version minor")
+SET(RV_REVISION_NUMBER "0" CACHE STRING "RV's revision number")
+SET(RV_VERSION_YEAR "2024" CACHE STRING "RV's year of release.")
+# cmake-format: on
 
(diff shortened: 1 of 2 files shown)


docs/conf.py
```diff
@@ -1,12 +1,39 @@
...
-copyright = '2023'
+copyright = '2024'
 author = 'Open RV contributors'
 
-release = '1.0'
+version = '0.0'
+release = '0.0.0'
+
+# Define the regex pattern to match the version numbers
+pattern = r'SET\((?:RV_MAJOR_VERSION|RV_MINOR_VERSION|RV_REVISION_NUMBER)\n\s+"(\d+)"\n\s+CACHE STRING'
```
geffrak, line 14: Can you add a link to regex101 in the comment telling what it's catching and what it's not catching?

geffrak (approved):

docs/conf.py
```diff
@@ -1,12 +1,39 @@
 # Configuration file for the Sphinx documentation builder.
 
 # -- Project information
+import re
 
 project = 'Open RV'
-copyright = '2023'
+copyright = '2024'
```
compaga, line 7: Can we get this from the an already existing Variable? Don't we have it else where?

cedrik-fuoco-adsk: Yea, I just noticed that too. We have RV_VERSION_YEAR in the same file.

geffrak:
It would be easier and safer to have a version file (or multiple) than to parse the CMakeList.txt

cedrik-fuoco-adsk:
> It would be easier and safer to have a version file (or multiple) than to parse the CMakeList.txt. Both the CMakeList.txt and the documentation script could read that file.

There are multiple methods used within the ASWF repository. Some parse the CMakeList.txt and some get it from their Python bindings. I used the CMakeLists.txt because this is where we have the versioning right now, but that could be an option.

CMakeLists.txt
```diff
@@ -5,23 +5,12 @@
...
-    "0"
-    CACHE STRING "RV's revision number"
-)
-
-SET(RV_VERSION_YEAR
-    "2024"
-    CACHE STRING "RV's year of release."
-)
+# These variables are parsed by sphinx for the documentation.
+# One-line formatting facilitates parsing and readability.
```
geffrak, line 9: Can you add the required snippet to avoid cmake-format processing? It's part of the pre-commit hooks and it would be great to not have the parsing broken by a future automatic formatting.

geffrak (approved):

CMakeLists.txt
cedrik-fuoco-adsk: Is that how you are supposed to use it? I've added it below.

geffrak (approved):

compaga (approved):

PR #394, opened 2024-02-13
Leverage new OCIOv2 GPU Processor

### LINKED ISSUE #365: OCIO V2 GPU support
OCIO v2 Phase 2: use the new GPU API

bernie-laberge:
### Linked issues
Fixes #365

Even though Open RV was recently upgraded to OCIOv2, it was only using the legacy GPU processor implementation of OCIOv2 which is less accurate and more limited than the new OCIOv2 GPU processor implementation. 
This commit leverages the new OCIOv2 GPU processor implementation in Open RV.
 
### Summarize your change.
Prior to this commit, only the legacy OCIOv2 GPU processor was used (OCIO::getOptimizedLegacyGPUProcessor).
In this commit, the new OCIOv2 GPU processor implementation is now used by default.
With the aim of maintaining backward compatibility of results, I have added a new environment variable which can be optionally defined to tell OCIO to use the legacy GPU processor implementation: 
`export RV_OCIO_USE_LEGACY_GPU_PROCESSOR=1`

The main change is that we are now calling OCIO::getOptimizedGPUProcessor() instead of OCIO::getOptimizedLegacyGPUProcessor() and that we have to handle the possibility of having an arbitrary number of 1D LUT(s) and 3D LUT(s). Prior to this commit, only 1x optional 3D LUT could be handled.
This is why I decided to model the 1D LUTs and the 3D LUTs in their own classes to hide the OCIO calls to fetch their values since that they translate to different OCIOv2 calls. This way the OCIOIPNode::evaluate() code ended up not being too much impacted by it. It is long enough as it is.

Built and Tested on Mac, RockyLinux9, and Windows.
Using this OCIOv2 config: studio-config-v2.0.0_aces-v1.3_ocio-v2.2.ocio.

Also fixed in this commit: Fixed a crash in RV when the provided OCIO config did not contain the OCIO::ROLE_SCENE_LINEAR color space (this was unlikely to occur with production OCIO configs but it happened during my tests using a very customized OCIO config).

ocio_display_menu

### DIFF
--- a/src/lib/ip/OCIONodes/OCIO1DLUT.cpp
+++ b/src/lib/ip/OCIONodes/OCIO1DLUT.cpp
@@ -1,0 +1,55 @@
+//
+// Copyright (C) 2024  Autodesk, Inc. All Rights Reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+
+#include <OCIONodes/OCIO1DLUT.h>
+#include <TwkExc/Exception.h>
+
+namespace IPCore
+{
+
+  OCIO1DLUT::OCIO1DLUT( OCIO::GpuShaderDescRcPtr& shaderDesc, unsigned int idx,
+                        const std::string& shaderCacheID )
+  {
+    // Get the information of the 1D LUT.
+    const char* textureName = nullptr;
+    const char* samplerName = nullptr;
+    unsigned width = 0;
+    unsigned height = 0;
+    OCIO::GpuShaderDesc::TextureType textureType =
+        OCIO::GpuShaderDesc::TEXTURE_RGB_CHANNEL;
+    OCIO::Interpolation interpolation = OCIO::INTERP_BEST;
+    shaderDesc->getTexture( idx, textureName, samplerName, width, height,
+                            textureType, interpolation );
+    if( !samplerName || !*samplerName || width == 0 )
+    {
+      TWK_THROW_EXC_STREAM( "The OCIO 1D LUT texture data is corrupted" );
+    }
+    const int channelSize =
+        ( textureType == OCIO::GpuShaderDesc::TEXTURE_RGB_CHANNEL ) ? 3 : 1;
+    m_samplerName = samplerName;
+    m_height = height;
+
+    m_lutfb = new TwkFB::FrameBuffer(
+        TwkFB::FrameBuffer::NormalizedCoordinates, width, height, 1 /*depth*/,
+        channelSize, TwkFB::FrameBuffer::FLOAT, nullptr /*data*/,
+        nullptr /*channelNames are optional*/, TwkFB::FrameBuffer::BOTTOMLEFT,
+        true /*deleteOnDestruction*/ );
+
+    m_lutfb->staticRef();
+    m_lutfb->setIdentifier( shaderCacheID + "|" + m_samplerName );
+
+    const float* values = nullptr;
+    shaderDesc->getTextureValues( idx, values );
+    if( !values )
+    {
+      TWK_THROW_EXC_STREAM( "The OCIO 1D LUT texture values are missing" );
+    }
+
+    memcpy( m_lutfb->pixels<float>(), values,
+            channelSize * sizeof( float ) * width * height );
+  }
+
+}  // namespace IPCore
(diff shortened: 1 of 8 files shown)


cedrik-fuoco-adsk (approved):

src/lib/ip/OCIONodes/OCIONodes/OCIO1DLUT.h
```diff
@@ -0,0 +1,42 @@
...
+   public:
+    OCIO1DLUT( OCIO::GpuShaderDescRcPtr& shaderDesc, unsigned int idx,
+               const std::string& shaderCacheID );
+    virtual ~OCIO1DLUT() {};
+
+    // Returns the LUT sampler type: sampler3D, sampler2D, or sampler1D
+    // Note: a 2D texture is needed to hold large LUTs.
+    std::string samplerType() const override
+    {
+      return m_height == 1 ? "sampler1D" : "sampler2D";
```
doug-walker, line 35: When you upgrade from OCIO 2.2 to 2.3, the getTexture call you use in OCIO1DLUT.cpp changes to return a TextureDimensions parameter that should be used rather than this check if the height is 1.

doug-walker:
This is awesome Bernard!

src/lib/ip/OCIONodes/OCIONodes/OCIO1DLUT.h
bernie-laberge: Thank for for this information ! 
OpenRV is currently aligned with VFXPlatform=CY2023 but we'll soon add a build option for the VFXPlatform=CY2024.
I will make sure to use the new TextureDimensions parameter once we add support for the VFXPlatform=CY2024.
Thanks!

geffrak (approved):

markreidvfx (commented):
This is fantastic!  Tested it out on CentOS and had no problems. We also tested a few transforms known to cause issues like banding and artifacting with the LegacyGPUProcessor and this completely resolves those issues. Great work!

src/lib/ip/OCIONodes/OCIO1DLUT.cpp
```diff
@@ -0,0 +1,55 @@
...
+    m_lutfb = new TwkFB::FrameBuffer(
+        TwkFB::FrameBuffer::NormalizedCoordinates, width, height, 1 /*depth*/,
+        channelSize, TwkFB::FrameBuffer::FLOAT, nullptr /*data*/,
+        nullptr /*channelNames are optional*/, TwkFB::FrameBuffer::BOTTOMLEFT,
+        true /*deleteOnDestruction*/ );
+
+    m_lutfb->staticRef();
+    m_lutfb->setIdentifier( shaderCacheID + "|" + m_samplerName );
+
+    const float* values = 0x0;
```
markreidvfx, line 44: nullptr?

src/lib/ip/OCIONodes/OCIO3DLUT.cpp
```diff
@@ -0,0 +1,49 @@
...
+    m_lutfb = new TwkFB::FrameBuffer(
+        TwkFB::FrameBuffer::NormalizedCoordinates, edgelen, edgelen, edgelen,
+        3 /*channelSize*/, TwkFB::FrameBuffer::FLOAT, nullptr /*data*/,
+        nullptr /*channelNames are optional*/, TwkFB::FrameBuffer::BOTTOMLEFT,
+        true /*deleteOnDestruction*/ );
+
+    m_lutfb->staticRef();
+    m_lutfb->setIdentifier( shaderCacheID + "|" + m_samplerName );
+
+    const float* values = 0x0;
```
markreidvfx, line 38: nullptr?

src/lib/ip/OCIONodes/OCIO1DLUT.cpp
bernie-laberge: Good point ! OK done. Thanks.

src/lib/ip/OCIONodes/OCIO3DLUT.cpp
bernie-laberge: Good point ! OK done. Thanks.

PR #399, opened 2024-02-26
Add build switch for VFX platform and implement it for OCIO

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Added a option to choose which VFX platform to use for the affected dependencies.
`RV_VFX_PLATFORM=CY2023` or `RV_VFX_PLATFORM=CY2024`

Handle different VFX platform

Tested on Windows and tried to use OCIO.
Build successful on MacOS and Rocky linux.

n/a

n/a


cmake/macros/rv_vfx.cmake
```diff
@@ -0,0 +1,61 @@
+#
+# Copyright (C) 2024  Autodesk, Inc. All Rights Reserved.
+#
+# SPDX-License-Identifier: Apache-2.0
+#
```
cedrik-fuoco-adsk, line 5: What's the correct copyright to use for new file in OpenRV?

geffrak (approved):
GG

cmake/dependencies/ocio.cmake
```diff
@@ -45,9 +51,18 @@ IF(RV_TARGET_WINDOWS)
   SET(_byproducts
       ""
   ) # Empty out the List as it has the wrong DLL name: it doesn't have the version suffix
-  SET(_ocio_win_sharedlibname
-      OpenColorIO_2_2.dll
-  )
+
+  # OpenColorIO shared library has the same name on Release and Debug.
+  IF(RV_VFX_CY2023)
```
geffrak, line 56: Instead of the specialized `RV_VFX_SET_VERSION` and  `RV_VFX_SET_DOWNLOAD_HASH` perhaps we should have something like

```
RV_VXF_SET_VARIABLE(
    CY2023 VAR_A
    CY2024 VAR_B
    OUTPUT_VARIABLE_NAME
)
```

Where you could then do 
```
RV_VFX_SET_VARIABLE(
  CY2023 "372d6982cf01818a21a12f9628701a91"
  CY2024 "8af74fcb8c4820ab21204463a06ba490"
  _download_hash
)
```

```
RV_VFX_SET_VARIABLE(
  CY2023 "2.2.1"
  CY2024 "2.3.2"
  _version
)
```


```
RV_VFX_SET_VARIABLE(
  CY2023 "OpenColorIO_2_2.dll"
  CY2024 "OpenColorIO_2.3.dll"
  _ocio_win_sharedlibname
)
```

cmake/macros/rv_vfx.cmake
```diff
@@ -0,0 +1,61 @@
...
+# RV_VFX_SET_DOWNLOAD_HASH output the right hash based on the VFX platform.
+# The value is choosen based on the RV_VFX_CY20XX variable.
+# Output variable: _download_hash
+macro(RV_VFX_SET_DOWNLOAD_HASH)
+  cmake_parse_arguments(
+      _rv_dep_hash
+      # options
+      ""
+      # one value keywords
+      "CY2023;CY2024"
```
geffrak, line 17: Can we avoid it being hardcoded? Using some variant of `IF(NOT "${_RV_VFX_PLATFORM}" IN_LIST _RV_VFX_SUPPORTED_OPTIONS)` and accessing the OPTION by name.

geffrak (approved):
Extra comments, but it's good enough the way it is.

cmake/dependencies/ocio.cmake
bernie-laberge: I really liked the scheme you came up with Cédrik like RV_VFX_SET_VERSION.
But it is not clear to the reader which variable is actually being set so there is a higher risk of typos.
Kerby's suggestion solves this issue of not knowing exactly which variable is being set. I think it would be totally worth it to implement it.
It brings your clever original scheme to a different level.
This is really nice !
```
RV_VFX_SET_VARIABLE(
  CY2023 "2.2.1"
  CY2024 "2.3.2"
  _version
)
```

cmake/macros/rv_vfx.cmake
bernie-laberge: It is the correct one. The important part is # SPDX-License-Identifier: Apache-2.0

cmake/dependencies/ocio.cmake
```diff
@@ -30,8 +36,10 @@ ELSE()
   )
 ENDIF()
 
-SET(_download_hash
-    "372d6982cf01818a21a12f9628701a91"
+RV_VFX_SET_VARIABLE(
```
bernie-laberge, line 39: Beautiful !

PR #421, opened 2024-03-21
Fix crashes on CentOS7 by tweaking the error handling of libpng

### LINKED ISSUE #420: [Bug]: Crash with an image sequence containing incomplete PNGs
### What happened?

OpenRV crashes while opening an image sequence of PNGs that has incomplete/corrupted images (e.g. size of 0 byte).
The crash happens specifically when using a CentOS 7 build. A Rocky Linux build does not crash.

CentOS 7

Workstation / Any

```shell
libpng error: Read Error
Segmentation fault (core dumped)
```

### Environment variables

_No response_

### Extra information

Image sequence of PNGs with incomplete file.

### Code of Conduct

cedrik-fuoco-adsk:
### Linked issues
Fixes #420

### Summarize your change.
Modified the custom error handler to support the jmpbuf address instead of relying on libpng to save the address.
It should make sure that the calling convention are similar.

OpenRV was crashing (CentOS7) when opening images sequence of PNGs that had incomplete PNG.

- [ ] Rocky Linux 8
- [ ] CentOS 7
- [ ] Windows (todo)
- [ ] MacOS (todo)

### DIFF
--- a/cmake/dependencies/png.cmake
+++ b/cmake/dependencies/png.cmake
@@ -63,7 +63,7 @@ RV_COPY_LIB_BIN_FOLDERS()
 
 ADD_DEPENDENCIES(dependencies ${_target}-stage-target)
 
-ADD_LIBRARY(PNG::PNG STATIC IMPORTED GLOBAL)
+ADD_LIBRARY(PNG::PNG SHARED IMPORTED GLOBAL)
 ADD_DEPENDENCIES(PNG::PNG ${_target})
 IF(NOT RV_TARGET_WINDOWS)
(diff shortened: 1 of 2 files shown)


cmake/dependencies/png.cmake
cedrik-fuoco-adsk, line 1: Not needed for the fix, but there were a missmatch of shared/static in png.cmake

bernie-laberge (approved):
Look good to me !
Thank you Cédrik !

geffrak (approved):

PR #439, opened 2024-04-11
RVPaint fix segment edge intersection when path has variable widths

markreidvfx:
This fixes an issue when using the annotation tools and having strokes with variable widths.

Variable widths happen when you use a stylus/wacom tablet with pen pressure. When widths can vary, it is possible for the widths of the stroke be just right so that the extruded edges become nearly parallel. This case currently isn't being handled. This creates "spikes" on your brush stroke that can seemly appear at random.

## Example Of The Spikes Occurring

https://github.com/AcademySoftwareFoundation/OpenRV/assets/814966/b7f1dd3f-1eee-4d62-97ec-add94ac12e11

## Isolated Case
Here is a minimal isolated case of the issue happening [spike_isolated.rv.](https://github.com/AcademySoftwareFoundation/OpenRV/files/14950061/spike_isolated.rv.zip)

image

I also graphed the polygons in desmos so its easier to see
https://www.desmos.com/calculator/gxmkrpdr2o

The issue occurs when the point of intersection is calculated between line segments AB and CD.
image

As we can see the two lines are is nearly parallel. 
image

When intersection point is calculated, away we go!
image

## The Solution

To fix this issue, I changed the code to perform the intersection as if the line has a constant width. This was the simplest thing I could think of. The spikes never seem to happen with constant width strokes.

### DIFF
--- a/src/lib/geometry/TwkPaint/Path.cpp
+++ b/src/lib/geometry/TwkPaint/Path.cpp
@@ -348,6 +348,9 @@ Path::computeSegmentQuads(SegmentVector& segments,
         s.points[2] = p1 + vpn * w1;
         s.points[3] = p1 - vpn * w1;
 
+        s.vpn = vpn;
+        s.w0 = w0;
+
         if (usecolor)
         {
             const Color& c0  = fcolors[i];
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):
I'm afraid I don't have anything better to propose so your proposed change looks good to me.
Many thanks for the fix and for such a nice PR description !

cedrik-fuoco-adsk (approved):

Issue #447, opened 2024-04-19
[Bug]: PNG decoder is not enabled (FFmpeg, Windows)


cedrik-fuoco-adsk:
### What happened?

PNG movies can not be decoder because the PNG decoder is not enabled on Windows.

PNG decoder is disabled on Windows because ZLIB is not found by FFmpeg.

### List all the operating systems versions where this is happening

Windows

### On what computer hardware is this happening?

Windows

### Relevant console log output

```shell
ERROR: MovieFFMpeg: Unsupported codec_id '61'
```

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

cedrik-fuoco-adsk:
Closed by #448

PR #448, opened 2024-04-19
Enable PNG decoder on Windows

cedrik-fuoco-adsk:
### Linked issues
#447 

### Summarize your change.
Make sure that FFmpeg knows where to 

PNG movies were not able to be decoded because PNG decoder in FFmpeg was disabled due to missing ZLIB library during the build of FFmpeg. 

Since ZLIB is installed by default on MacOS and Linux, it wasn't a problem on those systems. But it means that the wrong ZLIB was used by FFmpeg (not the one compiled by OpenRV).

- [ ] Windows only

n/a

n/a


geffrak (approved):

bernie-laberge (approved):
Nice catch !

PR #461, opened 2024-05-07
Support native Apple ARM64

cedrik-fuoco-adsk:
# Support native Apple ARM64

### Linked issues
n/a

### Summarize your change.
Since using CMAKE_OSX_ARCHITECTURES is common practice in the CMake world, I made changes to the build system to use that variable to decide whether the build is for x86_64 or ARM64. By default, CMAKE_OSX_ARCHITECTURES will be set to the native architecture of the system that the build is executed.

The dependencies that are using CMake will build for the right architecture based on CMAKE_OSX_ARCHITECTURES and those that do not use CMake are using the supported method to cross-compile (not using `arch -x86_64`) to another architecture (except `OpenSSL` and `Python`). 

Python build system does not support cross-compilation between Apple x86_64 and ARM64, only supports universal build. We could use the universal build of Python, but that would requires RV to do a universal build of OpenSSL and the other Python dependencies (readline, tcl-tk, xz, etc).

Support native build of Apple chipset.

- [ ] MacOS
- [ ] Windows (CI)
- [ ] Rocky Linux 9 (CI)


geffrak:
I'm not convinced that `CMAKE_OSX_ARCHITECTURES` is a good solution. Only some third-party dependencies support it (as you can see with Python and OpenSSL). I think we should get rid of it and build everything natively. If someone wants to build x86_64 from an Apple Silicon host, that person starts his terminal with `arch -x86_64` and builds.

cedrik-fuoco-adsk:
In my mind, there is a difference between cross-compiling and Rosetta since Rosetta is more an emulator/translator than a cross-compiler. For example, when you are building for iOS, you are not using Rosetta, you are using a cross-compiler. So at first, I treated the ARM64 the same way.

So the question is whether we want to support cross-compilation or just native builds. If we want to support cross-compilation, CMAKE_OSX_ARCHITECTURES is the right choice, otherwise it is not needed.

After discussion with Bernard about your comment @geffrak, we don't really see a use-case to support cross-compilation. So I removed everything related to CMAKE_OSX_ARCHITECTURES and we will just support native build. 

But like you said, you can still build for `x86_64` from `arm64` using a terminal launched with `arch -x86_64` (using Rosetta).

geffrak (approved):

bernie-laberge (approved):
Wow ! Beautiful and nice cleanup !
Thanks

PR #471, opened 2024-05-24
Add .rvpkg bundle support for rvpkg command line tools and UI

chxmberland:
### Summarize your change.

Users can now upload multiple packages at the same time through a compressed zip file using the extension **.rvpkgs**. These **.rvpkgs** files can be uploaded using the `rvpkg -add ...` command in the shell, or directly in the UI under the `Preferences/Packages` section of the menu.

In either case, **.rvpkgs** files will be immediately unzipped into the packages they contain before each package is added to OpenRV. 

Users can also add and install these **.rvpkgs** bundles using a chain command as follows.

`rvpkg -force -install -add <destination> <location of .rvpkgs file>`

This change was tested on MacOS. To replicate the test, recreate the following steps.

**Creating an .rvpkgs file**

1. Find the .rvpkg files you want to bundle
2. Create a zip folder containing these .rvpkg files
3. Rename this zip file following `<name>-<version>.rvpkgs`

**Testing UI upload**

1. Launch OpenRV
2. Navigate to the `Preferences/Packages` section of the menu
3. Select `Add Packages...` 
4. Select your own **.rvpkgs** file
5. Click open
6. The packages zipped into the **.rvpkgs** file should be added directly into RV

**Testing Command Line**

1. Determine the location of your **.rvpkgs** file
2. Determine the location you want to add/install your  packages to
3. Execute either of the following commands

`rvpkg -force -install -add <destination> <location of .rvpkgs file>`
`rvpkg -add <destination> <location of .rvpkgs file>`

Note that packages are often installed to `~/Library/Application\ Support/RV/` on Mac.

- [ ] Added bundle detection and handling functionality to the rvpkg main file (specifically -add and -install sections)
- [ ] Minor bug fix in RvPreferences where text indicated incorrectly that packages could not be uninstalled
- [ ] Added a multitude of utility functions to the PackageManager, compartmentalizing and globalizing unzip and bundle detection functionality

### DIFF
--- a/src/lib/app/RvPackage/PackageManager.cpp
+++ b/src/lib/app/RvPackage/PackageManager.cpp
@@ -16,10 +16,142 @@
 #include <stl_ext/string_algo.h>
 #include <fstream>
 #include <string_view>
+#include <iostream>
 
 namespace Rv
 {
   using namespace std;
+
+  QString extractFile(unzFile uf, const string& outputPath) {
+    char filename[256];
+    unz_file_info fileInfo;
+    
+    if (unzGetCurrentFileInfo(uf, &fileInfo, filename, sizeof(filename), nullptr, 0, nullptr, 0) != UNZ_OK) {
+        cerr << "ERROR: Unable to get file info" << endl;
+        return "";
+    }
+    
+    if (unzOpenCurrentFile(uf) != UNZ_OK) {
+        cerr << "ERROR: Unable to open file in zip" << endl;
+        return "";
+    }
+    
+    string fullPath = outputPath + "/" + filename;
+    FILE* outFile = std::fopen(fullPath.c_str(), "wb");
+    if (outFile == nullptr) {
+        cerr << "ERROR: Unable to open output file" << endl;
+        unzCloseCurrentFile(uf);
+        return "";
+    }
+    
+    char buffer[8192];
+    int bytesRead = 0;
+    while ((bytesRead = unzReadCurrentFile(uf, buffer, sizeof(buffer))) > 0) {
+        fwrite(buffer, 1, bytesRead, outFile);
+    }
+    
+    fclose(outFile);
+    unzCloseCurrentFile(uf);
+
+    return QString::fromStdString(fullPath);
+  }
+
+  vector<QString> unzipBundle(const string& zipFilePath, const string& outputPath) 
+  {
+    vector<QString> includedPackages;
+
+    // Opening zip file
+    unzFile uf = unzOpen(zipFilePath.c_str());
+    if (uf == nullptr) {
+      cerr << "ERROR: Unable to open zip file" << endl;
+      return {};
+    }
+    
+    // Checking for the presence of files
+    if (unzGoToFirstFile(uf) != UNZ_OK) {
+      cerr << "ERROR: Unable to find first file in zip" << endl;
+      unzClose(uf);
+      return {};
+    }
+    
+    // Extracting each file in the zip
+    do {
+      QString extractedFilePath = extractFile(uf, outputPath);
+      if (extractedFilePath == "") {
+        cerr << "ERROR: Unable to extract zip file" << endl;
+      } else {
+        includedPackages.push_back(extractedFilePath);
+      }
+    } while (unzGoToNextFile(uf) == UNZ_OK);
+    
+    unzClose(uf);
+
+    return includedPackages;
+  }
+
+  bool PackageManager::isBundle(const QString& infileNonCanonical)
+  {
+
+    // Getting package file extension
+    string filePath = infileNonCanonical.toStdString();
+    size_t dot = filePath.rfind('.');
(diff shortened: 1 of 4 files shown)


src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -1583,10 +1693,27 @@ namespace Rv
...
+      // Unpacking any bundles
+      if (isBundle(files[i])) {
+        cout << "INFO: Bundle detected, adding subpackages now." << endl;
+
+        // Installing bundle
+        vector<QString> includedPackages = handleBundle(files[i], path);
+        if (includedPackages.size() > 0) {
+          cout << "INFO: Bundle subpackages added." << endl;
+        } else {
+          cout << "ERROR: Unnable to add subpackages." << endl;
```
bernie-laberge, line 1711: You should use cerr for ERRORs (instead of cout)
(Note that I know that this rule is definitely not consistent in the legacy code, but at least for new code let's make it)

src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -1770,9 +1902,9 @@ namespace Rv
   bool PackageManager::installDependantPackages( const QString& msg )
   {
     return yesOrNo( "Some Packages Depend on This One",
-                    "Can't uninstall package because some other packages "
+                    "Can't install package because some other packages "
```
bernie-laberge, line 1905: Thank you for fixing this legacy issue ! From 14 years ago ! I checked.

src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -21,6 +21,112 @@ namespace Rv
...
+    return includedPackages;
+  }
+
+  bool PackageManager::isBundle(const QString& infileNonCanonical)
+  {
+
+    // Getting package file extension
+    string filePath = infileNonCanonical.toStdString();
+    size_t dot = filePath.rfind('.');
+    if (dot == string::npos) {
```
bernie-laberge, line 108: Say the last character is a dot '.' I believe the filePath.substr(dot + 1) will throw.
Perhaps you could make sure that this doesn't happen by adding a condition to that if.

src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -21,6 +21,112 @@ namespace Rv
...
+    string ext = filePath.substr(dot + 1);
+    return ext == "rvpkgs";
+  }
+
+  vector<QString> PackageManager::handleBundle(const QString& bundlePath, const QString& outputPath) {
+
+    // Attempting to unzip bundle into the Packages directory
+    vector<QString> includedPackages = unzipBundle(bundlePath.toStdString(), outputPath.toStdString());
+    if (includedPackages.size() == 0) {
+        cout << "ERROR: Unnable to unzip bundle." << endl;
```
bernie-laberge, line 123: typo: Unable

src/lib/app/RvPackage/PackageManager.cpp
bernie-laberge, line 1711: Typo: Unable (Unless I am mistaken ! I could be since I'm a frenchie!)

chxmberland: It is a typo, you're right!

bernie-laberge (approved):
Looks good to me. 
Nicely done Benjamin ! Thanks !

src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -231,7 +343,7 @@ namespace Rv
     {
       cerr << "ERROR: Illegal package file name: "
            << package.file.toUtf8().data() << endl
-           << "       should be either <name>.zip or <name>-<version>.rvpkg"
+           << "       should be either <name>.zip, <name>-<version>.rvpkg or <name>-<version>.rvpkgs"
```
geffrak, line 346: Why do a rvpkgs have a version (or require a version)?

geffrak (approved):
Looks good to me. Question: Where is the rvpkgs content extracted before being added to the rvpkg args?

src/lib/app/RvPackage/PackageManager.cpp
chxmberland: I thought versioning may be helpful if the packages included in bundles change overtime. For example, the Flow Production Tracking bundle that will eventually be published contains several packages which may be merged together or removed entirely. However, this nomenclature is not actually enforced and is meant to be merely a suggestion. Do you think this is appropriate?

geffrak: If it's not enforced, I would not mention it. But it's just my opinion. @bernie-laberge might have another, and his opinion is the final veto.

geffrak (approved):

bernie-laberge (approved):

PR #475, opened 2024-05-28
Make OpenRV buildable on Xcode 15

### LINKED ISSUE #278: [Bug]: OpenRV can't build with Xcode 15
### What happened?

OpenRV's PySide2 target can't be built with Xcode 15 because of https://bugreports.qt.io/browse/QTBUG-117225. Older versions of xcode are available on https://developer.apple.com

macOS

N/A

```shell
N/A
```

### Environment variables

N/A

### Extra information

N/A

### Code of Conduct

chxmberland:
### Linked issues

Fixes #278

### Summarize your change.

OpenRV will now default to using Boost 1.81 if the user has Xcode 15 or greater. Otherwise, OpenRV will default to using Boost 1.80, which is compliant with CY 2023 and works with Xcode 14.3.1. This also is outlined for the user in changes to documentation. 

Xcode 14.3.1 is deprecated on macOS Sonoma in favour of Xcode 15 and up, which means users on Sonoma or later will not be able to build OpenRV without this change.

A complete build of OpenRV using `rvbootstrap` on macOS Sonoma and Xcode 15.

- [ ] CMake automatically uses Boost 1.80 if the Clang version provided by Xcode is less than 15

### DIFF
--- a/cmake/dependencies/boost.cmake
+++ b/cmake/dependencies/boost.cmake
@@ -21,13 +21,44 @@ SET(_major_minor_version
     "1_80"
 )
 
+SET(_download_hash
+    077f074743ea7b0cb49c6ed43953ae95
+)
+
+# Note: Boost 1.80 cannot be built with XCode 15 which is now the only XCode version available on macOS Sonoma without a hack. Boost 1.81 has all the fixes
+# required to be able to be built with XCode 15, however it is not VFX Platform CY2023 compliant which specifies Boost version 1.80. With the aim of making the
+# OpenRV build on macOS smoother by default, we will use Boost 1.81 if XCode 15 or more recent.
+IF(RV_TARGET_DARWIN)
+  EXECUTE_PROCESS(
+    COMMAND clang --version
+    OUTPUT_VARIABLE CLANG_FULL_VERSION_STRING
+  )
+  STRING(
+    REGEX
+    REPLACE ".*clang version ([0-9]+\\.[0-9]+).*" "\\1" CLANG_VERSION_STRING ${CLANG_FULL_VERSION_STRING}
+  )
+  IF(CLANG_VERSION_STRING VERSION_GREATER_EQUAL 15.0)
+    MESSAGE(STATUS "Clang version ${CLANG_VERSION_STRING} is not compatible with Boost 1.80, using Boost 1.81 instead. "
+                   "Install XCode 14.3.1 if you absolutely want to use Boost version 1.80 as per VFX reference platform CY2023"
+    )
+
+    SET(_version
+        "1.81.0"
+    )
+
+    SET(_major_minor_version
+        "1_81"
+    )
+
+    SET(_download_hash
+        4bf02e84afb56dfdccd1e6aec9911f4b
+    )
+  ENDIF()
+ENDIF()
+
 STRING(REPLACE "." "_" _version_with_underscore ${_version})
 SET(_download_url
     "https://archives.boost.io/release/${_version}/source/boost_${_version_with_underscore}.tar.gz"
-)
-
-SET(_download_hash
-    077f074743ea7b0cb49c6ed43953ae95
 )
 
 # Set _base_dir for Clean-<target>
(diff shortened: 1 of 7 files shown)


cmake/dependencies/boost.cmake
```diff
@@ -21,15 +21,46 @@ SET(_major_minor_version
...
+SET(_download_hash
+    077f074743ea7b0cb49c6ed43953ae95
+)
+
+# Note: Boost 1.80 cannot be built with XCode 15 which is now the only XCode version available on macOS Sonoma without a hack. Boost 1.81 has all the fixes
+# required to be able to be built with XCode 15, however it is not VFX Platform CY2023 compliant which specifies Boost version 1.80. With the aim of making the
+# OpenRV build on macOS smoother by default, we will use Boost 1.81 if XCode 15 or more recent.
+IF(RV_TARGET_DARWIN)
+  EXECUTE_PROCESS(
+    COMMAND clang --version
```
geffrak, line 33: I would recommend you `xcrun clang --version` to protect the code against `clang` from a homebrew formula. 

Also, you also have `xcodebuild -version` that directly returns the version of Xcode.

rvcmds.sh
```diff
@@ -47,7 +47,7 @@ RV_BUILD_PARALLELISM="${RV_BUILD_PARALLELISM:-$(python3 -c 'import os; print(os.
 
 # ALIASES: Basic commands
 
-alias rvsetup="SETUPTOOLS_USE_DISTUTILS=${SETUPTOOLS_USE_DISTUTILS} python3 -m pip install --user --upgrade -r ${RV_HOME}/requirements.txt"
+alias rvsetup="SETUPTOOLS_USE_DISTUTILS=${SETUPTOOLS_USE_DISTUTILS} python3 -m pip install --user --upgrade -r ${RV_HOME}/requirements.txt --break-system-packages"
```
geffrak, line 50: It's never a good thing to break the system packages... It's even worse on a script that people execute without noticing it. What is the system package that is complaining?

geffrak (changes requested):

rvcmds.sh
cedrik-fuoco-adsk, line 50: That option is only valid for Python 3.12+. Should we check for the Python version beforehand? @chxmberland @bernie-laberge

cmake/dependencies/boost.cmake
bernie-laberge: Good point ! Thank you Kerby !

rvcmds.sh
chxmberland: That's a mistake, that change should not have been pushed through. I use it locally as a workaround. That will be removed immediately.

chxmberland: See [above](https://github.com/AcademySoftwareFoundation/OpenRV/pull/475#discussion_r1617258278). This will be removed immediately. Apologies.

bernie-laberge: @chxmberland : Kerby is right: We should remove this --break-system-packages until we find a better solution

cedrik-fuoco-adsk: It is a new behaviour in Python 3.12 - https://peps.python.org/pep-0668/. But I do agree with Kerby. If I recall,  the official recommendation is to use a virtual environment like pyenv.

src/build/make_pyside.py
```diff
@@ -86,16 +86,11 @@ def prepare() -> None:
...
         )
-        clang_version_str = "".join(clang_version_search.groups())
-        clang_version_int = int(clang_version_str)
-
-        if clang_version_int <= 120:
-            clang_filename_suffix = clang_version_str + "-based-mac.7z"
-        else:
-            clang_filename_suffix = clang_version_str + "-based-macos-universal.7z"
+        clang_version_str = ".".join(clang_version_search.groups())
+        clang_filename_suffix = clang_version_str + "-based-macos-universal.7z"
```
cedrik-fuoco-adsk, line 93: It seems ok, but have you tried to build using Xcode 14 with those changes?

rvcmds.sh
chxmberland: It's been removed, apologies for the confusion. It was not meant to be included in the PR in the first place.

bernie-laberge (approved):
Superb !
Thank you Benjamin

Issue #463, opened 2024-05-15
[Bug]: OCIO causes crashing unless reader threads set to 1


kenmcgaugh:
### What happened?

Enabling OCIO causes RV to become very unstable. The frequency of the crashes increases with the complexity of the python code that loads and processes the config before running OCIO.SetCurrentConfig(config). In my case it crashes 75% of the time the moment it first tries to render a frame.

The crashng happens in the "IPGraph Eval" thread within the IPCore::OCIOIPNode::evaluate() method.

The only workaround I've found is to set the number of reading threads to 1.

### List all the operating systems versions where this is happening

macOS 14.3

### On what computer hardware is this happening?

Apple macbook pro M1 16GB

### Relevant console log output

```shell
Sometimes it is just a Segmentation Fault. Other times it is something like:

RV (...) malloc: Heap corruption detected, free list is damaged at ...
*** Incorrect guard value: ...
```

### Environment variables

DISPLAY=/private/tmp/com.apple.launchd.Ho7Xi9pMTK/org.xquartz:0
EDITOR=emacs
GROUP=unknown
HOME=/Users/ken
HOST=legospace
HOSTTYPE=unknown
IOCIN_ARGS= --format RGB8_PLANAR --useChromaticities 0 --ioSize 61440 --ioMethod 2 --ioMaxAsync 16
IODPX_ARGS= --format RGB8_PLANAR --useChromaticities 0 --ioSize 61440 --ioMethod 2 --ioMaxAsync 16
IOEXR_ARGS= --rgbaOnly 0 --inherit 1 --noOneChannel 0 --readWindowIsDisplayWindow 0 --readWindow 3 --planar3channel 0 --stripAlpha 0 --convertYRYBY 0 --ioSize 61440 --ioMethod 3 --ioMaxAsync 16
IOJPEG_ARGS= --ioSize 61440 --ioMethod 0 --ioMaxAsync 16
IOTARGA_ARGS= --ioSize 61440 --ioMethod 2 --ioMaxAsync 16
IOTIFF_ARGS= --ioSize 61440 --ioMethod 2
LANG=C
LC_ALL=C
LOGNAME=ken
MACHTYPE=unknown
MOVIEFFMPEG_ARGS= --defaultFPS 24
MOVIEMISTIKA_ARGS= --format RGB8_PLANAR
[...]

PR #476, opened 2024-05-29
463: Unanonymised QMutexLocker usage in OCIOIPNode.cpp

kenmcgaugh:
### Linked issues
Should fix [#463](https://github.com/AcademySoftwareFoundation/OpenRV/issues/463)

### Summarize your change.
Added "lock" variable name to all anonymous QMutexLocker instances to extend their lifespan to the end of their scope. This matches the QMutexLocker usage in other parts of the app (RVConsoleWindow.cpp and QTAudioRenderer.cpp).

### Describe the reason for the change
Enabling OCIO was causing frequent crashing when reader threads were set greater than 1, in particular under MacOS Sonoma.


geffrak (approved):

bernie-laberge (approved):
Wow ! Good catch !
Thank you @kenmcgaugh !

PR #485, opened 2024-06-11
Fix rvio segmentation fault when using threads to encode image sequences

### LINKED ISSUE #486: [Bug]: rvio segmentation fault when using threads to encode image sequences
### What happened?

When trying to use the `rvio` library with more then one thread to transcode any type of media through the `-rthreads` option, a segmentation fault occurs. For example, the following line gives a segmentation fault:

`./rvio -rthreads 2 clip1.mov -o out.#.jpg`

Reported in Centos 7.8 with RV 2023.0.1 

Validated by support in MacOS 13.6 RV 2021, 2022, 2023 and Centos 8.5 RV 2023.0.1

```
dlopen(v??b??, 0x0001): tried: 'v??b??' (no such file), '/System/Volumes/Preboot/Cryptexes/OSv??b??' (no such file), '/Users/<user>/Dev/OpenRV/_build/stage/app/RV.app/Contents/Frameworks/v??b??' (no such file), '/Users/<user>/Dev/OpenRV/_build/stage/app/RV.app/Contents/lib/v??b??' (no such file),  '/usr/lib/v??b??' (no such file, not in dyld cache), 'v??b??' (no such file), '/usr/local/lib/v??b??' (no such file),  '/usr/lib/v??b??' (no such file, not in dyld cache) WARNING:  failed to load zsh: segmentation fault  rvio -rthreads 2 ~/Downloads/friendly-lot.mp4 -o ./Down
```

### Environment variables

_No response_

### Extra information

_No response_

### Code of Conduct

chxmberland:
### Linked issues

Fixes #486 

### Summarize your change.

The rvio library allows the usage of multiple threads, which causes issues as some functions are not thread safe. In this PR, I've added a thread lock to prevent double-freeing a pointer (line 473 and 484) using mutex.

Any user trying to use the `rvio` library with more than one thread caused a segmentation fault because of multiple double-frees. 

The reason for this was that the `rvio` library did not initialize `TwkFB::GenericIO`, which meant that the old thread locks were not initialized. This code was written before mutex was added to std as well, so the old thread locks have been replaced with `mutex`, since it is now included in the `std` library. This saves a significant amount of code. 

Tested the following command (used to replicate the bug initially) on MacOS.

`./rvio -rthreads 2 clip1.mov -o out.#.jpg`

See change summary.

Error before changes:

`dlopen(v??b??, 0x0001): tried: 'v??b??' (no such file), '/System/Volumes/Preboot/Cryptexes/OSv??b??' (no such file), '/Users/<user>/Dev/OpenRV/_build/stage/app/RV.app/Contents/Frameworks/v??b??' (no such file), '/Users/<user>/Dev/OpenRV/_build/stage/app/RV.app/Contents/lib/v??b??' (no such file),  '/usr/lib/v??b??' (no such file, not in dyld cache), 'v??b??' (no such file), '/usr/local/lib/v??b??' (no such file),  '/usr/lib/v??b??' (no such file, not in dyld cache) WARNING:  failed to load zsh: segmentation fault  rvio -rthreads 2 ~/Downloads/friendly-lot.mp4 -o ./Down`


src/lib/image/TwkFB/IO.cpp
```diff
@@ -463,7 +466,14 @@ GenericIO::addPlugin(FrameBufferIO* plugin)
 FrameBufferIO*
 GenericIO::loadFromProxy(Plugins::iterator i)
 {
+    std::lock_guard<std::mutex> guard(plugin_mutex);
```
bernie-laberge, line 469: Since the mutex is only used here, you could declare it static inside this GenericIO::loadFromProxy() method:
`static std::mutex plugin_mutex;
std::lock_guard<std::mutex> guard(plugin_mutex);`

bernie-laberge: You need to add the same mechanism in GenericIO::loadFromProxy()

src/lib/image/TwkFB/IO.cpp
```diff
@@ -463,7 +466,14 @@ GenericIO::addPlugin(FrameBufferIO* plugin)
...
+
     ProxyFBIO* pio = dynamic_cast<ProxyFBIO*>(*i);
+    if (pio == nullptr)
```
bernie-laberge, line 472: There is a case where it would be perfectly normal to have a pio==nullptr here because of multithreading.
Not an error. But actually normal behaviour.  :-)
Do you know what this GenericIO::loadFromProxy() mechanism is about ?
Hint: It has to do with the file PlugIns/ImageFormats/formats.gto

bernie-laberge: Moreover, there is another mutex here so your new one shouldn't be needed. 
What is going on ?
https://github.com/AcademySoftwareFoundation/OpenRV/blob/4639285133d740987fbdad0643256bd11667ee88/src/lib/image/TwkFB/IO.cpp#L495

bernie-laberge: It looks like loadFromProxy() is called simultaneously in 2 separate threads, defying the mutex above.
This is a Busytown mystery !

bernie-laberge: Note: You can add `-debug plugins` to get additional debug info.
Example:
`./Contents/MacOS/rvio -debug plugins -rthreads 2 /Users/labergb/media/mini_clip_1.mp4 -o /Users/labergb/tmp/toDelete/test/out.#.jpg`

chxmberland:
Turns out, the thread lock that we were using was not working properly. We needed to switch to the mutex thread lock, which fixed all of the errors.

bernie-laberge:
@chxmberland : I think I know why the original mutex was not working: I think it was never initialized:
!!!GenericIO::findByExtension()-before lockerGuard()-m_locker=0x0, threadID: 0x309789000
!!!GenericIO::findByExtension()-after lockerGuard()-m_locker=0x0, threadID: 0x309789000
!!!GenericIO::findByExtension()-before lockerGuard()-m_locker=0x0, threadID: 0x309b8c000
!!!GenericIO::findByExtension()-after lockerGuard()-m_locker=0x0, threadID: 0x309b8c000
And the PluginsLockerGuard() was not doing guarding anything with a null pointer.

Do you know why the m_locker is not initialized in this scenario ?

src/lib/image/TwkFB/IO.cpp
```diff
@@ -16,12 +15,15 @@
...
 #include <dlfcn.h>
 #include <sys/types.h>
+#include <mutex>
 
 /* AJG - stricmp */
 #ifdef _MSC_VER
 #define strcasecmp _stricmp
 #endif
 
+std::mutex plugin_mutex;
```
bernie-laberge, line 25: You should make it `static` so its name does not accidentally collide with other objects.
`static std::mutex plugin_mutex;`

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -26,6 +27,8 @@
 #define strcasecmp _stricmp
 #endif
 
+std::mutex plugin_mutex;
```
bernie-laberge, line 30: Same comment as above: add static

bernie-laberge:
OK I think I've found the source of the issue.
That's because unlike rv, rvio never calls TwkFB::GenericIO::init() and TwkMovie::GenericIO::init() which Initialize the plugins statics.

https://github.com/AcademySoftwareFoundation/OpenRV/blob/980236409479ec28120d1b70e627dfcde929b624/src/bin/apps/rv/main.cpp#L378

bernie-laberge:
So this why your PR, by replacing the original never initialized m_locker mutex, fixes the segmentation fault issue. This is great. Plus it removes code. Less code less bugs.
But to be like RV, rvio should also initialize the plugin statics here:
https://github.com/AcademySoftwareFoundation/OpenRV/blob/980236409479ec28120d1b70e627dfcde929b624/src/bin/imgtools/rvio/main.cpp#L1035

    TwkFB::GenericIO::init(); // Initialize TwkFB::GenericIO plugins statics
    TwkMovie::GenericIO::init(); // Initialize TwkMovie::GenericIO plugins statics 

And also call the associated shutdowns before exiting.

bernie-laberge (approved):
Looks good ! Thank you @chxmberland !
The PR description needs to be adapted to reflect the new understanding of this issue. Thanks!

geffrak (approved):

PR #497, opened 2024-06-21
Fix rotation metadata loading incorrectly with FFMPEG 6

chxmberland:
### Summarize your change.

Added an additional check to another metadata field when loading movie clips to ensure that rotation metadata is captured correctly.

In FFmpeg 6.0, rotation metadata can be loaded incorrectly without this check.

macOS Sonoma 14.5

- Added an additional check to the `side_data` field when movies are loaded in to RV
- Covered additional cases of negative rotation metadata, as this was not covered already

Roatate-Metadata-Missing

### DIFF
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -1491,7 +1491,7 @@ MovieFFMpegReader::snagOrientation(VideoTrack* track)
     rotation = (rotEntry) ? atoi(rotEntry->value) : 0;
 
     // If rotation metadata not in metadata, try to get it from side data
-    if (rotation == 0 && videoStream->nb_side_data > 0) {
+    if (!rotEntry && videoStream->nb_side_data > 0) {
         double sideRotation = 0;
         for (int i = 0; i < videoStream->nb_side_data; ++i) {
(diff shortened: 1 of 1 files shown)


src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -1484,20 +1485,38 @@ MovieFFMpegReader::snagOrientation(VideoTrack* track)
 
     int rotation = 0;
     AVDictionaryEntry *rotEntry;
+
+    // Trying to get rotation from metadata
     rotEntry = av_dict_get(videoStream->metadata, "rotate", NULL, 0);
     rotation = (rotEntry) ? atoi(rotEntry->value) : 0;
+
+    // If rotation metadata not in metadata, try to get it from side data
+    if (rotation == 0 && videoStream->nb_side_data > 0) {
```
bernie-laberge, line 1494: What if the "rotate" metadata exists and its value is equal to zero  ? atoi(rotEntry->value)==0 ?

Perhaps we should replace the rotation==0 condition with !rotEntry ?

chxmberland: On it!

bernie-laberge (approved):

rogernelson (approved):

PR #500, opened 2024-06-30
Modified to allow arbitrary metadata to show up in RV.

richardssam:
MKV in particular allows arbitrary key/value metadata pairs. The previous code only allowed a known list of parameters to be passed through to RV. This version has a list of parameters that should NOT be passed through.

I've tested loading mov and mkv files on windows.

Below is an example of getting a field called "samfield" to show up in the info display.

image

This metadata was added using the following ffmpeg command:
```
ffmpeg -i chimera_cars_srgb-test_mov-mjpeg.mov -metadata samfield=samtest -c copy chimera_cars_srgb-test_mov-mjpeg.mkv
```

### DIFF
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -840,7 +840,7 @@ isMetadataField(string check)
 
 
 bool
-ignoreMetadataField(string check)
+ignoreMetadataField(const string& check)
 {
     for (const char** p = ignoreMetadataFieldsArray; *p; p++)
     {
@@ -1252,6 +1252,14 @@ MovieFFMpegReader::openAVCodec(int index, AVCodecContext** avCodecContext)
         cout << "ERROR: MovieFFMpeg: Unsupported codec_id '" <<
             avStream->codecpar->codec_id << "' in " << m_filename << endl;
         return false;
+    }
+
+    if (avCodec->name == "libdav1d"){
+        const AVCodec *avCodec1 = avcodec_find_decoder_by_name("av1_nvdec");
+        if (avCodec1){
+            cerr << "WARNING: Using av1_nvdec codec rather than Dav1d\n";
+            avCodec = avCodec1; 
+        }
     }
 
     *avCodecContext = avcodec_alloc_context3(avCodec);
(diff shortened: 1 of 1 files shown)


src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -832,6 +838,17 @@ isMetadataField(string check)
     return false;
 }
 
+
+bool
+ignoreMetadataField(string check)
```
bernie-laberge, line 843: Minor detail : could be passed by reference (string&  check) 
That being said, I know it was exactly like that before ! :-)

bernie-laberge (approved):
Looks good to me. 
Thank you ! 
Your contribution is very much appreciated !

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
richardssam: No reason why not, I'll change it tomorrow AM.

richardssam: Should be addressed.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -5467,26 +5537,19 @@ MovieFFMpegIO::MovieFFMpegIO(CodecFilterFunction codecFilter,
...
-    //     {
-    //         int poolSize = atoi(c);
-    //     }
-    //     if (poolSize > 0) globalContextPool = new ContextPool(poolSize);
-    //     else
-    //     {
-    //         report("Disabling mio_ffmpeg context thread pool.", true);
-    //     }
-    // }
+    if (!globalContextPool)
```
bernie-laberge, line 5540: I believe this is a merge issue here.
This FFmpeg context pool should be commented out as it no longer works with FFmpeg 6+.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
bernie-laberge: Thanks !

richardssam:
We can close this, it should be a duplicate of #504, but with a clean commit.

PR #501, opened 2024-07-03
Automatically set QT_HOME variable if it is not already set

chxmberland:
### Linked issues

#502 

### Summarize your change.

When RV is built, we now search for QT_HOME (the QT installation location), rather than assuming it. 

Some customers have reported having issues with QT_HOME not being set correctly due to the static assumption we have made about the installation location of QT.

QT_HOME is set properly, regardless of QT installation location, for macOS Sonoma 14.5 and Windows 11. L 

### DIFF
--- a/rvcmds.sh
+++ b/rvcmds.sh
@@ -21,67 +21,44 @@ if [[ $SOURCED == 0 ]]; then
   exit 1
 fi
 
-QT_VERSION="${QT_VERSION:-5.15.2}"
-
 # Linux
 if [[ "$OSTYPE" == "linux"* ]]; then
   CMAKE_GENERATOR="${CMAKE_GENERATOR:-Ninja}"
-
-  # Searching for a Qt installation path if it has not already been set
-  if [ -z "$QT_HOME" ]; then
-    echo "Searching for Qt installation..."
-    QT_HOME=$(mdfind -name Qt | grep gcc_64 | head -n 1 | sed 's|\(.*clang_64\).*|\1|')
-    if [ -z "$QT_HOME" ]; then
-      echo "Could not find Qt installation. Please set QT_HOME to the correct path in .bashrc or .zshrc."
-      exit 1
-    else 
-      echo "Found Qt installation at $QT_HOME"
-    fi
-  else 
-    echo "Using Qt installation set at $QT_HOME"
-  fi
+  QT_REGEX=$(mdfind -name Qt | grep gcc_64 | head -n 1 | sed 's|\(.*clang_64\).*|\1|')
 
 # MacOS
 elif [[ "$OSTYPE" == "darwin"* ]]; then
   CMAKE_GENERATOR="${CMAKE_GENERATOR:-Ninja}"
-
-  # Searching for a Qt installation path if it has not already been set
-  if [ -z "$QT_HOME" ]; then
-    echo "Searching for Qt installation..."
-    QT_HOME=$(mdfind -name Qt | grep clang_64 | head -n 1 | sed 's|\(.*clang_64\).*|\1|')
-    if [ -z "$QT_HOME" ]; then
-      echo "Could not find Qt installation. Please set QT_HOME to the correct path in .bashrc or .zshrc."
-      exit 1
-    else 
-      echo "Found Qt installation at $QT_HOME"
-    fi
-  else 
-    echo "Using Qt installation set at $QT_HOME"
-  fi
+  QT_REGEX=$(mdfind -name Qt | grep clang_64 | head -n 1 | sed 's|\(.*clang_64\).*|\1|')
 
 # Windows
 elif [[ "$OSTYPE" == "msys"* ]]; then
   CMAKE_GENERATOR="${CMAKE_GENERATOR:-Visual Studio 17 2022}"
-
-  # Searching for a Qt installation path if it has not already been set
-  if [ -z "$QT_HOME" ]; then
-    echo "Searching for Qt installation..."
-    QT_HOME=$(find c:/Qt -name 'qmake.exe' | grep msvc2019_64 | head -n 1 | sed 's|/bin/qmake.exe||') # Assuming use of msvc2019_64
-    if [ -z "$QT_HOME" ]; then
-      echo "Could not find Qt installation. Please set QT_HOME to the correct path in your environment variables."
-    else 
-      echo "Found Qt installation at $QT_HOME"
-    fi
-  else 
-    echo "Using Qt installation set at $QT_HOME"
-  fi
-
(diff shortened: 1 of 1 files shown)


rvcmds.sh
```diff
@@ -22,15 +22,60 @@ if [[ $SOURCED == 0 ]]; then
...
+# Linux
 if [[ "$OSTYPE" == "linux"* ]]; then
   CMAKE_GENERATOR="${CMAKE_GENERATOR:-Ninja}"
-  QT_HOME="${QT_HOME:-$HOME/Qt/${QT_VERSION}/gcc_64}"
+
+  # Searching for a Qt installation path if it has not already been set
+  if [ -z "$QT_HOME" ]; then
+    echo "Searching for Qt installation..."
+    QT_HOME=$(mdfind -name Qt | grep gcc_64 | head -n 1 | sed 's|\(.*clang_64\).*|\1|')
+    if [ -z "$QT_HOME" ]; then
```
bernie-laberge, line 34: It would be preferable to prevent duplicating this code for all platform.

bernie-laberge: Perhaps something like setting QT_HOME_REGEX for each platform and then after the platform specific stuff, execute the command to fill QT_HOME and have the generic code only once.
Something like:
```
   CMAKE_GENERATOR="${CMAKE_GENERATOR:-Ninja}"
   QT_HOME_REGEX=mdfind -name Qt | grep gcc_64 | head -n 1 | sed 's|\(.*clang_64\).*|\1|'
elif [[ "$OSTYPE" == "darwin"* ]]; then
  CMAKE_GENERATOR="${CMAKE_GENERATOR:-Ninja}"
  QT_HOME_REGEX=mdfind -name Qt | grep clang_64 | head -n 1 | sed 's|\(.*clang_64\).*|\1|'
elif [[ "$OSTYPE" == "msys"* ]]; then
  CMAKE_GENERATOR="${CMAKE_GENERATOR:-Visual Studio 17 2022}"
    QT_HOME_REGEX=find c:/Qt -name 'qmake.exe' | grep msvc2019_64 | head -n 1 | sed 's|/bin/qmake.exe||'
  WIN_PERL="${WIN_PERL:-c:/Strawberry/perl/bin}"
  CMAKE_WIN_ARCH="${CMAKE_WIN_ARCH:--A x64}"
  SETUPTOOLS_USE_DISTUTILS=stdlib
else
  echo "OS does not seem to be linux, darwin or msys. Exiting."
  exit 1
fi

# Searching for a Qt installation path if it has not already been set
  if [ -z "$QT_HOME" ]; then
    echo "Searching for Qt installation..."
    QT_HOME=$(QT_HOME_REGEX)
    if [ -z "$QT_HOME" ]; then
      echo "Could not find Qt installation. Please set QT_HOME to the correct path in .bashrc or .zshrc."
      exit 1
    else 
      echo "Found Qt installation at $QT_HOME"
    fi
  else 
    echo "Using Qt installation set at $QT_HOME"

```

rvcmds.sh
```diff
@@ -21,24 +21,46 @@ if [[ $SOURCED == 0 ]]; then
...
   CMAKE_WIN_ARCH="${CMAKE_WIN_ARCH:--A x64}"
   SETUPTOOLS_USE_DISTUTILS=stdlib
+
 else
   echo "OS does not seem to be linux, darwin or msys. Exiting."
   exit 1
 fi
 
+# Searching for a Qt installation path if it has not already been set
+if [ -z "$QT_HOME" ]; then
```
bernie-laberge, line 48: Nice !!!

rvcmds.sh
```diff
@@ -21,24 +21,46 @@ if [[ $SOURCED == 0 ]]; then
...
 elif [[ "$OSTYPE" == "darwin"* ]]; then
   CMAKE_GENERATOR="${CMAKE_GENERATOR:-Ninja}"
-  QT_HOME="${QT_HOME:-$HOME/Qt/${QT_VERSION}/clang_64}"
+  QT_REGEX=$(mdfind -name Qt | grep clang_64 | head -n 1 | sed 's|\(.*clang_64\).*|\1|')
+
+# Windows
 elif [[ "$OSTYPE" == "msys"* ]]; then
   CMAKE_GENERATOR="${CMAKE_GENERATOR:-Visual Studio 17 2022}"
-  QT_HOME="${QT_HOME:-c:/Qt/${QT_VERSION}/msvc2019_64}"
+  QT_REGEX=$(find c:/Qt -name 'qmake.exe' | grep msvc2019_64 | head -n 1 | sed 's|/bin/qmake.exe||') # Assuming use of msvc2019_64
```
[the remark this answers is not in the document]
chxmberland: Just made some changes, had to add _some_ repetitive code, but not much.

bernie-laberge (approved):
Good job Benjamin !
Thank you !

rvcmds.sh
```diff
@@ -21,24 +21,50 @@ if [[ $SOURCED == 0 ]]; then
...
 else
   echo "OS does not seem to be linux, darwin or msys. Exiting."
   exit 1
 fi
 
+# Searching for a Qt installation path if it has not already been set
+if [ -z "$QT_HOME" ]; then
+  echo "Searching for Qt installation..."
+
+  if [[ "$OSTYPE" == "linux"* ]]; then
```
rogernelson, line 48: There are some assumptions here that on *nix systems Qt is in the homedir and that on Windows its on c:\, so it's not complete foolproof.  But we were also making those assumptions before, so overall this is still a nice improvement.

rogernelson (approved):
Nice work.

rvcmds.sh
geffrak: @bernie-laberge, someone with both Qt 6 and Qt 5 might end in a different situation than you what expect. We should check against 5.15, which is the only thing RV supports.

Also, on Mac, the/macOS (fat binary) works well, too, not just clang_64. I've been using this for a while now. It doesn't break the Intel build and is required for the Apple Silicon one. 

Finally, 5.15.2 is the more recent version that is provided by Qt without a commercial licence, that's why it was the default value.

geffrak:
@bernie-laberge, someone with both Qt 6 and Qt 5 might end in a different situation than you what expect. We should check against 5.15, which is the only thing RV supports.

Also, on Mac, the/macOS (fat binary) works well, too, not just clang_64. I've been using this for a while now. It doesn't break the Intel build and is required for the Apple Silicon one.

Finally, 5.15.2 is the more recent version that is provided by Qt without a commercial licence, that's why it was the default value.

bernie-laberge:
Good catch @geffrak . I'll create a ticket to address those issues you raised.
Thanks!

PR #504, opened 2024-07-05
Modified to allow arbitrary metadata to show up in RV (v2)

richardssam:
KV in particular allows arbitrary key/value metadata pairs. The previous code only allowed a known list of parameters to be passed through to RV. This version has a list of parameters that should NOT be passed through.

Describe what you have tested and on which operating system.
I've tested loading mov and mkv files on windows.

Below is an example of getting a field called "samfield" to show up in the info display.
image

This metadata was added using the following ffmpeg command:
```
ffmpeg -i chimera_cars_srgb-test_mov-mjpeg.mov -metadata samfield=samtest -c copy chimera_cars_srgb-test_mov-mjpeg.mkv
```


bernie-laberge (approved):
Beautiful ! 
Thank you Sam !

geffrak (approved):

PR #509, opened 2024-07-10
Search for QT 5.15 to avoid confusion with users who have multiple QT…

chxmberland:
### Linked issues

No linked issues, but created based on [this](https://github.com/AcademySoftwareFoundation/OpenRV/pull/501#issuecomment-2214457140) review of PR #501 

### Summarize your change.

Ensure that Open RV searches specifically for QT 5.15.* during setup, to avoid confusion with other QT versions that the user may have downloaded.

Tested on macOS Sonoma 14.5 and Windows 11, but testing must be done with Linux before merging.

- Simple changes to rvcmds.sh

### DIFF
--- a/rvcmds.sh
+++ b/rvcmds.sh
@@ -46,11 +46,11 @@ if [ -z "$QT_HOME" ]; then
   echo "Searching for Qt installation..."
 
   if [[ "$OSTYPE" == "linux"* ]]; then
-    QT_HOME=$(find ~/Qt* -type d -maxdepth 4 -path '*/gcc_64' | head -n 1)
+    QT_HOME=$(find ~/Qt/5.15* -type d -maxdepth 4 -path '*/gcc_64' | head -n 1)
   elif [[ "$OSTYPE" == "darwin"* ]]; then
-    QT_HOME=$(find ~/Qt* -type d -maxdepth 4 -path '*/clang_64' | head -n 1)
+    QT_HOME=$(find ~/Qt/5.15* -type d -maxdepth 4 -path '*/clang_64' | head -n 1)
   elif [[ "$OSTYPE" == "msys"* ]]; then
-    QT_HOME=$(find c:/Qt* -type d -maxdepth 4 -path '*/msvc2019_64' | head -n 1)
+    QT_HOME=$(find c:/Qt/5.15* -type d -maxdepth 4 -path '*/msvc2019_64' | head -n 1)
   fi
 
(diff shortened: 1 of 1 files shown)


rvcmds.sh
```diff
@@ -46,11 +46,11 @@ if [ -z "$QT_HOME" ]; then
   echo "Searching for Qt installation..."
 
   if [[ "$OSTYPE" == "linux"* ]]; then
-    QT_HOME=$(find ~/Qt* -type d -maxdepth 4 -path '*/gcc_64' | head -n 1)
+    QT_HOME=$(find ~/Qt/5.15* -type d -maxdepth 4 -path '*/gcc_64' | head -n 1)
```
geffrak, line 49: 1) Ideally, in this example, it would grab 5.15.14.

```
$ find ~/Qt/5.15* -type d -maxdepth 4 -path '*/clang_64'
/Users/geffrak/Qt/5.15.11/clang_64
/Users/geffrak/Qt/5.15.14/clang_64
/Users/geffrak/Qt/5.15.2/clang_64
```

But it won't with the current code.

```
$ find ~/Qt/5.15* -type d -maxdepth 4 -path '*/clang_64' | head -n 1
/Users/geffrak/Qt/5.15.11/clang_64
```


2) Also, macos should be preferred to clang_64, if available. (It's the Universal2 package of Qt)

```
$ find ~/Qt/5.15* -type d -maxdepth 4 -path '*/macos'
/Users/geffrak/Qt/5.15.11/macos
/Users/geffrak/Qt/5.15.14/macos

$ find ~/Qt/5.15* -type d -maxdepth 4 -path '*/clang_64'
/Users/geffrak/Qt/5.15.11/clang_64
/Users/geffrak/Qt/5.15.14/clang_64
/Users/geffrak/Qt/5.15.2/clang_64
```

On Mac, it might be more complex than a one-liner.

geffrak (changes requested):

rvcmds.sh
chxmberland: So it is preferable to use the most recent patch of 5.15._x_?

geffrak: Yes. 14 have more bugfixes than 2. There's no versions of 5.15 below 10 that ships a macos package, so you might be good to go with the current code.

geffrak (approved):

geffrak (approved):

bernie-laberge (approved):

PR #510, opened 2024-07-11
Fix Windows crash on exit with upcoming NDI video output plugin

bernie-laberge:
### Linked issues
NA

### Summarize your change.

This commit makes sure that the allocated video modules are released when RV exits.
To do so, the TwkApp::VideoModules container which previously contained raw pointers (std::vector<VideoModule*>) which were never released on exit, has been converted to std::vector<std::shared_ptr<VideoModule>>

This missing TwkApp::VideoModules cleanup on exit, was causing a crash on Windows when the upcoming NDI video output plugin was unloaded. 
More specifically: The NDI crash was caused when trying to unload the NDI dll without having called the NDIlib_destroy().
Note that NDIlib_initialize() is called when allocating an NDI video module while  the NDIlib_destroy() is called when releasing the NDI video module. So it became essential to fix this TwkApp::VideoModules  cleanup on exit to prevent a crash on Windows.

Successfully tested on macOS and Windows.

### DIFF
--- a/src/lib/app/TwkApp/Application.cpp
+++ b/src/lib/app/TwkApp/Application.cpp
@@ -26,6 +26,9 @@ Application::Application() : TwkUtil::Notifier()
 Application::~Application()
 {
     m_app = 0;
+
+    // Make sure to release all the video modules before we unload the plugins
+    m_videoModules.clear();
 }
 
 void
@@ -58,7 +61,7 @@ Application::receive(Notifier* originator,
 VideoModule* 
 Application::primaryVideoModule() const
 {
-    return m_videoModules.empty() ? 0 : m_videoModules.front();
+    return m_videoModules.empty() ? 0 : m_videoModules.front().get();
 }
 
 void
--- a/src/lib/ip/IPCore/IPCore/Application.h
+++ b/src/lib/ip/IPCore/IPCore/Application.h
@@ -44,7 +44,7 @@ class Application : public TwkApp::Application
     typedef std::vector<std::string>                    StringVector;
     typedef std::map<std::string,std::string>           PropEnvMap;
     typedef TwkApp::VideoModule                         VideoModule;
-    typedef std::vector<VideoModule*>                   VideoModules;
+    typedef std::vector<std::shared_ptr<VideoModule>>   VideoModules;
     typedef size_t                                      DispatchID;
     typedef std::function<void(DispatchID)>             DispatchCallback;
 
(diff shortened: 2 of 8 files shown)


src/lib/app/TwkApp/Application.cpp
```diff
@@ -26,6 +26,9 @@ Application::Application() : TwkUtil::Notifier()
 Application::~Application()
 {
     m_app = 0;
+
+    // Make sure to release all the video modules before we unload the plugins
+    m_videoModules.clear();
```
rogernelson, line 31: Although this is not harmful, you don't need to do it.  It will happen automatically in the dtor of the vector.

src/lib/ip/IPCore/IPCore/Application.h
```diff
@@ -44,7 +44,7 @@ class Application : public TwkApp::Application
     typedef std::vector<std::string>                    StringVector;
     typedef std::map<std::string,std::string>           PropEnvMap;
     typedef TwkApp::VideoModule                         VideoModule;
-    typedef std::vector<VideoModule*>                   VideoModules;
+    typedef std::vector<std::shared_ptr<VideoModule>>   VideoModules;
```
rogernelson, line 47: Nice! Love to see our code being modernized (and safer).  One point though, does it really need to be shared? I can't anywhere else owning the pointers besides the vector.  If so, you could consider an std::unique_ptr instead.

rogernelson (approved):
Nice work Bernie!

geffrak (approved):

src/lib/ip/IPCore/IPCore/Application.h
bernie-laberge: Good point ! It doesn't need to be shared. I just wanted to manage them. I'll use std::unique_ptr instead.
Thank you Roger!

bernie-laberge: Problemo: I get the following compile error when using std::unique_ptr instead of std::shared_ptr because by ysing the "for( auto module" construct I am actually sharing the smart pointer:

```
/Users/labergb/openrv_bernie2/OpenRV/src/lib/app/RvCommon/RvTopViewToolBar.cpp:488:37: error: call to implicitly-deleted copy constructor of 'const std::unique_ptr<TwkApp::VideoModule>'
                    for (const auto module : RvApp()->videoModules())
```

With the aim of keeping the nice "for( auto module" construct, I think I will stick with the std::shared_ptr.

rogernelson: That's fine. You just needed a `const auto & module` (the reference), but I am okay with leaving it shared.

PR #514, opened 2024-07-15
Build RV using a Python virtual environment

chxmberland:
### Linked issues

N/A

### Summarize your change.

The Python version included with MacOS Sonoma 14.5 adopted [PEP 668](https://peps.python.org/pep-0668/), which is meant to prevent conflicts between system level Python packages with project specific Python packages, like what is included in `requirements.txt`. 

This means that the user will encounter an error when attempting to install Python requirements unless they use the tag `--break-system-packages` tag. By using a virtual environment, system-level conflicts are avoided and the user has a safer build of Open RV. 

Mac OS Sonoma 14.5 and Windows 11.

PEP 668 error without a virtual environment, triggered by running `rvbuild`.

image


geffrak (changes requested):
Please don't name it pyenv. Pyenv is a different thing and might confuse people (I got confused at first).

venv (or .venv) is is fine.

Also, can you run rvenv multiple times without errors?

rvcmds.sh
```diff
@@ -73,8 +73,7 @@ RV_BUILD_PARALLELISM="${RV_BUILD_PARALLELISM:-$(python3 -c 'import os; print(os.
 
 # ALIASES: Basic commands
 
-# TODO: Get whichever Python version the user has installed
-alias rvenv="python3.12 -m venv _pyenv && source _pyenv/bin/activate"
+alias rvenv="python3.12 -m venv _virtualenv && source _virtualenv/bin/activate"
```
geffrak, line 76: You can rename this for `python3 ...`

bernie-laberge (approved):

geffrak (approved):

eloisebrosseau:
Would you mind adding your changes to the README so that it reflects the contents of rvcmds.sh?

README.md
```diff
@@ -77,13 +83,15 @@ Use the `rvsetup` common build alias to run the bootstrap step.
 
 #### Manually
 
+See the **Using a Python Virtual Environment** before running any of the commands in this section. Python packages should be installed inside of a virtual environment.
```
bernie-laberge, line 86: I stumbled on that yesterday while setuping the new build agents: we should also add this note here I think: https://github.com/AcademySoftwareFoundation/OpenRV/blob/main/docs/build_system/config_macos.md#install-the-python-requirements

PR #515, opened 2024-07-17
Native apple arm64

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
All usage of `arch -x86_64` have been removed and OpenRV will support only native build - which means that it will build the architecture of the host. (x86_64 for Intel chipset and arm64 for Apple Silicon)

On Apple Silicon, you can still build for x86_64 using a terminal launched with Rosetta or  by changing the architecture of the current terminal with the command `arch -x86_64 $SHELL`.

I added a condition to not change the rpath for Numpy since it causes issues during the installation step.

Also, I did a clean up in the cmake file for OpenRV's dependencies.

Support native build for ARM64 

- [x] Tested OpenRV ARM64 build on Macbook M1.
- [x] Tested OpenRV Intel build on MacBook M1 using a terminal launched with Rosetta.
- [x] Tested OpenRV Intel build on a Intel Macbook.

n/a


cedrik-fuoco-adsk:
The code is mostly the same as the previous PR #461, but with optimization for the install step (files under cmake/install and the remove_absolute_rpath.py script).

geffrak (approved):

bernie-laberge (approved):

PR #518, opened 2024-07-18
feat: SG-43395: Add Noto font so that RV can render Mandarin characters

chxmberland:
### Linked issues

N/A

### Summarize your change.

A new font has been added to RV's UI with the capability to render Latin and Mandarin characters.

RV uses hardcoded fonts for it's the UI. The existing fonts did not have the capability to render mandarin characters, so a new font ([Google Noto](https://fonts.google.com/noto/specimen/Noto+Sans+SC)) needed to be added to give RV the capability of rendering mandarin glyphs. 

macOS Sonoma 14.5.

A change that should be ignored is in `noto.cpp`, which contains a byte array of a `ttf` file. This is how RV consumes fonts.

**Before:**

Snipaste_2024-07-11_11-20-11

**After:**

image

### DIFF
--- a/src/lib/ui/TwkGLText/TwkGLText/noto.h
+++ b/src/lib/ui/TwkGLText/TwkGLText/noto.h
@@ -5,10 +5,9 @@
 // 
 //******************************************************************************
 
-#ifndef __NOTO_H__
-#define __NOTO_H__
+#pragma once
+
+// Sourced from Noto Sans Simplified: https://fonts.google.com/noto/specimen/Noto+Sans+SC
 
 extern const unsigned char notoFont[];
 const unsigned int notoFontSize = 10560380;
-
-#endif    // End #ifdef __NOTO_H__
--- a/src/lib/ui/TwkGLText/TwkGLText/noto.h
+++ b/src/lib/ui/TwkGLText/TwkGLText/noto.h
@@ -1,0 +1,14 @@
+//******************************************************************************
+// Copyright (c) 2024 Autodesk Inc. All rights reserved.
+// 
+// SPDX-License-Identifier: Apache-2.0
+// 
+//******************************************************************************
+
+#ifndef __NOTO_H__
+#define __NOTO_H__
+
+extern const unsigned char notoFont[];
+const unsigned int notoFontSize = 10560380;
+
+#endif    // End #ifdef __NOTO_H__
--- a/src/lib/ui/TwkGLText/CMakeLists.txt
+++ b/src/lib/ui/TwkGLText/CMakeLists.txt
@@ -11,7 +11,7 @@ SET(_target
 )
 
 SET(_sources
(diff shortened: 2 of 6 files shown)


src/lib/ui/TwkGLText/TwkGLText/noto.h
```diff
@@ -0,0 +1,14 @@
...
+// Copyright (c) 2024 Autodesk Inc. All rights reserved.
+// 
+// SPDX-License-Identifier: Apache-2.0
+// 
+//******************************************************************************
+
+#ifndef __NOTO_H__
+#define __NOTO_H__
+
+extern const unsigned char notoFont[];
```
bernie-laberge, line 11: We should add a link to where this font is coming from.
Is it https://fonts.google.com/noto/specimen/Noto+Sans+SC ?

src/lib/ui/TwkGLText/TwkGLText/noto.h
```diff
@@ -0,0 +1,14 @@
+//******************************************************************************
+// Copyright (c) 2024 Autodesk Inc. All rights reserved.
+// 
+// SPDX-License-Identifier: Apache-2.0
+// 
+//******************************************************************************
+
+#ifndef __NOTO_H__
```
bernie-laberge, line 8: use `#pragma once` instead
(When RV was created, I don't think they could use it, which is why they used this older method)

geffrak (approved):
Cool! You are so inclusive @chxmberland :D

kohyuk91:
This may be a non related question,
but does this solve the 'only english' supported annotation text tool issue?

chxmberland:
> This may be a non related question, but does this solve the 'only english' supported annotation text tool issue?

It doesn't look like it. The annotation tool does not yet detect non-English keyboard options.

kohyuk91:
> > This may be a non related question, but does this solve the 'only english' supported annotation text tool issue?
> 
> It doesn't look like it. The annotation tool does not yet detect non-English keyboard options.

Thank you for your reply! Really wish that could be fixed too.

src/lib/ui/TwkGLText/TwkGLText/noto.h
pbergeron-adsk: @chxmberland @bernie-laberge - From the header of the file, it appears the noto.cpp file is indeed "Noto Sans SC" (Simplified Chinese). It should contain regular latin, cyrilic, greek letters as well as Hangul (modern Korean) Hiragana/Katakana (Japanese, but is missing Kanji symbols). 

This font file has interesting support for asian languages, but, there may be a better choice?
https://fonts.google.com/noto/specimen/Noto+Sans+Mono

The full Noto Sans Mono contains glyph support for up to 775 languages and 236 regions and I believe is a superset of  Noto Sans SC.  It's surely a beefer of a font file, but, if we want to add international characters, wouldn't the full font file be a better choice?

image

PR #521, opened 2024-07-24
Update config_linux_rocky9.md based on a clean Rocky 9 install

shidarin:
### Summarize your change.

With a clean(ish) install of Rocky Linux 9.4, I ran through the build instructions and updated where they fell short. It's clean(ish) because I had installed Maya 2024, MtoA, and the Deadline Cloud submitters.

Following the original instructions still results in missing libraries and errors. I referenced additional information I found in #6 or discovered the proper solution to get past it, and updated the documentation. 

Rocky Linux 9.4

>  Make sure to note the destination of the Qt install, as you will have to set the `QT_HOME` environment variable to this location's build dir.

This is actually an oversimplification, but I'm unsure if the problem is my fault. The online qt installer built Qt into `/opt/Qt/5.15.2/gcc_64` (the dir might not actually be `gcc_64`). Setting `QT_HOME` to `/opt/Qt/5.15.2/gcc_64` resolves the initial build errors, but encounters new ones where it seemingly ignores the set `QT_HOME` and tries to find the directories `/opt/Qt/5.15.2/resources` and `/opt/Qt/5.15.2/translations`. Renaming `/opt/Qt/5.15.2/gcc_64` to `/opt/Qt/5.15.2` and setting `QT_HOME` to the same resolved the issue.

IN THE END IT BUILT!

image

### DIFF
--- a/docs/build_system/config_linux_rocky9.md
+++ b/docs/build_system/config_linux_rocky9.md
@@ -32,7 +32,15 @@ sudo dnf config-manager --set-disabled devel
 
 ### GLU
 
-You may or may not have libGLU on your system depending on your graphics driver setup.  To know you can check if you have the lib /usr/lib64/libGLU.so.1.  If it's there you can skip this step, there's nothing further to do in this section.  If not, you can install the graphics drivers for your system. As as last resort, you can install a software (mesa) libGLU version, but you may not have ideal performance with this option.  To do so:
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
Looks good to me.
Thank you @shidarin !

geffrak (approved):

PR #523, opened 2024-07-28
Fix make_python.py on Apple Silicon + Sonoma 14.5

bernie-laberge:
### Linked issues
NA

### Summarize your change.

Fix the python3 build by specifying --with-openssl-rpath in addition to --with-openssl.

Since the merge of the Apple Silicon native build for Open RV, some users were experiencing a build break when build the RV_DEPS_PYTHON3 dependency:

This was reproduced on a Mac M1 running Sonoma 14.5 using the latest Open RV code:

```
Could not build the ssl module!
Python requires a OpenSSL 1.1.1 or newer

Executing ['make', 'install', '-j10', '-s'] from /Users/labergb/git5/rv/_build/RV_DEPS_PYTHON3/src
Traceback (most recent call last):
  File "/Users/labergb/git5/rv/OpenRV/src/build/make_python.py", line 713, in <module>
    install()
  File "/Users/labergb/git5/rv/OpenRV/src/build/make_python.py", line 646, in install
    ).check_returncode()
      ^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.6/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py", line 502, in check_returncode
    raise CalledProcessError(self.returncode, self.args, self.stdout,
subprocess.CalledProcessError: Command '['make', 'install', '-j10', '-s']' returned non-zero exit status 2.
ninja: build stopped: subcommand failed.

```

Successfully tested on M1+macOS=14.5. Will also test on other OSes.

### DIFF
--- a/src/build/make_python.py
+++ b/src/build/make_python.py
@@ -124,7 +124,9 @@ def get_python_interpreter_args(python_home: str) -> List[str]:
     python_interpreters = glob.glob(
         os.path.join(python_home, python_name_pattern), recursive=True
     )
-    python_interpreters += glob.glob(os.path.join(python_home, "bin", python_name_pattern))
+    python_interpreters += glob.glob(
+        os.path.join(python_home, "bin", python_name_pattern)
+    )
 
     # sort put python# before python#-config
     python_interpreters = sorted(
@@ -164,7 +166,9 @@ def patch_python_distribution(python_home: str) -> None:
             shutil.move(failed_lib, failed_lib.replace("_failed.so", ".so"))
     if OPENSSL_OUTPUT_DIR:
         if platform.system() == "Darwin":
-            openssl_libs = glob.glob(os.path.join(OPENSSL_OUTPUT_DIR, "lib", "lib*.dylib*"))
+            openssl_libs = glob.glob(
+                os.path.join(OPENSSL_OUTPUT_DIR, "lib", "lib*.dylib*")
+            )
             openssl_libs = [l for l in openssl_libs if os.path.islink(l) is False]
 
             python_openssl_libs = []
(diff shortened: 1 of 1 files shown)


src/build/make_python.py
```diff
@@ -440,7 +450,7 @@ def configure() -> None:
 
         if OPENSSL_OUTPUT_DIR:
             configure_args.append(f"--with-openssl={OPENSSL_OUTPUT_DIR}")
-
+            configure_args.append(f"--with-openssl-rpath={OPENSSL_OUTPUT_DIR}/lib")
```
geffrak, line 453: Does it sets it to the full path? It doesn't matter much while in stage but we need to make sure it's relocatable after the install stage.

geffrak (approved):

src/build/make_python.py
bernie-laberge: The fix is not working consistently anyhow. 
I only started to get this make_python.py install error when I upgraded my macOS from 14.0 to 14.5.
Good good good.

I stil get:
```
Could not build the ssl module!
Python requires a OpenSSL 1.1.1 or newer
Custom linker flags may require --with-openssl-rpath=auto

Executing ['make', 'install', '-j10', '-s'] from /Users/labergb/openrv_bernie3/OpenRV/_build/RV_DEPS_PYTHON3/src
Traceback (most recent call last):
  File "/Users/labergb/openrv_bernie3/OpenRV/src/build/make_python.py", line 729, in <module>
    install()
  File "/Users/labergb/openrv_bernie3/OpenRV/src/build/make_python.py", line 656, in install
    ).check_returncode()
      ^^^^^^^^^^^^^^^^^^
  File "/opt/homebrew/Cellar/python@3.11/3.11.6/Frameworks/Python.framework/Versions/3.11/lib/python3.11/subprocess.py", line 502, in check_returncode
    raise CalledProcessError(self.returncode, self.args, self.stdout,
subprocess.CalledProcessError: Command '['make', 'install', '-j10', '-s']' returned non-zero exit status 2.
ninja: build stopped: subcommand failed.
```

Back to the drawing board.

bernie-laberge: I know that the ssl module is patched anyhow afterwards. What I don't understand is why are we only seeing this make_python.py error now ?

PR #524, opened 2024-07-30
feat: SG-43396: Allow users to paste text into the annotations tool

chxmberland:
### Linked issues

N/A

### Summarize your change.

This change will allow users to paste text when annotating using `ctrl + v`.

This change was made in an effort to allow users to paste in East Asian languages into their annotations, which is a feature that has been requested for a very long time. Since it is far too complicated to directly support East Asian languages through keyboard inputs, this is the next best thing. 

This PR is paired with PR #518 and will only be effective _after_ PR #518 has been merged.

Tested in macOS Sonoma 14.5.

This **must** be tested in Windows and Linux, as the way you interact with clipboards changes across operating systems.


src/lib/app/RvCommon/RvApplication.cpp
```diff
@@ -1446,6 +1449,26 @@ RvApplication::putUrlOnClipboard(string url, string title, bool doEncode)
...
+{
+    QString text;
+    QClipboard *clipboard = QApplication::clipboard();
+
+    #if defined(PLATFORM_DARWIN)
+        text = clipboard->text(QClipboard::Clipboard);
+    #elif defined(PLATFORM_WINDOWS)
+        text = clipboard->text(QClipboard::Clipboard);
+    #else
+        text = clipboard->text(QClipboard::Selection);
```
bernie-laberge, line 1463: Why is the code different on Linux ? Can't it be the same code on all platforms ?

src/lib/app/RvCommon/MuUICommands.cpp
```diff
@@ -2050,25 +2060,6 @@ NODE_IMPLEMENTATION(putUrlOnClipboard, void)
     RvApp()->putUrlOnClipboard(url->c_str(),
                                title->c_str(),
                                doEncode);
-
-    /*
```
geffrak, line 2054: lol. thanks for the dead code removal

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -622,6 +622,24 @@ class: AnnotateMinorMode : MinorMode
         populateAnnotationList();
     }
 
+    method: getTextFromClipboard (void; Event event)
```
bernie-laberge, line 625: Detail: Perhaps rename to pasteTextFromClipboard() so that it is clear what the method actually does ?

bernie-laberge:
This is great @chxmberland !

src/lib/app/RvCommon/RvApplication.cpp
geffrak: Agree with Bernie. Looking at the doc, it seems like we want Clipboard everywhere... https://doc.qt.io/qt-6/qclipboard.html#Mode-enum

src/bin/imgtools/rvio/UICommands.cpp
```diff
@@ -279,6 +279,10 @@ void initUICommands(MuLangContext* context)
                                       new Param(c, "doEncode", "bool", Value(true)), 
                                       End),
 
+                         new Function(c, "getTextFromClipboard", getTextFromClipboard, None,
```
bernie-laberge, line 282: Detail: I believe the RV founders did not like the getxxx() methods. My guess is that they would have preferred : 
textFromClipboard()
I believe this is a coding philosophy. I will try to find references to this philosophy and get back to you. :-)

src/lib/app/RvCommon/RvApplication.cpp
chxmberland: Windows and MacOS both have a single clipboard, while Linux has two. The first clipboard in Linux is the primary selection clipboard, which stores text that has been highlighted. The second clipboard is accessed using explicit commands, like `ctrl + c`. So, we have to check to see if the first clipboard is empty before checking the second. However, I could make this simpler by doing something like...

```
#ifdef PLATFORM_LINUX 

text = clipboard->text(QClipboard::Selection);
if (text.isEmpty()) {
    text = clipboard->text(QClipboard::Clipboard);
}

#else

text = clipboard->text(QClipboard::Clipboard);
```

geffrak: I understand, but we could have Linux behaving like Mac and Windows by ignoring the selection clipboard.

src/lib/app/RvCommon/MuUICommands.cpp
chxmberland: NO WORKY!

src/bin/imgtools/rvio/UICommands.cpp
chxmberland: Just switched the name to `pasteTextFromClipboard`!

src/lib/app/RvCommon/RvApplication.cpp
chxmberland: I see! On it. That makes sense.

bernie-laberge (approved):

cedrik-fuoco-adsk (approved):

pbergeron-adsk (approved):
lgtm

PR #528, opened 2024-08-01
Add environment variable that allows users to disable loading video output plugins

chxmberland:
### Linked issues

N/A

### Summarize your change.

Add a simple environment variable check to see if the user would like to load video output plugins.

These are heavy plugins and actually slow dow the time it takes for RV to boot. If the user so chooses, they should have the option to disable them from being loaded.

Tested on MacOS 14.5, disabling these plugins increased loading speed by about 20% when opening a piece of media with Open RV/ 


bernie-laberge (approved):
Looks good to me.
Thanks

eloisebrosseau (approved):

cedrik-fuoco-adsk (approved):

PR #530, opened 2024-08-05
Add function to detect long paths for files on Windows

chxmberland:
### Linked issues

N/A

### Summarize your change.

Added a function that takes advantage of the features added in Windows 10, version 1607 and later (see [this](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry#enable-long-paths-in-windows-10-version-1607-and-later) documentation).

A user has requested that OpenRV/RV supports long file paths on Windows (longer than 260 characters).

Tested opening `.mov` files with paths longer than 260 characters on...

- Windows 11
- macOS Sonoma 14.5 (already worked, still works with these changes)


src/lib/base/TwkUtil/File.cpp
```diff
@@ -194,8 +222,17 @@ bool fileExists( const char *fname )
 #else
     struct stat statBuf;
 #endif
+
     int status = TwkUtil::stat( fname, &statBuf );
 
+#ifdef PLATFORM_WINDOWS
```
bernie-laberge, line 228: I would move this Windows specific section at the beginning of fileExists().
So that if the filePath.length()>260 we call this new windows helper and return right away.
This way you wouldn't have to call TwkUtil::stat() in that scenario.

src/lib/base/TwkUtil/File.cpp
```diff
@@ -194,8 +222,17 @@ bool fileExists( const char *fname )
...
+
+    string filePath = fname;
+    bool fileExists = (filePath.length() > 260) && detectLongPathInWindows(filePath);
```
bernie-laberge, line 231: About detectLongPathInWindows(): I would rename it. Sorry :-) !

Rationale:
If I only look at the name (detectLongPathInWindows()), it tells me that if the supplied path is a long path then it returns true. So if I provide it with a short path then it returns false ?

Perhaps something like 
fileExistsWithLongPathSupport()

bernie-laberge: Also don't forget to replace the `boost::filesystem::exists()` instances in the code (5 I think) with `TwkUtil::fileExists() `

bernie-laberge:
Great catch ! Thanks !

src/lib/base/TwkUtil/File.cpp
chxmberland: Good point! Will do

src/lib/base/TwkUtil/File.cpp
chxmberland: I like that name better as well! That makes more sense. We're okay replacing every instance of `boost::filesystem::exists()`?

src/lib/base/TwkUtil/File.cpp
```diff
@@ -53,6 +54,33 @@ string to_utf8(const wchar_t *wc)
...
+        FILE_ATTRIBUTE_NORMAL,
+        nullptr
+    );
+
+    if (fileHandle != INVALID_HANDLE_VALUE) {
+        CloseHandle(fileHandle);
+        return true;
+    } else {
+        DWORD error = GetLastError();
+        return error == ERROR_FILE_NOT_FOUND || error == ERROR_PATH_NOT_FOUND;
```
bernie-laberge, line 79: I believe it should return false here.
In any case, no matter what the error is, I would return false.

    if (fileHandle != INVALID_HANDLE_VALUE) {
        CloseHandle(fileHandle);
        return true;
    } else {
       return false;
    }

src/lib/base/TwkUtil/File.cpp
```diff
@@ -194,6 +222,15 @@ bool fileExists( const char *fname )
 #else
     struct stat statBuf;
 #endif
+
+#ifdef PLATFORM_WINDOWS
+
+    string filePath = fname; // Casting to string
+    bool fileExists = (filePath.length() > 260) && fileExistsWithLongPathSupport(filePath);
+    if fileExists return true;
```
bernie-laberge, line 230: The way I see it is that if you are on windows and the filepath is > 260 characters then you need to call your fileExistsWithLongPathSupport(). Otherwise TwkUtil::stat will not be reliable.
So how about:
```
if (filePath.length()>260) {
   return fileExistsWithLongPathSupport(filePath);
}
```

src/lib/base/TwkUtil/File.cpp
bernie-laberge: Yes. Let's always use TwkUtil::fileExists() from now on.

src/lib/base/TwkUtil/File.cpp
chxmberland: Ah I see what you mean, sorry about that!

src/lib/base/TwkUtil/File.cpp
```diff
@@ -53,6 +54,33 @@ string to_utf8(const wchar_t *wc)
...
+        OPEN_EXISTING,
+        FILE_ATTRIBUTE_NORMAL,
+        nullptr
+    );
+
+    if (fileHandle != INVALID_HANDLE_VALUE) {
+        CloseHandle(fileHandle);
+        return true;
+    } else {
+        DWORD error = GetLastError();
```
bernie-laberge, line 78: dead code ? I would remove it

bernie-laberge (approved):

cedrik-fuoco-adsk (approved):

src/lib/app/TwkApp/Bundle.cpp
```diff
@@ -9,6 +9,7 @@
 #include <TwkApp/Bundle.h>
 #include <sstream>
 #include <boost/filesystem.hpp>
+#include <TwkUtil/file.h>
```
markreidvfx, line 12: case sensitive typo here. This should have been "#include <TwkUtil/File.h>"

PR #532, opened 2024-08-09
Fix issue install step tries to delete rpath multiple times (MacOS ARM64)

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Fix an issue during the install step where it tries to delete a RPATH multiple time due to a universal library.
Keep a list of deleted rpath for a specific file and do not delete the same RPATH multiple time.

Install step was failing on MacOS ARM64

- [x] MacOS ARM64


bernie-laberge (approved):
Beautiful !
Thank you Cédrik !

eloisebrosseau (approved):

PR #544, opened 2024-08-21
Update dav1d version to 1.4.3

eloisebrosseau:
### Summarize your change.

Update the version library from 1.0.0 to 1.4.3.

Sam Richards (Disney) asked at the last Open RV TSC if we could update the libdav1d component from 1.0 to latest 1.4.3 which has better performances according to him.

The build step has been tested on all three platforms.

### DIFF
--- a/cmake/dependencies/dav1d.cmake
+++ b/cmake/dependencies/dav1d.cmake
@@ -12,14 +12,14 @@ SET(_target
 )
 
 SET(_version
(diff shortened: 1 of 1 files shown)


cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
Looks good to me.
Thanks!

PR #545, opened 2024-08-21
Default sequences replaced when comparing media from FPTR 

chxmberland:
### Linked issues

N/A

### Summarize your change.

This When a user opens media from Flow Production Tracking, and then compares it to another piece of media in Flow Production Tracking, the default sequences are overwritten. Meaning, the user can no longer access the clips that they initially opened from FPTR. This can interrupt workflows.

If this change is implemented, along with another PR into OpenRV, users will have the option to enable a parameter in `Flow Production Tracking / Session Prefs` that will prevent this from happening.

This PR, paired with a PR in Commercial RV, will resolve this issue. The changes made in this PR include a command that modifies a global preference to prevent media from being overwritten while comparing media. 

User wanted to be able to compare media without overwriting the original media. 

MacOS 14.5 Sonoma

### DIFF
--- a/src/lib/app/RvApp/Options.cpp
+++ b/src/lib/app/RvApp/Options.cpp
@@ -458,6 +458,8 @@ Options::Options()
     presentFormat = (char*)"";
     presentData = (char*)"";
 
+    addSourceToDefaultView = true;
+
 #ifdef PLATFORM_DARWIN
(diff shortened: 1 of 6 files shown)


pbergeron-adsk:
Taking over this task and incomplete PR, so closing this PR.

PR #550, opened 2024-08-30
Replace IODevIL file format with IOoiio

eloisebrosseau:
### Summarize your change.

Remove code related to the IODevIL library and add the missing file formats to OpenImageIO library instead. Note that the following image file formats are not covered by the OpenImageIO library at the moment:

- .cut (Dr. Halo Cut File)
- .cur (Palette)
- .lbm (Interlaced Bitmap)
- .lif (Homeworld File)
- .lmp(Doom Walls/Flats)
- .mdl (Half-Life Model)
- .pcd (PhotoCD)
- .pcx (ZSoft PCX)
- .wal (Quake2 Texture)

The annual ASWF license scan for Open RV came back highlighting that the DevIL library have a LGPL-2.1 license which is in conflict with the Apache-2 license of this project.

The different image file formats were tested on MacOS.


bernie-laberge (approved):
ASWF' OpenImageIO is the way to go!
Thanks!

cedrik-fuoco-adsk (approved):

PR #552, opened 2024-09-03
Remove libDevil from Notice.txt

eloisebrosseau:

The libDevil mention is no longer required in Notice.txt since the library has been removed [here](https://github.com/AcademySoftwareFoundation/OpenRV/pull/550).

### DIFF
--- a/Notice.txt
+++ b/Notice.txt
@@ -966,11 +966,7 @@ We are required to state that
 
 Libaio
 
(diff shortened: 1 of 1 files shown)


cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
Thank you !

PR #572, opened 2024-09-09
Update AJA SDK to 17.1

eloisebrosseau:
### Summarize your change.

- [x] Update the SDK version in aja.cmake and fix issues related with OS specific header files that are no longer copied and the addition of the mbedtls library that we need to link against
- [x] Replace code related to `ScanHardware()` and `GetDeviceInfoList()` since both functions have been deprecated

The SDK needs to be at the latest version.

Presentation mode with the AJA Io 4K Plus device with HDMI and SDI on MacOS.


cmake/dependencies/aja.cmake
```diff
@@ -7,19 +7,22 @@
 INCLUDE(ProcessorCount) # require CMake 3.15+
 PROCESSORCOUNT(_cpu_count)
 
-RV_CREATE_STANDARD_DEPS_VARIABLES("RV_DEPS_AJA" "16.2" "make" "")
-RV_SHOW_STANDARD_DEPS_VARIABLES()
```
cedrik-fuoco-adsk, line 11: Is there a reason to remove this? 

We use that macro on almost all `*.cmake` for dependencies. I updated a bunch of them when I did the work on ARM64. I think it would be good if the code between the dependencies is similar.

src/plugins/output/AJADevices/KonaVideoDevice.cpp
```diff
@@ -654,12 +655,6 @@ namespace AJADevices
         m_writeBufferIndex( 0 ),
         m_writeBufferCount( 0 )
   {
-    //
-    //  Default ringbuffer size is now 4, but allow setting of size by env var
-    //  just in case.
-    //
-    m_info = &reinterpret_cast<CNTV2DeviceScanner*>( m->deviceScan() )
-                  ->GetDeviceInfoList()[devicenum];
```
bernie-laberge, line 658: We might want to remove the comment on line 725 which references m_info since m_info is no longer a thing:
// XXX this 0 might need to be m_info->deviceIndex

bernie-laberge (approved):
I believe the build from scratch needs to be fixed but apart from that it looks good.
Thanks!

cmake/dependencies/aja.cmake
```diff
@@ -117,12 +135,26 @@ SET_PROPERTY(
...
+FILE(MAKE_DIRECTORY ${_aja_include_dir} ${_aja_ntv2_include_dir} ${_aja_ntv2_os_specific_include_dir} ${_mbedtls_include_dir})
 TARGET_INCLUDE_DIRECTORIES(
   aja::ntv2
-  INTERFACE ${_aja_include_dir} ${_aja_ntv2_include_dir}
+  INTERFACE ${_aja_include_dir} ${_aja_ntv2_include_dir} ${_aja_ntv2_os_specific_include_dir} ${_mbedtls_include_dir}
 )
 
+TARGET_LINK_LIBRARIES(
+  aja::ntv2 INTERFACE
+  ${_mbedtls_lib_dir}/libeverest.a ${_mbedtls_lib_dir}/libmbedcrypto.a ${_mbedtls_lib_dir}/libmbedtls.a ${_mbedtls_lib_dir}/libmbedx509.a ${_mbedtls_lib_dir}/libp256m.a)
```
bernie-laberge, line 156: Is this part applicable to all platforms ?

src/plugins/output/AJADevices/KonaVideoDevice.cpp
eloisebrosseau: I agree!

cmake/dependencies/aja.cmake
eloisebrosseau: The files that I looked at like `bmd.cmake` didn't have the two lines so I thought this was not the way to do it anymore. But I agree with being consistent between the dependencies and I put them back

cmake/dependencies/aja.cmake
eloisebrosseau: The .a files are only for MacOS and Linux, but it should be .lib for Windows. I corrected it, thanks!

.github/workflows/ci.yml
```diff
@@ -116,7 +116,7 @@ jobs:
           dnf install -y epel-release
           dnf config-manager --set-enabled ${{ matrix.extra_repo }} devel
           dnf groupinstall "Development Tools" -y
-          dnf install -y alsa-lib-devel autoconf automake avahi-compat-libdns_sd-devel bison bzip2-devel cmake-gui curl-devel flex gcc gcc-c++ git libXcomposite libXi-devel libaio-devel libffi-devel nasm ncurses-devel nss libtool libxkbcommon libXcomposite libXdamage libXrandr libXtst libXcursor mesa-libOSMesa mesa-libOSMesa-devel meson openssl-devel patch pulseaudio-libs pulseaudio-libs-glib2 ocl-icd ocl-icd-devel opencl-headers python3 python3-devel qt5-qtbase-devel readline-devel sqlite-devel tcl-devel tcsh tk-devel yasm zip zlib-devel wget
```
cedrik-fuoco-adsk, line 119: You had the right package, but it should be devel package. I looked on my Rocky, and it's `systemd-devel` that gives you `/usr/include/libudev.h`

cedrik-fuoco-adsk: We should add that new package to the documentation for Rocky Linux.

eloisebrosseau: Thank you! And yes, I wanted to confirm that I had the right package before adding it to the documentation, but it's done now!

cmake/dependencies/aja.cmake
```diff
@@ -95,14 +123,13 @@ EXTERNALPROJECT_ADD(
   DOWNLOAD_EXTRACT_TIMESTAMP TRUE
   SOURCE_DIR ${RV_DEPS_BASE_DIR}/${_target}/src
   INSTALL_DIR ${_install_dir}
-  CONFIGURE_COMMAND ${CMAKE_COMMAND} -G Ninja -DCMAKE_INSTALL_PREFIX=${_install_dir} -DCMAKE_OSX_ARCHITECTURES=${CMAKE_OSX_ARCHITECTURES}
-                    -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DAJA_BUILD_APPS=OFF ${RV_DEPS_BASE_DIR}/${_target}/src
-  # Not using _cmake_build_command and _cmake_install_command since the build dir need to change.
-  BUILD_COMMAND ${CMAKE_COMMAND} --build ${RV_DEPS_BASE_DIR}/${_target}/src --config ${CMAKE_BUILD_TYPE} -j${_cpu_count}
-  INSTALL_COMMAND ${CMAKE_COMMAND} --install ${RV_DEPS_BASE_DIR}/${_target}/src --prefix ${_install_dir} --config ${CMAKE_BUILD_TYPE}
+  CONFIGURE_COMMAND ${CMAKE_COMMAND} -DCMAKE_INSTALL_PREFIX=${_install_dir} -DCMAKE_OSX_ARCHITECTURES=${CMAKE_OSX_ARCHITECTURES}
+                    -DCMAKE_BUILD_TYPE=${CMAKE_BUILD_TYPE} -DAJA_BUILD_APPS=OFF -DAJANTV2_DISABLE_DEMOS=ON ${RV_DEPS_BASE_DIR}/${_target}/src
```
cedrik-fuoco-adsk, line 127: Question: Do we need `libajantv2 tools` executables? Otherwise, we could disable them.

cedrik-fuoco-adsk (approved):

PR #574, opened 2024-09-10
Create a different build folder for debug

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Create a debug folder when building in debug. (`_build_debug` vs `_build`)

It can cause problem if someone switch between built type (having mixed of debug and release).

- [x] Rocky Linux 8


bernie-laberge (approved):
Superb.
Thanks!

PR #575, opened 2024-09-12
Fix issue with venv on Windows

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
The Python virtual environment was not functioning correctly through rvcmd.sh.
Changed the path to activate the virtual environment based on the OS.

- [x] Windows


bernie-laberge (approved):
Thank you Cédrik !

PR #601, opened 2024-10-01
Fix OTIO tests

eloisebrosseau:
### Summarize your change.

1.  Replace clip_if() with find_clips() since it has been deprecated in OTIO
2. Fix _create_transition() where `post_item` was undefined and `if pre_item.source_range` was unhashable

This was preventing `otio_tests::export_multi_clip_with_transition()` and `otio_tests::round_trip()` from passing.

The tests were done on MacOS.


rogernelson (approved):
Thanks for these.

bernie-laberge (approved):
LGTM

PR #602, opened 2024-10-10
Add RV as a participant using the new Live Review schema

eloisebrosseau:
### Summarize your change.

The `read_otio_string` function was added to parse the OTIO object containing the timeline sent by the Review App when RV is joining a new session as a participant. Since the timeline contains some annotation fields, the annotation schema was added in order to let OTIO know that it is a known schema during the deserialization step. However, nothing is done with the annotations at this stage.

This is the first step in adding back the new OTIO schema as the preferred way of communication between RV and the Review App.

Live review using the new schema was tested between a local Review App where the feature was enabled with https://creative-collab.dev.shotguncloud.com and RV on MacOS.

### DIFF
--- a/src/plugins/rv-packages/otio_reader/otio_reader.py
+++ b/src/plugins/rv-packages/otio_reader/otio_reader.py
@@ -281,7 +281,7 @@ def _get_global_transform(tl):
     # that can contain all clips
     def find_display_bounds(tl):
         display_bounds = None
-        for clip in tl.clip_if():
+        for clip in tl.find_clips():
             try:
                 bounds = clip.media_reference.available_image_bounds
                 if bounds:
--- a/src/plugins/rv-packages/otio_reader/otio_reader.py
+++ b/src/plugins/rv-packages/otio_reader/otio_reader.py
@@ -59,6 +59,21 @@ class NoNodeFromHook(otio.exceptions.OTIOError):
     pass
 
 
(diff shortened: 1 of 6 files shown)


src/plugins/rv-packages/otio_reader/otio_reader.py
```diff
@@ -59,6 +59,24 @@ class NoNodeFromHook(otio.exceptions.OTIOError):
...
+    """
+    Main entry point to expand a given otio string into the current RV session.
+
+    Returns the top level node created that represents this otio
+    timeline.
+    """
+    otio_obj = otio.adapters.read_from_string(otio_string)
+    timeline = otio_obj["otio"]
+
+    if host_prefix is not None:
```
rogernelson, line 72: Simpler way of doing that:
```
context = {"sg_url": host_prefix} if host_prefix else None
```

src/plugins/rv-packages/otio_reader/otio_reader.py
```diff
@@ -276,12 +294,12 @@ def _create_track(in_seq, context=None):
     return new_seq
 
 
-def _get_global_transform(tl):
+def _get_global_transform(tl) -> dict:
```
rogernelson, line 297: Great, thanks for adding these.

rogernelson (approved):
Looks good, thanks!

bernie-laberge (approved):
LGTM
Thank you Éloïse!

PR #603, opened 2024-10-16
Fix build break on CentOS

eloisebrosseau:
### Summarize your change.

Set `pyzstd` and `py7zr` to the maximum version working with Python 3.8.

The version of `pyzstd` required by the latest version of `py7zr` doesn't work with Python 3.8 which is used with CentOS 7 which is causing a build break. This fix should be removed once the support for CentOS is dropped.

The build step was tested on CentOS 7.


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #608, opened 2024-10-24
Fix potential pyzstd requirement error

bernie-laberge:
Fix potential pyzstd requirement error

### Linked issues
NA

### Summarize your change.

It was reported that the previously pinned version of pyzstd==0.15.9 (which was required to build successfully on CentOS7), was not available on all python versions.
Now only installing this pinned version for Python version <= 3.10.

Also in this commit:
Fixed RV_VFX_SUPPORTED_OPTIONS when unsupported RV_VFX_PLATFORM was specified.
This was flagged by Will! Thank you Will !

Added pyenv version (.python_version) to gitignore.

It was reported that the previously pinned version of pyzstd==0.15.9 (which was required to build successfully on CentOS7), was not available on all python versions.

Successfully tested on macOS using two different versions of Python. 3.10.13 and 3.11.9


cmake/defaults/rv_options.cmake
```diff
@@ -49,29 +49,31 @@ SET_PROPERTY(
...
+SET(_RV_VFX_PLATFORM
+    "CY2023"
+)
 
 IF(DEFINED RV_VFX_PLATFORM)
   # Match lowercase and uppercase.
   STRING(TOUPPER "${RV_VFX_PLATFORM}" _RV_VFX_PLATFORM)
   IF(NOT "${_RV_VFX_PLATFORM}" IN_LIST RV_VFX_SUPPORTED_OPTIONS)
-    MESSAGE(FATAL_ERROR "RV_VFX_PLATFORM=${RV_VFX_PLATFORM} is unsupported. Supported value are: ${_RV_VFX_SUPPORTED_OPTIONS}")
+    MESSAGE(FATAL_ERROR "RV_VFX_PLATFORM=${RV_VFX_PLATFORM} is unsupported. Supported value are: ${RV_VFX_SUPPORTED_OPTIONS}")
```
eloisebrosseau, line 69: Typo: "Supported value*s* are"

requirements.txt
```diff
@@ -1,9 +1,10 @@
...
-pyzstd==0.15.9      # License: BSD License (BSD-3-Clause)
-py7zr==0.22.0       # License: GNU Lesser General Public License v2 or later (LGPLv2+) (LGPL-2.1-or-later)
-dohq-artifactory    # License: MIT License (MIT License)
-pre-commit          # License: MIT License (MIT)
+pyyaml                                   # License: MIT License (MIT)
+requests                                 # License: Apache Software License (Apache 2.0)
+wheel                                    # License: MIT License (MIT)
+pyzstd==0.15.9; python_version <= '3.10' # License: BSD License (BSD-3-Clause)
+pyzstd; python_version > '3.10'          # License: BSD License (BSD-3-Clause)
+py7zr==0.22.0                            # License: GNU Lesser General Public License v2 or later (LGPLv2+) (LGPL-2.1-or-later)
```
eloisebrosseau, line 8: I believe you could do the same Python version check for `py7zr` since it only needs to be set to 0.22.0 for Python 3.8. What do you think?

eloisebrosseau (approved):

requirements.txt
bernie-laberge: Excellent idea! I'll do it.

cedrik-fuoco-adsk (approved):

PR #612, opened 2024-10-29
Add annotation in OTIO reader for Live Review

eloisebrosseau:
### Summarize your change.

- [X] Add the paint and point schemas used to defined an annotation drawn by a presenter
- [X] Add an annotation hook to handle the display of annotations sent from an OTIO string
- [X] Handle the exception where a source is not active by manually setting the aspect ratio
- [X] Add a parameter to set the node to use in findAnnotatedFrames

A participant to a Live Review session is now able to see annotations drawn by the presenter.

The changes were tested on MacOS arm64.

### DIFF
--- a/src/plugins/rv-packages/otio_reader/PACKAGE
+++ b/src/plugins/rv-packages/otio_reader/PACKAGE
@@ -2,10 +2,10 @@ package: OTIO Reader
 author: Contributors to the OpenTimelineIO project
 organization: OpenTimelineIO project
 contact: opentimelineio@pixar.com
-version: 1.1
+version: 1.2
 url: http://opentimeline.io
-rv: 4.0.8
-openrv: 1.0.0
+rv: 2024.2.0
+openrv: 2.1.0
 
 modes: 
   - file: otio_reader_plugin.py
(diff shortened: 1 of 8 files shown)


src/plugins/rv-packages/otio_reader/PACKAGE
```diff
@@ -42,6 +44,10 @@ files:
     location: SupportFiles/$PACKAGE
   - file: multiRepPostExportHook.py
     location: SupportFiles/$PACKAGE
+  - file: paint_schema.py
+    location: SupportFiles/$PACKAGE
+  - file: point_schema.py
+    location: SupportFiles/$PACKAGE
```
bernie-laberge, line 50: This is not essential but I typically bump the version of the package when making changes to a package.
Like in this case, you could bump version: 1.1 to version: 1.2

bernie-laberge (approved):
This is great!
Thank you @eloisebrosseau

src/plugins/rv-packages/otio_reader/PACKAGE
eloisebrosseau: Done!

cedrik-fuoco-adsk (approved):

PR #613, opened 2024-11-07
Fix live review annotation opacity

eloisebrosseau:
### Summarize your change.

Add a new points to the list of points to draw only if it's not already there.

The number of points was causing the opacity level too be too opaque.

The opacity level was tested by changing it's value in the ReviewApp and seeing the resulting annotation in RV on MacOS.


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

Issue #366, opened 2024-01-16
[Maintenance]: VFX Reference Platform CY24


compaga:
- Python 3.11.x
- Qt 6.5.x
- Open EXR 3.2.x
- OCIO 2.3.x

[https://vfxplatform.com](https://vfxplatform.com) 
```[tasklist]
### Tasks
- [ ] https://github.com/AcademySoftwareFoundation/OpenRV/issues/367
- [ ] https://github.com/AcademySoftwareFoundation/OpenRV/issues/368
- [ ] https://github.com/AcademySoftwareFoundation/OpenRV/issues/369
- [ ] https://github.com/AcademySoftwareFoundation/OpenRV/issues/370
```

pbergeron-adsk:
@compaga @cedrik-fuoco-adsk After verification, the following:

- Python 3.11.x
- Qt 6.5.x
- Open EXR 3.2.x
- OCIO 2.3.x

Are indeed now supported in OpenRV main. Should probably close this issue unless you have something else to add.

cedrik-fuoco-adsk:
VFX2024 has been completed since the feature/qt6 merge into main. Closing the issue.

PR #618, opened 2024-11-08
Qt6 feature branch - Various fix for all platforms and VFX 2024 CI (Except Windows)

cedrik-fuoco-adsk:
### Linked issues
#366 

### Summarize your change.

Here's are the changes:

- Added VFX 2024 CI (except for Windows) without ci-openrv from aswf-docker for now (Linux only, not tested yet)
- Code changes to fix issues on Windows
- Code changes to fix issues on MacOS
- Added OpenSSL 3 for VFX2024
- Added some logic to all `.wrapper` to detect which VFX platform is active (using Python version) and then use the right OpenSSL version.
- FIx issue with Python not finding Tkinter on MacOS. The issue was with VFX2024, but the fix make sense for VFX2023 too.

**There are a lot of reformatting because of the pre-commit hook.
Windows is not included in the VFX 2024 CI because of issues with PySide6 in the CI.**

- [x] Rocky linux 8/9
- [x] MacOS
- [ ] Windows (partial, can compile with my windows, but not in CI because of pyside6 issue)


bernie-laberge (approved):

src/lib/app/RvCommon/RvApplication.cpp
```diff
@@ -471,6 +471,7 @@ static void
 callRunCreateSession(void* a)
 {
     RvApplication* app = reinterpret_cast<RvApplication*>(a);
+    std::cout << "??? Calling runCreateSession" << std::endl;
```
eloisebrosseau, line 474: I think you forgot to remove this

eloisebrosseau (approved):

src/lib/app/RvCommon/RvApplication.cpp
cedrik-fuoco-adsk: ah, I think it came back with my merge. I'll remove it!

PR #628, opened 2024-11-15
CI - Run job on all changes

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Remove the path-ignores in the CI.

It causes issues where the CI is not running but is still mandatory.


bernie-laberge (approved):
LGTM

PR #641, opened 2024-11-28
Qt6 Feature branch - Fix issue with menu on MacOS

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
The QMenuBar and QMenu has changed in Qt6. Some behaviours seems to have changed in Qt6, and the code around a memory leak (based on the comments) is not longer working.

I made some changes that works on MacOS and Rocky Linux 8. There no information about how to reproduce that memory leak.
**We'll need to test around that to make sure that it is not the case with Qt6 anymore.**

Multiple menu were appearing on the menu bar on MacOS.

- [x] MacOS
- [x] Rocky Linux 8


bernie-laberge (approved):
LGTM

PR #642, opened 2024-11-28
Fix default Python locale and now honor PYTHONUTF8 env var

bernie-laberge:
### Linked issues
NA

### Summarize your change.

Now pre-initializing the Python interpreter in RV to ensure that the PyConfig.use_environment is set to 1 (default).
Without this pre-initialization, PyConfig.use_environment defaults to 0 which means that the currently set locale gets ignored as well as the PYTHONUTF8 env var if set by the user.

Since commercial RV 2023 (Python 3.9.13), the default Python locale was no longer respecting the currently set environment variables (including the PYTHONUTF8 env var) and the script described below, was always returning:
US-ASCII
0

Prior to this release, such as commercial RV 2022.3.1 (Python 3.7.13), the default Python locale was respecting both the currently set environment variables, including PYTHONUTF8 env var. So we would generally have
UTF-8
1

I used this python script to validate the fix:

```
import locale
import sys
print(locale.getpreferredencoding(False))
print(sys.flags.utf8_mode)
```
It now returns:
UTF-8
1

and now honours the PYTHONUTF8 env var
export PYTHONUTF8=0 

It now returns:
US-ASCII
0

### DIFF
--- a/src/lib/app/PyTwkApp/PyInterface.cpp
+++ b/src/lib/app/PyTwkApp/PyInterface.cpp
@@ -255,6 +255,15 @@ namespace TwkApp
 
   void initPython( int argc, char** argv )
   {
+    // PreInitialize Python
+    // Note: This is necessary for Python to utilize environment variables like PYTHONUTF8
+    PyPreConfig preconfig;
+    PyPreConfig_InitPythonConfig(&preconfig);
+    PyStatus status=Py_PreInitialize(&preconfig);
+    if (PyStatus_Exception(status)) {
+      Py_ExitStatusException(status);
+    }
+
     Py_InitializeEx( 1 );
     static wchar_t delim = L'\0';
 


src/lib/app/PyTwkApp/PyInterface.cpp
```diff
@@ -255,6 +255,12 @@ namespace TwkApp
 
   void initPython( int argc, char** argv )
   {
+    // PreInitialize Python
+    // Note: This is necessary for Python to utilize environment variables like PYTHONUTF8
+    PyPreConfig preconfig;
+    PyPreConfig_InitPythonConfig(&preconfig);
+    Py_PreInitialize(&preconfig);
```
eloisebrosseau, line 262: Minor: Since `Py_PreInitialize` has a return value, I think we should use it to validate the call like in the example from [PEP0587](https://peps.python.org/pep-0587/#:~:text=to%20NULL.-,Preinitialization%20with%20PyPreConfig,-The%20PyPreConfig%20structure) before initializing the interpreter:
`status = Py_PreInitialize(&preconfig);
if (PyStatus_Exception(status)) {
    Py_ExitStatusException(status);
}`

eloisebrosseau (approved):

src/lib/app/PyTwkApp/PyInterface.cpp
bernie-laberge: Good catch ! Thanks. I'm on it

rogernelson (approved):

PR #645, opened 2024-12-02
Update ClangFormat

eloisebrosseau:
### Summarize your change.

The .clang-format was updated to the latest version and added to the .pre-commit-config.yaml. I also modernized the formatting style based on LLVM. Every options related to the include statements formatting was removed since it might have some side effects on some legacy parts of the codebase if they were to be reorder. Once this PR is merged, the same formatting should be applied to the commercial RV repository for consistency. Moreover, I suggest to make another PR to format all the files afterwards and to add a `.git-blame-ignore-revs` to avoid having the formatting commit show up when using `git blame`. Ideally, once this is done, we should add a formatting validation in our CI pipeline to avoid relying only on users installing the pre-commit hook to follow the formatting convention.

The .clang-format had a lot of deprecated options and not all the files were formatted the same way.

Here's an example of what a formatted file would look like with the new .clang-format. Please let me know if you don't agree with any of the chosen options.

Screenshot 2024-12-02 at 11 47 35 AM
Screenshot 2024-12-02 at 11 47 59 AM
Screenshot 2024-12-02 at 11 48 22 AM
Screenshot 2024-12-02 at 11 48 33 AM

### DIFF
--- a/.clang-format
+++ b/.clang-format
@@ -1,94 +1,153 @@
 ---
 Language:        Cpp
-AccessModifierOffset: -1
+AccessModifierOffset: -4
 AlignAfterOpenBracket: Align
-AlignConsecutiveAssignments: false
-AlignConsecutiveDeclarations: false
-AlignEscapedNewlinesLeft: true
-AlignOperands:   true
-AlignTrailingComments: true
+AlignArrayOfStructures: None
+AlignConsecutiveAssignments: None
+AlignConsecutiveBitFields: None
+AlignConsecutiveDeclarations: None
+AlignConsecutiveMacros: None
+AlignConsecutiveShortCaseStatements:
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):
LGTM

Note that when we first made a public release of Open RV, we had a clang-format that turned out introducing an issue that we didn't have time to investigate at the time. Which is why we decided not to include clang-format as part of the pre-commit at the time. To be honest, I don't remember what was the symptom but I believe it was pretty catastrophic (RV wouldn't launch) so it will be easy to detect. 

Thank you @eloisebrosseau for looking into this!

cedrik-fuoco-adsk (approved):

rogernelson (approved):

PR #646, opened 2024-12-02
Clean up the OTIO reader annotation

eloisebrosseau:
### Clean up the OTIO reader annotation

This PR aims to clean up a bit the code related to the OTIO reader for annotations in Live Review.

### Question
I see that some files for other hooks and schemas are in camel case instead of snake case. Does anyone knows if there is a reason? Is everyone ok with me renaming the files inside `otio_reader` to use snake case instead?

### DIFF
--- a/src/plugins/rv-packages/otio_reader/annotation_hook.py
+++ b/src/plugins/rv-packages/otio_reader/annotation_hook.py
@@ -48,8 +48,8 @@ def hook_function(
                     "brush": layer.brush,
                     "debug": 1,
                     "join": 3,
-                    "cap": 2,
-                    "splat": 1,
+                    "cap": 1,
+                    "splat": 0,
                     "mode": 0 if layer.type == "COLOR" else 1,
                 },
             )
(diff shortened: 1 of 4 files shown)


src/plugins/rv-packages/otio_reader/annotation_hook.py
```diff
@@ -11,18 +11,21 @@
...
             else:
-                range = otio.opentime.TimeRange(
+                time_range = otio.opentime.TimeRange(
                     layer.layer_range["start_time"], layer.layer_range["duration"]
                 )
 
-            relative_time = range.end_time_inclusive()
-            frame = relative_time.to_frames()
+            relative_time = time_range.end_time_inclusive()
+            frame = end_time.to_frames()
```
bernie-laberge, line 28: end_time  ?
Shouldn't it be **relative_time** here instead ?
frame = relative_time.to_frames()

eloisebrosseau: You're right. Thank you!

bernie-laberge (approved):
LGTM

Thank you @eloisebrosseau !

src/plugins/rv-packages/otio_reader/annotation_hook.py
```diff
@@ -70,13 +73,11 @@ def hook_function(in_timeline, argument_map=None) -> None:
 
             for point in layer.points:
                 points = commands.getFloatProperty(points_property)
-                if (
+                if not (
                     len(points) > 1
                     and points[-1] == point.y * global_scale.y
```
rogernelson, line 78: wouldn't you need to change the `and` to `or` to maintain the same logic? Or was the old logic wrong?

PR #647, opened 2024-12-02
Add default global scale value if missing in the annotation hook

eloisebrosseau:
### Summarize your change.

A default global scale value is calculated based on the aspect ratio of the first source if it couldn't be obtained from the OTIO.

When no `available_image_bounds` is present in the OTIO, no global scale value was being calculated. This was causing an issue where the annotations couldn't be displayed in RV during a Live Review Session.

Successfully test on MacOS

### DIFF
--- a/src/plugins/rv-packages/otio_reader/annotation_hook.py
+++ b/src/plugins/rv-packages/otio_reader/annotation_hook.py
@@ -6,6 +6,7 @@
 # *****************************************************************************
 
 
+import logging
 import effectHook
 import opentimelineio as otio
 from rv import commands, extra_commands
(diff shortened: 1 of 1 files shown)


src/plugins/rv-packages/otio_reader/annotation_hook.py
```diff
@@ -58,6 +58,14 @@ def hook_function(in_timeline, argument_map=None) -> None:
             )
 
             global_scale = argument_map.get("global_scale")
+            if global_scale is None:
+                first_source_node = commands.sourcesAtFrame(0)[0]
+                media_info = commands.sourceMediaInfo(first_source_node)
+                height = media_info["height"]
+                aspect_ratio = media_info["width"] / height
```
bernie-laberge, line 65: Perhaps we should protect against division by zero error here.
I see a similar logic here:
https://github.com/AcademySoftwareFoundation/OpenRV/blob/ee1ecd528431d4a2f446406fc11c890ac57fbce3/src/plugins/rv-packages/otio_reader/otio_reader.py#L662

eloisebrosseau: I agree! Done!

src/plugins/rv-packages/otio_reader/annotation_hook.py
```diff
@@ -58,6 +59,21 @@ def hook_function(in_timeline, argument_map=None) -> None:
             )
 
             global_scale = argument_map.get("global_scale")
+            if global_scale is None:
+                try:
+                    first_source_node = commands.sourcesAtFrame(0)[0]
+                    media_info = commands.sourceMediaInfo(first_source_node)
+                    height = media_info["height"]
+                    aspect_ratio = media_info["width"] / height
```
rogernelson, line 67: Maybe worth logging something here too that we couldn't find a global scale, so we use the aspect ratio of the first media file?

eloisebrosseau: I agree!

Issue #367, opened 2024-01-16
[Maintenance]: Update to Qt6.5.x


compaga:

pbergeron-adsk:
@cedrik-fuoco-adsk See Richardssam commits from 2 weeks ago --- was this merged in? If so can this issue be closed?

cedrik-fuoco-adsk:
I am not sure why this name appear there since the commits are from me. This is resolved since the feature/qt6 merge into the main branch.

PR #651, opened 2024-12-06
Qt6 Feature - Fix issues with Python 3.11 and PySide6 compilation on Windows

cedrik-fuoco-adsk:
### Linked issues
#367 

I created two functions to do the install step of Python. One for Python 3.10 (VFX2023) and one for Python 3.11 (VFX2024). I think it will be easier to maintain and remove later on. The difference was only for Windows, but both functions contains the code for all platforms.

Since Python 3.11, the `PLATLIBDIR` directory was renamed to `DLLs` instead of `lib`. You can changed it by modifying the Visual studio project file, but I found some place in the Python source code that was hardcoded to `DLLs`. I opted to keep `DLLs` and copy that over to the OpenRV's stage directory. It does not change anything functionally because the Python executable is looking for the DLLs library automatically.

**The changes above also fixes the issue with PySide6 compilation on Windows.**

Windows

stage directory:
image


src/build/make_python.py
```diff
@@ -651,7 +651,75 @@ def build() -> None:
         ).check_returncode()
 
 
-def install() -> None:
+def install_python_vfx2023() -> None:
```
cedrik-fuoco-adsk, line 654: This is the same code as the install() method on the main branch.

eloisebrosseau (approved):

PR #653, opened 2024-12-11
Add Python initialization in rvls

eloisebrosseau:
### Summarize your change.

Initialize Python in rvls main function to match the behaviour of rvio inside the ShotGrid license validation.

The license validation was failing on Linux when using rvls, but the same code was working fine with rvio because of the Python initialization in main(). The Python interpreter from rvio had more capabilities that were needed to properly import Python modules on Linux.

This change was tested on Rocky8, MacOS Sonoma and Windows 11.


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #656, opened 2024-12-17
Format Open RV codebase

eloisebrosseau:
### Summarize your change.

This PR formats the Open RV codebase with the new .clang-format that was recently added. Some small tweaks were made to the .clang-format to make sure no options could modify the code in a way that would break it. Note that the pub and WFObj submodules are also being formatted here: https://github.com/shotgunsoftware/openrv-pub/pull/20 and https://github.com/shotgunsoftware/openrv-WFObj/pull/2

The code needed to be formatted with the new .clang-format

Open RV is building and working with any issues on MacOS Sonoma.


bernie-laberge (approved):
LGTM

PR #660, opened 2025-01-08
Feature/Qt6 - Fixing the audio settings

cedrik-fuoco-adsk:
### Linked issues
n/a

The code that was handling the audio sample rate changes was refactored in this PR because of the Qt Audio module refactor in Qt 6. In Qt 5, you could get a list of valid sample rate from the audio device, but in Qt 6, you only get a minimum and maximum value.

With Qt6 OpenRV, the widget to change the sample rate is a SpinBox instead of a ComboBox.

**_What is left to do for Audio:_**

- [ ] There are still some memory issues (crash) when changing audio sample rate on MacOS (tested on ARM64). **This PR does not fix that.**
- [ ] Other refactor to support audio format (e.g. int8) that are not available by default in Qt 6 anymore.

Windows, MacOS and Rocky.

image


src/lib/audio/QTAudioRenderer/QTAudioRenderer.cpp
cedrik-fuoco-adsk, line 1: Even though I split the Qt5 in QT5AudioRender.cpp, I keep the `#if defined( RV_VFX_CY2023 )` as reference for now. I was thinking of removing them for this file once it is working great on all platforms.

PR #668, opened 2025-01-27
Add Support for the Cannon Raw 3 File Type

chrisc-lee:
Add Support for the Cannon Raw 3 File Type

### Summarize your change.
Added support for OpenRV to read cr3 files.

This allows OpenRV to read cr3 files which is becoming more common as it is intended to replace cr2.

I have tested this on both Windows 11 and AlmaLinux 9 by loading a cr3 file in OpenRV.

Since this is something supported by the version of libraw that OpenRV is currently using, it is as simple as just adding another type to the IORaw.cpp file.

Screenshot 2025-01-27 002058


bernie-laberge (approved):
Fantastic !
Thank you for your contribution !

eloisebrosseau (approved):

PR #669, opened 2025-01-27
Add .git-blame-ignore-revs

eloisebrosseau:
### Summarize your change.

Add a .git-blame-ignore-revs file and update README.md to explain how to use it.

A .git-blame-ignore-revs file lists commits to ignore when running `git blame`, such as formatting commits. This allows you to use `git blame` without these commits cluttering the Git history. For now, this file only contains the formatting of all the C++ codebase in Open RV based on the new .clang-format. Other refactoring changes that are not modifying the code logic can be added to this file as needed.

### DIFF
--- a/README.md
+++ b/README.md
@@ -30,6 +30,14 @@ following command:
 git submodule update --init --recursive
 ```
 
+## Configuring Git to use the ignore file with `git blame`
+
+A `.git-blame-ignore-revs` file lists commits to ignore when running `git blame`, such as formatting commits. This allows you to use `git blame` without these commits cluttering the Git history. To configure Git to use this file when running `git blame`, use the following command:
+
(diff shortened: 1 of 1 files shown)


.git-blame-ignore-revs
```diff
@@ -0,0 +1,2 @@
+# Format all C++ files with clang-format
+9d6b1be0e431a570d1475ceb4293106669b7c20c
```
bernie-laberge, line 2: Out of curiosity: how did you end up with this value ? :-)

eloisebrosseau: This is the commit hash for the formatting PR I did here! #656

cedrik-fuoco-adsk (approved):

PR #670, opened 2025-01-27
Optimisations to OCIOIPNode

kenmcgaugh:
Refactored to move all the OCIO shader building code out of the evaluate() method and into its own method that is called during propertyChanged() instead.

Removed all mutex handling as the code being protected now only ever executes in the main thread (I think).

Changed behavior where an empty "ocio_color.outColorSpace" property will default to scene_linear. This now
matches the existing behavior of "ocio_look.outColorSpace".

Changed behavior where an empty "ocio_display.display" property uses the config's default display.

Changed behavior where an empty "ocio_display.view" property uses the config's default view.

The above changes to these properties prevent error messages when OCIOIPNode's are instantiated since the shader building now executes before there is a chance for the properties to be set.

Added check for gpuProcessor->isNoOp() so that the node becomes a pass-through.

When the Shader::Function is constructed, passed numTextures+num3Dtextures for the fetches argument rather than 1. Felt like the right thing to do but I may be completely wrong.

In my tests on a MacBook Pro (Apple M1 Pro), I'm getting almost 2x speed up in playback performance.


bernie-laberge:
Hello @kenmcgaugh,

I just want to let you know that we are very excited about this PR of yours !
We apologize for not getting back to you earlier.
We'll do our best to review it this week.

Thank you for contributing to the Open RV cause !

jorxster:
Nice work Ken!

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -352,83 +344,21 @@ namespace IPCore
...
+            m_matrix_rec709_to_xyz = createMatrixTransformXYZToRec709();
+            m_matrix_rec709_to_xyz->setDirection(
+                OCIO::TRANSFORM_DIR_INVERSE);
         }
 
         return m_matrix_rec709_to_xyz;
     }
 
-    IPImage* OCIOIPNode::evaluate(const Context& context)
+    void OCIOIPNode::updateFunction()
```
bernie-laberge, line 355: Excellent idea !

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -644,7 +565,15 @@ namespace IPCore
             gpuProcessor->extractGpuShaderInfo(shaderDesc);
 
             string shaderCacheID = gpuProcessor->getCacheID();
-            if (m_state->shaderID != shaderCacheID)
+
+            if (gpuProcessor->isNoOp())
```
bernie-laberge, line 569: nice !

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -680,7 +609,7 @@ namespace IPCore
 
                 m_state->function = new Shader::Function(
                     shaderDesc->getFunctionName(), glsl,
-                    Shader::Function::Color, 1 /*numFetchesApprox*/);
+                    Shader::Function::Color, numTextures+num3DTextures);
```
bernie-laberge, line 612: Good catch !

bernie-laberge (approved):
LGTM
Great idea to separate the OCIO Gpu processor generation business from the OCIOIpNode::evaluate() even though I believe it was supposed to be cached at the OCIO level. Something we had never revisited since this node was introduced 12 years ago (I checked). 
I have tested it and it really makes a difference.
Thanks you @kenmcgaugh !

And I apologize for the time it took for the review: it is crunch time here for us.

kenmcgaugh:
Thanks for your feedback. Interestingly enough, just yesterday I discovered a couple of small issues when reading .rv session files. I’ll commit my update fixing them. One is to add a call to updateFunction() in the readFinished() method. The other is to return early from updateFunction()  if m_state->linear.empty().

eloisebrosseau (approved):

PR #671, opened 2025-01-27
Set default value for ffmpeg threads to 1 instead of the implicit 0.

kenmcgaugh:
When set to 0 RV's thread count will explode since it will create (num_cpus+1)*(num_sources)*(num_reader_threads) threads. In my test of loading 25 prores clips RV's thread count grew to 726 and consumed 10GB of memory. With this change in place, the thread count remained stable at 49 and memory used was 7.5GB.

To restore previous behaviour you can set the env var RV_MOVIEFFMPEG_ARGS="--codecThreads 0".


pbergeron-adsk:
Discusing with Bernard, it appears that we lose realtimeness with this PR. Will need to investigate more.

kenmcgaugh:
Perhaps there could be some heuristic in place based off of the number of reader threads at least. A conservative heuristic would be to set the codec threads to the number of CPU’s minus the number of reader threads.

Would the ultimate solution to this be to make MovieFFMpeg multithreadable?

PR #684, opened 2025-02-11
Feature/Qt6 - Add QTimezone to the MuQt5 bindings and ignore templates for clang-format

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
This PR adds QTimezone to the MuQt5 bindings and added .clang-format-ignore to ignore the MuQt templates.

This is needed to fix some Mu plugins.


bernie-laberge (approved):
LGTM

PR #686, opened 2025-02-12
Prevent 'ERROR: Unable…' when loading OTIO with MMR

bernie-laberge:
### Linked issues
NA

To fix the following error when loading OTIO with multiple media representations
'ERROR: Unable to determine aspect ratio…' 

### Summarize your change.

When multiple media representation sources are created, we can only retrieve the media_info of a source if it is the active source. That's because multiple media representation in Open RV is optimized to only load a source when it gets activated.
If the current source is not the active source, we can't get the media_info so we fall back to the active source's media_info as our best guess.

Successfully tested on macOS


rogernelson (approved):
Looks good to me, nice work Bernie.

PR #691, opened 2025-02-14
Feature/Qt6 - Fix issue where Mu Command API Browser is not opening

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
There were some changes in the Qt 6 QVariant type and it seems like we need to explicitly define a QVariant constructor for QIcon in Qt 6.

Mu Command API Browser is not opening in the build using Qt 6.

Rocky 8


bernie-laberge (approved):
LGTM

eloisebrosseau (approved):

PR #700, opened 2025-02-24
Feature/Qt6 - Fix dropperSampler not working correctly

pbergeron-adsk:
### Summarize your change.

The dropperSampler tool in RV was not working correctly anymore in the qt6 branch. It had a strange offset (horizontally and vertically) because the OpenGL function glReadPixels was reading from the wrong context and the wrong framebuffer object.

in annotate_mode.mu, the method dropperSampler() called an OpenGL method (glReadPixels) while not in a rendering pass. This somehow worked in the old Qt5/GLWidget implementation, because the current OpenGL context and FBO always remained set to the main view widget.  But since switching to Qt6/QOpenGLWidget, this no longer holds true - the Qt6/QOpenGLWidget backend has a OpenGL compositor that works differently, and the OpenGL context/fbo of the main view was no longer current after the application was done refreshing.

Since we want to query the final rendered view using glReadPixels, we implemented a mu method called framebufferPixelValue(x, y) which internally eventually sets the OpenGL context/fbo to the correct one, calls glReadPixels(), and then restores the old one. 

See first paragraph of the "Summarize your change" section

I have tested the dropperSampler tool, with different layouts and docked windows, but on Linux64 only. Please test in Darwin and Windows.

n/a

n/a

### DIFF
--- a/src/lib/app/RvCommon/GLView.cpp
+++ b/src/lib/app/RvCommon/GLView.cpp
@@ -330,10 +330,92 @@ namespace Rv
 #endif
     }
 
+    bool GLView::validateReadPixels(int x, int y, int w, int h)
+    {
+        int r = x + w;
+        int t = y + h;
+
+        // are the extents of the read region out of bounds?
+        if (x < 0 || y < 0 || r > width() || t > height())
+            return false;
+
+        return true;
+    }
+
+    QImage GLView::readPixels(int x, int y, int w, int h)
+    {
+
+        // If out of bounds, return an empty image.
+        if (validateReadPixels(x, y, w, h) == false)
+        {
+            QImage image(0, 0, QImage::Format_RGBA8888);
+            return image;
+        }
+
+        //
+        // Why this method exists, and why saving and restoring the context is
+        // important:
+        //
+        // With the old QGLWidget implement with Qt 5.15.2, the current context
+        // stayed bounded to the main view (eg: the instance of GLView) after
+        // each paint of the GLView. This made it possible This made it possible
+        // to call gl functions (eg: glReadPixels) even when not in the context
+        // of a render call For example, in Mu:
+        //
+        // method: render (void; Event event)
+        //
+        // Now with the new QOpenGL paradigm, after the application is done
+        // rendering, the current OpenGL context is no longer bound to the
+        // context of the GLView, and is instead bound to the context of the
+        // main Qt frame window.
+        //
+        // This caused a problem with some tools, like the dropperSample() tool
+        // of the mu annotation toolset, which blindly called glReadPixels
+        // whenever the mouse was clicked/dragged over the view. Since the
+        // context was no longer bound to the view, but to the qapplication's
+        // main window, glReadPixels would return invalid colors.
+        //
+        // Therefore we implemented this readPixels on the glView, and exposed a
+        // "framebufferPixelValue(x,y)" method which ends up here, allowing us
+        // to perform some save/retore operations of the context before
+        // returning.
+        //
+        // Note that we can't simply make the GLiew's context current without
+        // restoring the old context, since this creates side effects and causes
+        // the glDebug() to complain about invalid states elsewhere.
+        //
+
+        // Save the current context
+        QOpenGLContext* prevContext = QOpenGLContext::currentContext();
+        QSurface* prevSurface = prevContext->surface();
+
+        // Make the current context the one of the GL view
+        makeCurrent();
+
+        QImage image(w, h, QImage::Format_RGBA8888);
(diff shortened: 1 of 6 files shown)


src/lib/app/RvCommon/RvCommon/MuUICommands.h
```diff
@@ -89,6 +89,7 @@ namespace Rv
     NODE_DECLARATION(launchTLI, void);
     NODE_DECLARATION(rvioSetup, void);
     NODE_DECLARATION(javascriptMuExport, void);
+    NODE_DECLARATION(framebufferPixelValue, Mu::Vector4f);
```
bernie-laberge, line 92: Please add this new command to the following file so that this new command is also available in python:
OpenRV/src/lib/app/py_rvui/rv_commands_setup.py

src/lib/app/RvCommon/GLView.cpp
```diff
@@ -330,10 +330,92 @@ namespace Rv
...
+
+        // image.save("/home/<username>>/<orv_folder>/fbo.png");
+    }
+
     void GLView::swapBuffersNoSync()
     {
         glFlush();
-        glBindFramebuffer(GL_FRAMEBUFFER, 0);
+        glBindFramebuffer(GL_FRAMEBUFFER,
+                          context()->defaultFramebufferObject());
```
bernie-laberge, line 418: Good catch ! Thanks!

bernie-laberge (approved):
Excellent.
Thank you @pbergeron-adsk !

src/lib/app/RvCommon/GLView.cpp
```diff
@@ -330,10 +330,92 @@ namespace Rv
...
+            return image;
+        }
+
+        //
+        // Why this method exists, and why saving and restoring the context is
+        // important:
+        //
+        // With the old QGLWidget implement with Qt 5.15.2, the current context
+        // stayed bounded to the main view (eg: the instance of GLView) after
+        // each paint of the GLView. This made it possible This made it possible
```
eloisebrosseau, line 361: Typo: "This made it possible" is written twice

src/lib/app/RvCommon/GLView.cpp
```diff
@@ -330,10 +330,92 @@ namespace Rv
...
+        //
+        // This caused a problem with some tools, like the dropperSample() tool
+        // of the mu annotation toolset, which blindly called glReadPixels
+        // whenever the mouse was clicked/dragged over the view. Since the
+        // context was no longer bound to the view, but to the qapplication's
+        // main window, glReadPixels would return invalid colors.
+        //
+        // Therefore we implemented this readPixels on the glView, and exposed a
+        // "framebufferPixelValue(x,y)" method which ends up here, allowing us
+        // to perform some save/retore operations of the context before
```
eloisebrosseau, line 380: Typo: re**s**tore

src/lib/app/RvCommon/GLView.cpp
```diff
@@ -330,10 +330,92 @@ namespace Rv
...
+    }
+
+    void GLView::debugSaveFramebuffer()
+    {
+        // Create a QImage with the same size as the FBO
+        QImage image(width(), height(), QImage::Format_RGBA8888);
+        glReadPixels(0, 0, width(), height(), GL_RGBA, GL_UNSIGNED_BYTE,
+                     image.bits());
+
+        // image.save("/home/<username>>/<orv_folder>/fbo.png");
```
eloisebrosseau, line 411: Is this commented out line for debugging purposes?

eloisebrosseau (approved):

src/lib/app/RvCommon/GLView.cpp
bernie-laberge: I believe it is for debugging purposes and it could be pretty useful to help investigate future issues.

src/lib/app/RvCommon/GLView.cpp
pbergeron-adsk: fixed

src/lib/app/RvCommon/GLView.cpp
pbergeron-adsk: correct, this is for debugging purposes, to grab the contents of the opengl window.

src/lib/app/RvCommon/GLView.cpp
pbergeron-adsk: Fixed

pbergeron-adsk:
Before merging, please merge PR 707, since there's a change in here that will only work if that PR is merged.

cedrik-fuoco-adsk (approved):

PR #702, opened 2025-02-24
Qt6: Set correct locale before Qt gets a chance to complain about it …

pbergeron-adsk:
### Summarize your change.

This fix is to prevent QCoreApplication::initLocale() method to complain that an invalid locale is currently set. Before this bugfix, RV's init code explicitly set the current locale to "C". At the same time, this locale is now explicitly unsupported by Qt6, and instead will instead switch to "C.UTF-8" under Linux, or "UTF-8" under Darwin (macOS)

For more information about how Qt initializes the locale, look at initLocale() in the file:

QT_HOME/Qt6.x.x/Src/qtbase/src/corelib/kernel/qcoreapplication.cpp

See above.

Tested on Rocky 9.5, waiting on Cedrik to test on Darwin.

n/a

n/a

### DIFF
--- a/src/bin/apps/rv/main.cpp
+++ b/src/bin/apps/rv/main.cpp
@@ -220,6 +220,9 @@ static void setPlatformSpecificLocale()
     //    ..."no need to try language or region specific CTYPEs, as they
     //        all point back to the same generic UTF-8 CTYPE."
     //
+    // NOTE: Leaving this here for completeness for the curious minded,
+    // but the main.cpp file for macOS is actually src/bin/nsapps/RV/main.cpp
+
     locale = "UTF-8";
 #elif defined(PLATFORM_LINUX)
     // this is valid for Rocky Linux. As per the Qt code, some other Unixes may
--- a/src/bin/apps/rv/main.cpp
+++ b/src/bin/apps/rv/main.cpp
@@ -194,6 +194,51 @@ TwkApp::QTBundle bundle(EXECUTABLE_SHORT_NAME, MAJOR_VERSION, MINOR_VERSION,
 extern "C" int XInitThreads();
 #endif
 
+//
+// This function is an attempt to prevent QCoreApplication::initLocale() method
+// to complain that an invalid locale is currently set.
+//
+// Before this bugfix, RV's init code explicitely set the current locale to "C".
+// At the same time, this locale is now explicitely unsupported by Qt6, and
+// instead will instead switch to "C.UTF-8" under Linux, or "UTF-8" under Darwin
+// (macOS)
+//
+// For more information about how Qt initializes the locale, look at
+// initLocale() in the file:
+//
+// QT_HOME/Qt6.x.x/Src/qtbase/src/corelib/kernel/qcoreapplication.cpp
+//
+static void setPlatformSpecificLocale()
+{
+    std::string locale = "";
+
+#if defined(PLATFORM_DARWIN)
+    // In Darwin, "C.UTF-8" isn't directly supported, but we can use any
+    // "<region>.UTF-8" locale.  However, as per Qt's initLocale() method
+    // for in macOS, there is...
+    //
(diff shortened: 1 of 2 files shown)


src/bin/apps/rv/main.cpp
```diff
@@ -194,6 +194,51 @@ TwkApp::QTBundle bundle(EXECUTABLE_SHORT_NAME, MAJOR_VERSION, MINOR_VERSION,
...
+// initLocale() in the file:
+//
+// QT_HOME/Qt6.x.x/Src/qtbase/src/corelib/kernel/qcoreapplication.cpp
+//
+static void setPlatformSpecificLocale()
+{
+    std::string locale = "";
+
+#if defined(PLATFORM_DARWIN)
+    // In Darwin, "C.UTF-8" isn't directly supported, but we can use any
```
bernie-laberge, line 216: Note that RV on macOS has a different main.cpp file:
OpenRV/src/bin/nsapps/RV/main.cpp

src/bin/apps/rv/main.cpp
```diff
@@ -194,6 +194,51 @@ TwkApp::QTBundle bundle(EXECUTABLE_SHORT_NAME, MAJOR_VERSION, MINOR_VERSION,
...
+    // is no need.
+    locale = "C.UTF-8";
+#elif defined(WIN32)
+    // Under Windows, it seems Qt does not try to enfore switching to "C.UTF-8",
+    // (as this function only seems to be compiled for unix-variants).
+    // Therefore, we leave it as the "C" locale.
+    locale = "C";
+#endif
+
+    setEnvVar("LANG", locale);
```
bernie-laberge, line 236: Note that the reason we are setting the locale to C is to make sure that the regex used by RV works for image sequences.
Something like this would not work without setting LANG to C (like if it were set to French for example)
./rv [ /tmp/Small-DPX/img.#.dpx /tmp/Small-DPX/audio.wav ]

Please validate that with your fix, you can still load an image sequence from the RV command line by following the repro steps of SG-19003.
Thanks!

src/bin/apps/rv/main.cpp
pbergeron-adsk: Fixed - added the main.cpp file for macos. I suppose I should remove the references to (PLATFORM_DARWIN) here?

bernie-laberge: Yes please. Otherwise it would be misleading. Thanks.

src/bin/apps/rv/main.cpp
pbergeron-adsk: Ok, let me dig a bit.

pbergeron-adsk: @bernie-laberge  It turns out that Qt does nothing with the "LANG" locale, but it does override "LC_CTYPE". 
From what I gather of your comment above, the proper fix would be: 

setEnvVar("LANG", "C");
setEnvVar("LC_CTYPE", "C.UTF-8")   ?

pbergeron-adsk: @bernie-laberge After some more investigation (but I need to test), it turns out that using "C.UTF-8" vs "C" is very likely to work. What "C.UTF-8" does is to continue to maintain compatibility with the "C" locale, but also expands support for extended UTF-8 characters.  Even for regex evaluation, it should work. 

The difference with C.UTF-8 would be like so: 

**// In the "C" locale**
printf("%d\n", regex_match("Hello", "H.*o")); // Will match and print '1'
printf("%d\n", regex_match("你好", "你.*好")); // Won't match in "C" locale

**// In the "C.UTF-8" locale**
printf("%d\n", regex_match("Hello", "H.*o")); // Will still match and print '1'
printf("%d\n", regex_match("你好", "你.*好")); // Will match and print '1' in "C.UTF-8" locale
 
Nevertheless, I will perform some tests as per your recommendations.

bernie-laberge (approved):
LGTM

PR #704, opened 2025-02-26
Feature/Qt6 - fix crash when clicking on rvio Export widget while it's exporting quicktime video  (#35176)

pbergeron-adsk:

When we clicked or dragged the widget during a quicktime export (or any instance of the ProcessInfo widget for that matter), we (almost) inevitably always crashed. Happened in Qt5 and Qt6 branch, and as far back as RV 2023, apparently. 

### Summarize your change.

When we clicked on the ProcessInfo widget, the code tries to access the _buttons[] array to see if we are clicking in the (X) button, but simply reading this array crashed RV. 

It turns out that for the ProcessInfo widget, the render() method recreated the _buttons[] array, which contained the coordinates and process info for the (X) cancel button. This was in a different thread than the event-handling thread. So, while the render method was executing and the _buttons[] array was being recreated, the event-handling thread was trying to read the same array to see if w were clicking in the (X) button. Since the memory had been disposed and replaced in the meantime, this caused a crash .

The fix was simply to initialize the array when the ProcessInfo widget is created, and in the render() method we update the x/y/w/h/pid. This does not cause a change in memory space, and the crash no  longer happens. Ideally, we'd need a mutex to sync the read/write tof the x/y/w/h/pid values, but, we're really just reading/writing coordinates here, and the pid always remains constant, so even if there was concurrent read/write access it should be completely harmless.  

Either way, the crash no longer occurs.

### Describe what you have tested and on which operating system
.
This was 100% reproducible on Linux, and I think also macOS, the problem was no platform-specific but Mu-specifc.

There is only 1 file to review, so, no big deal.

n/a

### DIFF
--- a/src/lib/app/mu_rvui/HUD.mu
+++ b/src/lib/app/mu_rvui/HUD.mu
@@ -487,6 +487,9 @@ class: ProcessInfo : Widget
         //
         setEventTableBBox(_modeName, "global", vec2f(0,0), vec2f(0,0));
 
+        // create memory space for the (X) button, which we update in render()
+        _buttons = Button[] {{0,0,0,0, nil}};
+
         _x = 40;
         _y = 60;
     }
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
Great Catch !

rogernelson (approved):

PR #705, opened 2025-02-27
Handle hide annotations

eloisebrosseau:
### Summarize your change.

The paint schema was updated to reflect the latest changes done to the Live Review spec, and the annotation hook is now using the `visible` attribute to set the visibility of the annotations based on the OTIO timeline received by the ReviewApp.

During a Live Review session, a presenter can click on a `hide annotations` button to hide all the annotations on all the frames. The PR allows RV to correctly display the annotations as a Live Review participant based on the presenter's choice.

The changes were successfully tested on MacOS.


bernie-laberge (approved):
LGTM

rogernelson (approved):

PR #707, opened 2025-03-04
Feature/Qt6 - Fix error popup occurs when opening presentation device (eg: NDI)

pbergeron-adsk:
Fixes #38133

Could not open presentation device after setting Presenation device to NDIDevice, and then doing CTRL-P to enable it.  The error message was due to a ";" at the end of an if-statement, but then it caused opengl errors down the line due to the gl context being invalid after calling swapBuffers.

Turns out all the swapBuffer management is no longer needed with QOpenGLWidget, explanation in the summary below.

### Summarize your change.

Note for Qt6 . QOpenGLWidget implementation:

With the new QOpenGLWidget (Qt6 branch), all of the rendering of each widget is done in its own FBOs (aka: off-screen buffers), as opposed to the old Qt5/QGLWidget branch where the rendering was done in each GGLWidget's backbuffer (aka: GL_BACK, an on-screen framebuffer surface).

With QOpenGLWidget, it is now the WainWindow's responsibility (more technically, the MainWindow's rendering backend, which happens to be Qt's OpenGL rendering backend when QOpenGLWidget are present in the children tree) to gather and composite all of the  off-screen buffers (regardless of which image type they are, eg: cpu-memory images, gpu/opengl images, etc) and finally to call swapBuffers to show the final contents of the mainWindow. 

As a result, I'm not sure there's a point -- at all -- in calling swapBuffers on the GLView anywhere amnymore. From old comments in the previous version of this tile, it appears that calling swapBuffers was done to force a quicker visual update, or to minimize visual tearing of some sort. This would have worked with the old QGLWidget (becayuse each QGLWidget had its own context, and its rendering target was directly the GL_BACK 
framebuffer, but, again, with QOpenGLWidget, the rendering target is no longer GL_BACK, it is an FBO that is meant to be used at the end of the application's drawing / visual update pipeline.

I am not sure how this would affect (if at all?) the display of the rendered buffer on external video devices. This needs to be tested extensively.

I have tested on Linux64 and have seen no adverse effects.

**Note for QA**: Please test this extensively on all 3 platforms, and with all VideoDevices (NDI, SDI, AJA, BlackMagic), to make sure this causes no problem.

n/a

### DIFF
--- a/src/lib/app/RvCommon/GLView.cpp
+++ b/src/lib/app/RvCommon/GLView.cpp
@@ -407,11 +407,12 @@ namespace Rv
             //          The issue I see on MacOS that the background is brown
             //          istead of black with the default background.
             //
-            // NOTE PB: Not sure wy we just can't do the following 5 lines of 
-            // gl code below for all cases, for the "else" statement as well. 
-            // Are the glPush/glPopAttrib, and the glClear of also the depth 
-            // buffer that bad on modern hardware? It would simplify this 
-            // already-complicated function.
+            // Even on Qt6/QOpenGLWidget, we need to call 
+            // glBindFramebuffer(); it otherwise complains and fails 
+            // on glClear() on macOS
+            glBindFramebuffer(
+                GL_FRAMEBUFFER, 
+                QOpenGLContext::currentContext()->defaultFramebufferObject());
 
             glPushAttrib(GL_COLOR_BUFFER_BIT);
             TWK_GLDEBUG;
--- a/src/lib/app/RvCommon/GLView.cpp
+++ b/src/lib/app/RvCommon/GLView.cpp
@@ -335,7 +335,7 @@ namespace Rv
     {
         IPCore::Session* session = m_doc->session();
         bool debug = IPCore::debugProfile && session;
-        
+
         if (!m_postFirstNonEmptyRender && session
             && session->postFirstNonEmptyRender())
         {
@@ -362,7 +362,7 @@ namespace Rv
     //  and holding them in the display even when rv redraws.  the
     //  effect being that parts of previous displays will be "left
     //  behind" and not updated even when rv plays.  especially when
-    //  going to/from fullscreen. 
+    //  going to/from fullscreen.
     //
 #ifdef PLATFORM_WINDOWS
     SetWindowRgn (this->winId(), 0, false);
@@ -411,7 +411,7 @@ namespace Rv
             // glBindFramebuffer(); it otherwise complains and fails 
             // on glClear() on macOS
             glBindFramebuffer(
-                GL_FRAMEBUFFER, 
+                GL_FRAMEBUFFER,
                 QOpenGLContext::currentContext()->defaultFramebufferObject());
 
             glPushAttrib(GL_COLOR_BUFFER_BIT);
--- a/src/lib/app/RvCommon/GLView.cpp
+++ b/src/lib/app/RvCommon/GLView.cpp
@@ -330,22 +330,12 @@ namespace Rv
 #endif
     }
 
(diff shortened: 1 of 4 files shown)


src/lib/app/RvCommon/GLView.cpp
```diff
@@ -330,22 +330,12 @@ namespace Rv
...
-                             GL_COLOR_BUFFER_BIT, GL_NEAREST);
-        glDrawBuffer(GL_BACK);
-    }
 
     void GLView::paintGL()
     {
         IPCore::Session* session = m_doc->session();
         bool debug = IPCore::debugProfile && session;
-
+
```
bernie-laberge, line 338: No biggie but have you forgotten to install the pre-commit on that clone by chance ?
`pre-commit install`

src/lib/app/RvCommon/GLView.cpp
```diff
@@ -416,16 +406,12 @@ namespace Rv
...
-            // NOTE_QT: glClear causes an issue on MacOS. We need to call
-            // glBindFramebuffer.
-
-            // Not sure why it is not complaining on Linux or Windows, but this
-            // make sure that we are drawing onto the default framebuffer with
-            // those OpenGL functions below.
-            glBindFramebuffer(
-                GL_FRAMEBUFFER,
-                QOpenGLContext::currentContext()->defaultFramebufferObject());
+            //
```
bernie-laberge, line 409: I would get rid of this whole comment section : from //Linux... to // alreadu-complicated function.
If it works as expected, then great. We don't need these comments.
The original code dates back 16 years so these limitations are surely long gone.

src/lib/app/RvCommon/GLView.cpp
```diff
@@ -480,6 +466,37 @@ namespace Rv
...
+        // comments in the previous version of this tile, it appears that 
+        // calling swapBuffers was done to force a quicker visual update, or 
+        // to minimize visual tearing of some sort. This would have worked 
+        // with the old QGLWidget (becayuse each QGLWidget had its own 
+        // context, and its rendering target was directly the GL_BACK 
+        // framebuffer, but, again, with QOpenGLWidget, the rendering target
+        // is no longer GL_BACK, it is an FBO that is meant to be used at 
+        // the end of the application's drawing / visual update pipeline.
+        // 
+        // I am not sure hwo this would affect (if at all?) the display of
```
bernie-laberge, line 496: I would remove this comment section from // I am not... 
Let's assume that it will work. And if not then you'll fix it and add the relevant coments.
Otherwise those comments will remain in the code forever :-)

src/lib/app/RvCommon/RvApplication.cpp
```diff
@@ -1595,7 +1595,7 @@ namespace Rv
         Rv::Session* session = static_cast<Rv::Session*>(doc);
         RvDocument* rvDoc = (RvDocument*)session->opaquePointer();
 
-        if (m_presentationMode)
+            if (m_presentationMode)
```
bernie-laberge, line 1598: pre-commit ?

src/lib/app/RvCommon/RvApplication.cpp
```diff
@@ -1628,7 +1628,6 @@ namespace Rv
                                    == QApplication::screens().indexOf(
                                        QApplication::screenAt(
                                            rvDoc->mapToGlobal(QPoint(0, 0)))))
-                            ;
```
bernie-laberge, line 1631: Great catch !!!

bernie-laberge (approved):
Beautiful. Thanks!

cedrik-fuoco-adsk (approved):

PR #708, opened 2025-03-05
Fix MMR support in OTIO reader

bernie-laberge:
### Linked issues
NA

There were a couple issues associated with multiple media representation support in OTIO reader:
- Local file path could not be used because the server URL was always preprended.
- When creating a video track, its duration was 1 frame too long
- New OTIO based media representation names were inconsistent with the previously named MMR in RV such as 'mp4', 'path_to_movie', and 'path_to_frames'

### Summarize your change.

- Now only prepending the server URL when the target url starts with /file_serve/version
- Fixed the duration in  _set_sequence_edl()
- Renamed known media representation names to standard media rep names used by RV such as Streaming, Movie, and Frames.

Successfully tested on macOS. The fix is not OS specific

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -469,7 +469,7 @@ jobs:
           python -c "import sys; print(sys.executable)"
 
       - name: Cache CMake for Windows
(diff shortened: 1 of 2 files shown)


eloisebrosseau (approved):
LGTM!

cedrik-fuoco-adsk (approved):

PR #709, opened 2025-03-05
Feature/Qt6 - Fix hdpi issue that magically fixes itself when resizing the main window

pbergeron-adsk:
### Summarize your change.

Added a singleshot timer to force a resize event to the main window, in order to fix an image scale issue when first starting RV.

On macOS (and, presumably, any platform that supports HDPI) the initial display of the main view's image will not be scaled properly. After much investigation, this appears to be a Qt bug but were being unable to truly identify the root cause (coordinates are ok, pixel scale is ok, pixel ratio is ok, gl surface is ok, gl viewport is ok, etc etc. 

Oddly enough, the issue disappeared when resizing the view, even 1 pixel, but this has to be done after the view is first shown.

As a result, we dimply decided to force two resize events to occur (+1 pixel, and then -1 pixel). This has no visible effect on startup (or, you'd really need eagle eyes) to go: "A-HA!"

This was reported on macOS, so I tested it on macOS, but it's possible it may occur on other platofrms as well. Unfortunately I did not have HDPI displays where the OS does pixel-doubling by itself like on macOS.

This is a simple change in RvDocument, where  we initially set up the main view.

No need, simply confirm that RV now starts correctly on macOS

### DIFF
--- a/src/lib/app/RvCommon/RvDocument.cpp
+++ b/src/lib/app/RvCommon/RvDocument.cpp
@@ -187,6 +187,26 @@ namespace Rv
         m_glView->setFocus(Qt::OtherFocusReason);
         // qApp->installEventFilter(m_glView);
 
+        // #ifdef PLATFORM_DARWIN
+        //  Under macOS, for Qt6/QOpenGLWidget port,
+        //  the initial display shows incorrect pixel
+        //  scaling due to some sort of effect related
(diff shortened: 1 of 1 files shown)


eloisebrosseau (approved):

bernie-laberge (approved):
LGTM

PR #714, opened 2025-03-11
Feature/Qt6 - Fix various qt console warnings

pbergeron-adsk:
This fixes SG-36654 SG-36655 SG-36656 

### Summarize your change.

Added a method to filter out some Qt messages that are known/harmless warnings.

One such warning had Qt reporting a warning about a scroll event occuring from a HID device whenever it was connected/disconnected/kvm'd.   This appears to be fixed with Qt 6.6.3 but we are at 6.5.3, so we shut it up.

Another such warning was about a Qt handler for a type conversion already registered (but when initializing PyQt it was tried to be initialized again), etc.

Pollutes the console window needlessly with harmless warnings which are Qt bugs that were later fixed.

macOS and Linux.

### DIFF
--- a/src/lib/app/RvCommon/RvApplication.cpp
+++ b/src/lib/app/RvCommon/RvApplication.cpp
@@ -99,7 +99,7 @@ namespace
 
     bool isSilencedQtMessage(const string_view& text)
     {
-        array<string_view, 5> silenced = {
+        const array<string_view, 5> silenced = {
             // In Qt6, when KVM or YubiKey or other HID device is plugged in or
             // out. This warning seems fixed with QT 6.6.3
             "scroll event from unregistered device"sv,
@@ -160,7 +160,7 @@ namespace
             return;
         RecursionLock l;
 
-        string msg = qmsg.toUtf8().constData();
+        const string msg = qmsg.toUtf8().constData();
 
         // We always report Qt debug messages except for known issues.
         // In which case the following function can be used to check
--- a/src/lib/app/RvCommon/RvApplication.cpp
+++ b/src/lib/app/RvCommon/RvApplication.cpp
@@ -92,6 +92,127 @@ static void init()
 extern void rvLazyBuildThirdPartyCustomization();
 #endif
 
+namespace
+{
+    using namespace boost;
+    using namespace std;
+
+    bool isSilencedQtMessage(const string_view& text)
+    {
+        array<string_view, 5> silenced = {
+            // In Qt6, when KVM or YubiKey or other HID device is plugged in or
+            // out. This warning seems fixed with QT 6.6.3
+            "scroll event from unregistered device"sv,
+
+            // Another relic of an unneeded warning resulting from the Qt6 port.
+            // This one originates when loading Python which calls
+            // PyInit_QtConcurrent(), which then calls
+            // qRegisterMetaType<QFuture<QString>>(char const*){}
+            // This is just a warning that the registration is already done, so
+            // we can safely ignore it.
+            "Type conversion already registered from type"
+            " QFuture<QString> to type QFuture<void>"sv,
+
+            // Qt warning about KVO observers. Qt have gotten rid of the check
+            // for this error in Qt 6.0. "ERROR: has active key-value observers
+            //(KVO)!
+            //        These will stop working now that the window is recreated,
+            //        and will result in exceptions when the observers are
+            //        removed. Break in QCocoaWindow::recreateWindowIfNeeded to
+            //        debug."
+            "has active key-value observers (KVO)! These will stop"
+            " working now that the window is recreated"sv,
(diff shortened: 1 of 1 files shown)


src/lib/app/RvCommon/RvApplication.cpp
```diff
@@ -114,8 +114,57 @@ namespace Rv
...
+        while (!silenced[i].empty())
+        {
+            if (text.find(silenced[i]) != string::npos)
+                return true;
+
+            i++;
+        }
+
+        return false;
+    }
```
eloisebrosseau, line 164: You can replace `const string&` with `string_view` in the function argument. The C-style array of strings can be replaced with `std::array` containing `string_view literals`. Maybe I am missing something, but I am not exactly sure why you are using a while loop. I think you could replaced it with `std::any_of` for readability.

It could look something like this:
```
bool isSilencedQtMessage(std::string_view text)
{
    using namespace std::literals::string_view_literals;
    
    constexpr std::array silenced = {
        "scroll event from unregistered device"sv,
        "Type conversion already registered from type QFuture<QString> to type QFuture<void>"sv,
        "has active key-value observers (KVO)! These will stop working now that the window is recreated"sv,
        "This tablet device is unknown"sv,
        "Release of profile requested but WebEnginePage still not deleted"sv
    };
    
    return std::any_of(silenced.begin(), silenced.end(), [text](const auto& msg) {
        return text.find(msg) != std::string::npos;
    });
```

src/lib/app/RvCommon/RvApplication.cpp
```diff
@@ -127,6 +176,12 @@ namespace Rv
 
         string msg = qmsg.toUtf8().constData();
```
eloisebrosseau, line 177: `qmsg` should be passed as a `QStringView` instead of a `const QString&` and `msg` should be a `const string_view`

src/lib/app/RvCommon/RvApplication.cpp
```diff
@@ -114,8 +114,57 @@ namespace Rv
...
+            if (text.find(silenced[i]) != string::npos)
+                return true;
+
+            i++;
+        }
+
+        return false;
+    }
+
+    void myMsgHandlerQT(QtMsgType t, const QMessageLogContext& context,
```
eloisebrosseau, line 166: `myMsgHandler` and `isSilencedQtMessage` should be moved into an anonymous namespace

eloisebrosseau (changes requested):

src/lib/app/RvCommon/RvApplication.cpp
pbergeron-adsk: Ok, I've fixed it.  The while loop was because of old habits, and I find easier to read, but, admittedly less modern.  I've fixed it to your if statement which uses a lambda.

src/lib/app/RvCommon/RvApplication.cpp
pbergeron-adsk: Ok, if you insist.

src/lib/app/RvCommon/RvApplication.cpp
```diff
@@ -92,6 +92,127 @@ static void init()
 extern void rvLazyBuildThirdPartyCustomization();
 #endif
 
+namespace
+{
+    using namespace boost;
+    using namespace std;
+
+    bool isSilencedQtMessage(const string_view& text)
```
eloisebrosseau, line 100: It should be `std::string_view` not `const std::string_view&`

src/lib/app/RvCommon/RvApplication.cpp
eloisebrosseau: Thank you! Note that the `text` parameter should be a `std::string_view` not a `const std::string_view&`

src/lib/app/RvCommon/RvApplication.cpp
eloisebrosseau: You are totally right, my mistake. But I think `msg` should still be const, am I wrong?

pbergeron-adsk: That was fixed also.

src/lib/app/RvCommon/RvApplication.cpp
pbergeron-adsk: They were not in the global namespace, they were in the Rv namespace, along with other classes/methods, but, it's ok, they are not in an anonymous namespace.

eloisebrosseau (approved):

bernie-laberge (approved):

cedrik-fuoco-adsk (approved):

cedrik-fuoco-adsk:
Can I resolved all the conversations? @eloisebrosseau

PR #718, opened 2025-03-11
Feature/qt6 - Support both Qt 5 and Qt 6 in rvcmd.sh

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
I added checks to detect existing installations of Qt 5 or Qt 6. The search prioritise Qt 6 over Qt 5. 
Once Qt is found, the script set the right variables for the aliases. 

If QT_HOME is already set, I added a check to detect if it is Qt 5 or Qt 6.

Note that the logic only look for `Qt*` and either `5.15*` or `6.5*` in the path. The script is not looking at the actual libraries files.

rvcmd.sh had to be reworked to support both Qt 5 and Qt 6.

Rocky Linux 8

### DIFF
--- a/rvcmds.sh
+++ b/rvcmds.sh
@@ -49,29 +49,58 @@ if [ -z "$QT_HOME" ]; then
   echo "Searching for Qt installation..."
 
   if [[ "$OSTYPE" == "linux"* ]]; then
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
Beautiful !
Thanks!

PR #720, opened 2025-03-12
Feature/Qt6 - Fix OpenGL Context corruption after exiting Desktop Presentation Mode

pbergeron-adsk:
### Summarize your change.

This is an initial code fix while debugging and fixing SG-25296 (can't use 2nd monitor desktop window as presentation device on macOS M computers). This change does NOT Fix the issue (we still get a black screen) but during the port to Qt6, we introduced a change that deleted the current FBO in the current OpenGL context (a GLFBO class destroyed a FBO object that was owned by Qt, not by GLFBO) and made the GLView class unstable. This fix only addresses the incorrect OpenGL FBO destruction that occurs after exiting presentation mode.

More concretely: GLFBO has 2 constructors: one where a FBO object is created using glGenFramebuffersEXT (in which case the GLFBO instance owns the FBO) and another one where the FBO handle is passed as an argument to the constructor (in which case the GLFBO instance does not own the FBO).  Upon calling the destructor, any non-zero handle was destroyed, but since the non-owned FBOs actually belonged to Qt, we ended up corrupting the current OpenGL context because it no longer had a rendering surface to render onto. To fix this, I have just added a bool member to GLFBO to guide what happens in the GLFBO destructor.

Also, while investigating this issue, I noticed that several places missed a call to the de debug macro TWK_GLDEBUG, so I have added them.

Finally, I updated the formatting of the error log for gl errors, which was overly verbose.

After exiting 2nd-monitor-desktop presentation mode, the OpenGL context got scrapped because an FBO object was mistakenly destroyed, which caused RV to stop displaying content until the view got resize/recreated with new FBOs.

macOS. Should work just the same on other platforms.

n/a

n/a

### DIFF
--- a/src/lib/app/RvCommon/GLView.cpp
+++ b/src/lib/app/RvCommon/GLView.cpp
@@ -369,7 +369,6 @@ namespace Rv
         // image.save("/home/<username>>/<orv_folder>/fbo.png");
     }
 
-
     void GLView::paintGL()
     {
         IPCore::Session* session = m_doc->session();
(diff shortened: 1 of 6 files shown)


cedrik-fuoco-adsk:
You did not select the right branch to merge it in. That's why you have 2000 changes. @pbergeron-adsk 
You will need to close and re-open the PR I think.

PR #724, opened 2025-03-18
Feature/qt6 - Clean up code in QTAudioRenderer class

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Removed #ifdef related to RV_VFX_CY2023 since they are not needed in QTAudioRenderer class (QT5AudioRenderer is used for CY2023).

Clean up code to make it more readable.

Rocky Linux 8

### DIFF
--- a/src/lib/audio/QTAudioRenderer/QTAudioRenderer.cpp
+++ b/src/lib/audio/QTAudioRenderer/QTAudioRenderer.cpp
@@ -18,6 +18,7 @@
 #include <QtCore/qendian.h>
 #include <QtCore/QCoreApplication>
 #include <QtCore/QDir>
+#include <QSysInfo>
 
 #include <iostream>
 #include <sstream>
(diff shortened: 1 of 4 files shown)


src/lib/audio/QTAudioRenderer/QTAudioRenderer.cpp
```diff
@@ -655,21 +642,29 @@ namespace IPCore
...
+                    bool isLittleEndian = false;
+
+                    union
+                    {
+                        uint32_t i;
+                        char c[4];
+                    } test = {0x01020304};
+
+                    isLittleEndian = (test.c[0] == 0x04);
+
```
cedrik-fuoco-adsk, line 661: This is new. I am using this to detect if the system is little or big endian.

bernie-laberge: And where are you using this isLittleEndian variable ?
The PowerPC were end of life in 2006. I think we can safely assume that the system is Little Endian right ?

bernie-laberge (approved):
Nice cleanup !

src/lib/audio/QTAudioRenderer/QTAudioRenderer.cpp
cedrik-fuoco-adsk: oh, it suppose to be used in AudioRenderer::transformFloat32ToInt24 below, on line 662. Forgot to add it.

eloisebrosseau: Based on [CPP Core Guidelines](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c183-dont-use-a-union-for-type-punning), I think we should avoid using union for type punning. Like Bernard said, if we can safely assume the system is Little Endian, I believe it would be better to not rely on that piece of code

cedrik-fuoco-adsk: I updated the code. The code will assume that the system is little endian.

eloisebrosseau (approved):

PR #725, opened 2025-03-19
Fix the failing readthedocs build

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
I added the `configuration` key to the `.readthedocs.yaml` because it is required.

Since December 2024, the `configuration` key is required.
See https://about.readthedocs.com/blog/2024/12/deprecate-config-files-without-sphinx-or-mkdocs-config/

### DIFF
--- a/.readthedocs.yaml
+++ b/.readthedocs.yaml
@@ -11,6 +11,8 @@ python:
     - requirements: docs/requirements.txt
    
 sphinx:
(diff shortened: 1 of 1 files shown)


eloisebrosseau (approved):

bernie-laberge (approved):
Thanks!

PR #727, opened 2025-03-24
Remove show property from annotation hook

eloisebrosseau:
### Summarize your change.

The `visible` layer attribute is not used anymore to indicate if annotations should be shown or not. Therefore, the `show` attribute of the RVPaint node should not be set with that attribute value. Show and hide annotation actions are now only being handled in the Live Review plugin. The required changes were made in this [PR](https://git.autodesk.com/media-and-entertainment/rv/pull/336).

Since Creative Review is now sending an event as part of the `PLAYBACK_SETTINGS` command, we should not use the `visible` attribute of the OTIO Timeline to know if the annotations should be shown or not.

This change was tested on macOS in a Live Review session between RV and the Creative Review prototype.

### DIFF
--- a/cmake/dependencies/python3.cmake
+++ b/cmake/dependencies/python3.cmake
@@ -20,17 +20,9 @@ SET(PYTHON_VERSION_MAJOR
     3
 )
 
-RV_VFX_SET_VARIABLE(
-  PYTHON_VERSION_MINOR
-  CY2023 "10"
-  CY2024 "11"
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
LGTM

PR #729, opened 2025-03-26
Add PNG as slow random access codec

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Adding PNG codec to the slow random access array and changing the seeking offset to -1 instead of -3 because RV was using that an offset of -1 before FFmpeg 6.

There are some issues when RV is using two readers threads and FFmpeg multithreading. By adding PNG as a slow random access codec, the decoding performance is a lot better.

Rocky Linux 8

An offset of -1 during seeking had better performance during my testing.

### DIFF
--- a/cmake/dependencies/ffmpeg.cmake
+++ b/cmake/dependencies/ffmpeg.cmake
@@ -247,6 +247,13 @@ IF(RV_TARGET_WINDOWS)
   LIST(APPEND RV_FFMPEG_COMMON_CONFIG_OPTIONS "--toolchain=msvc")
 ENDIF()
 
+# Change the condition to TRUE to be able to debug into FFmpeg.
+IF(FALSE)
+  LIST(APPEND RV_FFMPEG_COMMON_CONFIG_OPTIONS "--disable-optimizations")
+  LIST(APPEND RV_FFMPEG_COMMON_CONFIG_OPTIONS "--enable-debug=3")
+  LIST(APPEND RV_FFMPEG_COMMON_CONFIG_OPTIONS "--disable-stripping")
+ENDIF()
+
 # Controls the EXTERNALPROJECT_ADD/BUILD_ALWAYS option
 SET(${_force_rebuild}
     FALSE
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -24,6 +24,7 @@
 #include <math.h>
 #include <algorithm>
 #include <array>
+#include <string_view>
 #include <stdlib.h>
 #include <stl_ext/stl_ext_algo.h>
 #include <stl_ext/string_algo.h>
@@ -70,7 +71,6 @@ namespace TwkMovie
 #define FPS_PRECISION_LIMIT 1000 // 5 sig figs
 #define RV_OUTPUT_VIDEO_CODEC "mjpeg"
 #define RV_OUTPUT_AUDIO_CODEC "pcm_s16be"
-#define RV_SEEK_FRAME_OFFSET 1
 
 #if 0
 #define DB_LOG_LEVEL AV_LOG_ERROR
@@ -620,6 +620,7 @@ namespace TwkMovie
 
     namespace
     {
+        constexpr int rv_seek_frame_offset = 1;
 
         //----------------------------------------------------------------------
         //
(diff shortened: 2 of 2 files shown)


cmake/dependencies/ffmpeg.cmake
```diff
@@ -247,6 +247,13 @@ IF(RV_TARGET_WINDOWS)
   LIST(APPEND RV_FFMPEG_COMMON_CONFIG_OPTIONS "--toolchain=msvc")
 ENDIF()
 
+# Change the condition to TRUE to be able to debug into FFmpeg.
+IF(FALSE)
+  LIST(APPEND RV_FFMPEG_COMMON_CONFIG_OPTIONS "--disable-optimizations")
+  LIST(APPEND RV_FFMPEG_COMMON_CONFIG_OPTIONS "--enable-debug=3")
+  LIST(APPEND RV_FFMPEG_COMMON_CONFIG_OPTIONS "--disable-stripping")
+ENDIF()
+
```
cedrik-fuoco-adsk, line 256: I had commented those lines before, the clang-format was putting them all in one line, so I decided to it using IF(FALSE) instead.

I did not add an OPTION() because we do not use those flags often, but they are useful sometimes.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -3347,22 +3317,25 @@ namespace TwkMovie
...
-        while (ret != AVERROR_EOF)
-            ret =
-                avcodec_receive_frame(track->avCodecContext, track->videoFrame);
+        int ret;
+        while ((ret = avcodec_receive_frame(track->avCodecContext,
+                                            track->videoFrame))
+               != AVERROR_EOF)
+        {
+            // Discard received frames until the end of the stream is reached
+        }
```
cedrik-fuoco-adsk, line 3338: I think that was changed by the clang-format.

cmake/dependencies/ffmpeg.cmake
```diff
@@ -247,6 +247,13 @@ IF(RV_TARGET_WINDOWS)
   LIST(APPEND RV_FFMPEG_COMMON_CONFIG_OPTIONS "--toolchain=msvc")
 ENDIF()
 
+# Change the condition to TRUE to be able to debug into FFmpeg.
+IF(FALSE)
```
bernie-laberge, line 251: Nice ! Thanks !

bernie-laberge (approved):
LGTM

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -70,6 +70,7 @@ namespace TwkMovie
 #define FPS_PRECISION_LIMIT 1000 // 5 sig figs
 #define RV_OUTPUT_VIDEO_CODEC "mjpeg"
 #define RV_OUTPUT_AUDIO_CODEC "pcm_s16be"
+#define RV_SEEK_FRAME_OFFSET 1
```
eloisebrosseau, line 73: Preprocessor macros should not be used for named constants. I would replace it with `constexpr int rv_seek_frame_offset = 1`

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -631,49 +632,18 @@ namespace TwkMovie
...
+            "div1", "div2", "div3",       "div4",
+            "div5", "div6", "divx",       "dnxhd",
+            "dx50", "h263", "h264",       "i263",
+            "iv31", "iv32", "m4s2",       "mp42",
+            "mp43", "mp4s", "mp4v",       "mpeg4",
+            "mpg1", "mpg3", "mpg4",       "pim1",
+            "png",  "s263", "svq1",       "svq3",
+            "u263", "vc1",  "vc1_vdpau",  "vc1image",
+            "viv1", "wmv3", "wmv3_vdpau", "wmv3image",
+            "xith", "xvid", "libdav1d",   0};
```
eloisebrosseau, line 646: Since you made a change to `slowRandomAccessCodecsArray`, I think it would be nice if you could look into transforming it into a `constexpr std::array` instead of C-style array. It should look something like the suggestion I made to Patrick [here](https://github.com/AcademySoftwareFoundation/OpenRV/pull/714#discussion_r1994002025).

 I think we should also avoid naming our variable with their type, and just name them for what they hold, so I would rename it to simply `slowRandomAccessCodecs`.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
cedrik-fuoco-adsk: I will change it in the next commit

cedrik-fuoco-adsk:
All the comments has been address in the previous commit.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -753,15 +722,17 @@ namespace TwkMovie
         bool codecHasSlowAccess(string name)
```
eloisebrosseau, line 722: Forgot to mention it in my previous comment, but the `name` parameter should be a `std::string_view` instead of a `std::string`.

PR #736, opened 2025-04-01
Feature/Qt6: Fix broken Desktop Presentation Device for all platforms

pbergeron-adsk:

With the port to Qt6/QOpenGLWidget, the DesktopVideoDevice implementation was broken for all platforms (macOS, Linux, Windows).  This is a long-time desired fix, since on Apple-Silicon macOS Arm64, DesktopPresentationDevice was broken for over 3 years.

This PR aims to restore desktop presentation device functionality, by using the common Qt path to open a full-screen desktop window on the target screen. This was the path taken for Windows and Linux, but for macOS, there was a special CGDesktopVideoDevice, which used Apple's CoreGraphics API to achieve this.

### Summarize your change.

This PR contains a surprisingly large amount of code changes to restore this functionality using QOpenGLWidget.  The main difference is that where QGLWidget used the default BackBuffer to render to screen (GL FBO id 0), this is no longer allowed using QOpenGLWidget, which exclusively uses off-screen FBOs to render the view. Previous RV code made many, many assumptions about when it was safe to use FBO 0, and this was the root cause of the problem. The other major cause was that during RV's rendering of the session, the temporary FBOs created for the presenation device were created in the wrong OpenGL context. Somehow, this used to "work" in the old implementation (presumably because we were rendering to FBO 0 so it didn't matter) but when we started using actual off-screen FBOs, the FBO for the external desktop view never had any content rendered to it.

Thus, the approach taken for this fix are as follows:

0) In RV main for macOS, Added setAttribute to allow accessing GL from different threads. This was enabled for Windows and Linux for the QT path, so we also need it.

1) Since we decided to use the Qt path for the core technology to render to another full-screen display, I merged the relevant (specialized) code from class QTDesktopPresentationDevice back into class DesktopPresentationDevice.

2) The QTDesktopVideoDevice (or was it DesktopVideoDevice?)'s bind() and bind2() methods were removed, which was causing the FBOs to be created in the wrong opengl context.  This legacy code caused a very mysterious issue where the FBO to blit to the external screen was always empty. This is because the FBO for the screen was created in the wrong context. Fwiw, no other external video device overrode bind() and bind2() either... not sure why this was ever done in the first place.

3) In DesktopVideoModule, we checked the 3 platforms and added platform-specific code to enumerate and instantiate the proper DesktopVideoDevices. This enumeration/instantiation code was moved the various implementations of the VideoDevice classes (DesktopVideoDevice, CGDesktopVideoDevice - no longer used for now, CGDesktopVideoDeviceArm - no longer used for now, but soon)

4) On macOS, showng the rendered framebuffer on the screen was done using some kind of janky multi-threaded swapbuffer kludge.  That kludge was completely removed, and besides, we don't use swapBuffer anymore -- it's not allowed with QOpenGLWidget - we simply force the QOpenGLWidget to repaint itself in the main thread after the transfer function is called. (See SyncBufferThreadData stuff that was removed in GLView)

5) An important, but subtle new implementation note, has to do with changes to GLFBO, which has 2 constructors: The first one, where an FBO is explicitly created at resolution width/height, and with its own pixel buffer (aka render buffer), and the Second, where an GLFBO is created to attach to a video device, which was originally thought to output to a "default" framebuffer (aka: FBO 0, the GL backbuffer directly, which is not allowed anymore).  In an early version of the FeatureQt6, the id of the FBO was take from the OpenGL Current context, but, this was incorrect.  The real solution is that the id of the GLFBO, when created with constructor 2, is in fact the ID of the FBO owned by the QOpenGLWidget attached to the VideoDevice. It is NOT the id of the current context (which happened to work when there was no DesktopVideoDevice). Therefore, I updated "GLFBO::fboId()", to not return m_id, but to return the id of the QOpenGLWidget's fbo id. It's important to always query the view, because the QOpenGLWidget's fbo id might actually be recycled/changed if the view is changed/resized/etc. Querying the ID of the widget's fbo follows the same patter as asking for the width/height of the fbo when using the constructor 2.

6) At the moment, the DesktopVideoModule initialization method is still a bit janky. You'll see this with some code that checks which platform we're running on, even though we now we always instantiate the DesktopVideoDevice class (the same Qt class).  I've left this platform-checking code because It will be required anyway to fix 25296 properly, and I'll clean that a bit better for that PR, see the note below.

7) As part of this work, I had to implement debugging tools to see what was going on with the state of the current openGL context, and the state of the currently bound FBO. In GL.cpp, I've implemented an OpenGL Context Tracker, which is updated every time the TWK_GLDEBUG; macro is invoked. This tracker will query the opengl state and currently bound context, and will report any changes to the currently bound opengl state. I could have destroyed this code, but I thought it would be helpful in the future, so I've left it in GL.cpp but deactivated it through a #define.

PLEASE NOTE: 

This PR is mainly a fix for 38492-Broken-on-all-platforms and thus successfully restores functionality on macOS (25296).

However we do lose a bit of functionality on macOS versus the old RV -- in particular the ability to select the target screen's resolution and change the screen's resolution when activating presentation mode (because we use the Qt path instead of the CoreGraphics path). A lot of code in this PR is necessary to fully implement 25296, and since I was working on both at the same time, some of the code serving as groundwork to fully fix 25296 is included in this PR, but I've decided to break it into 2 PRs.

(That groundwork code is found in CGDesktopVideoDevice.cpp and CGDesktopVideoDeviceArm.cpp)

**I have tested this on macOS, but testing on Windows and Linux is needed.**

I also suspect that the Stereo or Quad mode remains broken. But since I have no way to test this. I will create another Jira ticket to address this.

As per above, we need to test this functionality EXTENSIVELY, on all platforms.

For MacOS, support for DisplayLink was removed, however DisplayLink  may already be transparently taken care of using the Qt implementation of the DesktopVideoDevice. (I have no way to test this). Even if this does not work, Bernard approved its sacrifice.


src/lib/app/RvCommon/CGDesktopVideoDeviceArm.cpp
```diff
@@ -0,0 +1,544 @@
+//
```
bernie-laberge, line 1: You could put the standard header:
```
//
//  Copyright (c) 2025 Autodesk, Inc. All Rights Reserved.
//
//  SPDX-License-Identifier: Apache-2.0
//
```

src/lib/app/RvCommon/CGDesktopVideoDevice.cpp
```diff
@@ -538,4 +548,91 @@ namespace Rv
         return m_colorProfile;
     }
 
+    std::vector<VideoDevice*> CGDesktopVideoDevice::createDesktopVideoDevices(
+        TwkApp::VideoModule* module, const QTGLVideoDevice* shareDevice)
+    {
+        std::vector<VideoDevice*> devices;
+
+        // this used to be in DesktopVideoModule.cpp but moved here
```
bernie-laberge, line 556: Good idea!

src/lib/app/RvCommon/CMakeLists.txt
```diff
@@ -62,8 +61,10 @@ SET(_sources
     RvJavaScriptObject.cpp
 )
 
+# IF(RV_TARGET_DARWIN) LIST(APPEND _sources CGDesktopVideoDevice.cpp CGDesktopVideoDeviceArm.cpp DisplayLink.cpp) ENDIF()
+
```
cedrik-fuoco-adsk, line 65: Is that line suppose to be commented out? I don't see where `CGDesktopVideoDevice.cpp` and `DesktopVideoDeviceArm.cpp` are added to the `_sources` list.

You can use `RV_TARGET_APPLE_ARM64` to differentiate with Apple Intel.

src/lib/app/RvCommon/DesktopVideoModule.cpp
```diff
@@ -32,123 +29,63 @@ namespace Rv
...
-        //  Use the CoreGraphics devices
-        //
-
-        CGDirectDisplayID idArray[20];
-        uint32_t idNum;
-        CGGetOnlineDisplayList(20, idArray, &idNum);
-        // CGDirectDisplayID mainID = CGMainDisplayID();
-
-        for (size_t i = 0; i < idNum; i++)
+        if (isDarwinArm()) // Apple Arm
```
cedrik-fuoco-adsk, line 66: Letting you know that there two compiler define available to differentiate `x86_64` and `arm64` on Apple:
`RV_ARCH_X86_64` and `RV_ARCH_ARM64`

src/lib/app/RvCommon/DesktopVideoModule.cpp
```diff
@@ -166,7 +103,7 @@ namespace Rv
     {
         TwkApp::VideoDevice* device = 0;
 
-#if defined(PLATFORM_DARWIN)
+#if defined(PLATFORM_DARWIN) && 0
```
cedrik-fuoco-adsk, line 106: I think it would be a good idea to add a comment to explain and to not forget to change this later? I assume this is temporary.

src/lib/app/RvCommon/DesktopVideoDevice.cpp
cedrik-fuoco-adsk, line 1: Windows is complaining about `use of undefined type` for `QScreen` and `QWindow`

src/lib/app/RvCommon/RvDocument.cpp
```diff
@@ -490,7 +488,7 @@ namespace Rv
 
     void RvDocument::playStopSlot(const std::string&)
     {
-#ifdef PLATFORM_DARWIN
+#if defined(PLATFORM_DARWIN) && 0
```
cedrik-fuoco-adsk, line 491: I assume this is temporary until the other fix for mac ?

src/lib/app/RvCommon/CGDesktopVideoDeviceArm.cpp
```diff
@@ -0,0 +1,544 @@
...
+        //
+        //   Docs for CGFlushDrawable say it includes a glFlush and user should
+        //   "should not call glFlush() prior", so don't call it here.
+        //
+        //   ARG!  This breaks presentation mode.  putting it back in.
+        //
+
+        glFlush();
+#ifdef __PB
+        CGLFlushDrawable(m_context);
```
bernie-laberge, line 123: Perhaps simply remove it if it is no longer necessary

src/lib/app/RvCommon/CGDesktopVideoDeviceArm.cpp
```diff
@@ -0,0 +1,544 @@
...
+
+        // module()
+        TwkApp::VideoModule* m = const_cast<TwkApp::VideoModule*>(module());
+
+        m_viewDevice =
+            new QTGLVideoDevice(m, "AppleSilicon Presentation Device", &m_view);
+        m_viewDevice->makeCurrent();
+
+        m_view.showFullScreen();
+        /*
```
bernie-laberge, line 164: If I understand correctly : This file is not used yet (it will be used when you fix the DesktopVideoDevice on Apple Silicon. 
Perhaps it would be best not to include CGDesktopVideoDeviceArm.cpp and .h in this PR and save it for the DesktopVideoDevice fix on Apple Silicon.

src/lib/app/RvCommon/CMakeLists.txt
bernie-laberge: I don't think Patrick needs it in this PR. It will be added in a future PR for the DesktopVideoDevice support on Apple Silicon. So this line should be removed in this PR.

src/lib/app/RvCommon/DesktopVideoModule.cpp
```diff
@@ -32,123 +29,63 @@ namespace Rv
...
+                }
+                return ret > 0;
+            #else
+                return false;
+            #endif
+        }
+    */
+
+    // force intel code path for this commit
+    static bool isDarwinArm() { return true; }
```
bernie-laberge, line 54: I would remove all this and only support plain vanilla DesktopVideoDevice for all platforms in this PR.

src/lib/app/RvCommon/DesktopVideoModule.cpp
```diff
@@ -32,123 +29,63 @@ namespace Rv
...
-            CFStringRef localName = NULL;
-            io_connect_t displayPort = CGDisplayIOServicePort(aID);
-            CFDictionaryRef dict =
-                (CFDictionaryRef)IODisplayCreateInfoDictionary(
-                    displayPort, kIODisplayOnlyPreferredName);
-            CFDictionaryRef names = (CFDictionaryRef)CFDictionaryGetValue(
-                dict, CFSTR(kDisplayProductName));
-
-            if (names)
+            if (useQtOnDarwinArm())
```
bernie-laberge, line 68: If I understand correctly, in a future PR, we will have the option on Apple Silicon to have a plain vanilla Qt based DesktopVideoDevice OR a CGDesktopVideoDeviceArm that will be capable of setting a resolution which is a better option. How about we just add an environment variable to act as the useQtOnDarwinArm() ?
Like this:
`static ENVVAR_BOOL(evQtOnDarwinArm, "RV_USE_QT_ON_DARWIN_ARM", false);`
Again I would save that for the future PR that fixes the apple Silicon DesktopVideoDevice issue.

src/lib/app/RvCommon/DesktopVideoModule.cpp
```diff
@@ -32,123 +29,63 @@ namespace Rv
...
-                QString name = QString("%1 %2 %3")
-                                   .arg(w->manufacturer())
-                                   .arg(w->model())
-                                   .arg(w->name());
-
-                QTDesktopVideoDevice* sd = new QTDesktopVideoDevice(
-                    this, name.toUtf8().constData(), screen, shareDevice);
-                m_devices.push_back(sd);
-            }
+            m_devices = DesktopVideoDevice::createDesktopVideoDevices(
```
bernie-laberge, line 81: I would only keep this section, as this is the only one available, in this PR.
And save the CGDesktopVideoDevice and CGDesktopVideoDeviceArm for a future PR.

src/lib/app/RvCommon/DesktopVideoDevice.cpp
```diff
@@ -29,67 +36,187 @@ namespace Rv
...
     {
         m_glGlobalState = new GLState();
         m_glGlobalState->useGLProgram(textureRectGLProgram());
+
+        const QList<QScreen*> screens = QGuiApplication::screens();
+        if (screen < 0 || screen >= screens.size())
+        {
+            // Handle invalid screen index.
+            // TODO_QT
+            std::cout << "Screen index is out of range" << std::endl;
```
bernie-laberge, line 61: std::cerr

src/lib/app/RvCommon/DesktopVideoModule.cpp
pbergeron-adsk: Yes, but if it's a fat binary, the determination needs to be made at runtime, no?

src/lib/app/RvCommon/DesktopVideoModule.cpp
pbergeron-adsk: I will think of something better

src/lib/app/RvCommon/RvDocument.cpp
pbergeron-adsk: yes temporary I will fix this better

src/lib/app/RvCommon/CGDesktopVideoDeviceArm.cpp
pbergeron-adsk: This entire file is likely to be removed. I will derive DesktopVideoDeviceMacArm, where I Will migrate the resolution-shifting code there.

src/lib/app/RvCommon/CGDesktopVideoDeviceArm.cpp
pbergeron-adsk: No problem, I will remove it from the PR

src/lib/app/RvCommon/DesktopVideoModule.cpp
bernie-laberge: No: we build 2 separate binaries. One for mac Intel and another one for mac Apple Silicon

src/lib/app/RvCommon/DesktopVideoDevice.cpp
```diff
@@ -29,67 +36,187 @@ namespace Rv
...
+        , m_view(0)
     {
         m_glGlobalState = new GLState();
         m_glGlobalState->useGLProgram(textureRectGLProgram());
+
+        const QList<QScreen*> screens = QGuiApplication::screens();
+        if (screen < 0 || screen >= screens.size())
+        {
+            // Handle invalid screen index.
+            // TODO_QT
```
bernie-laberge, line 60: I would simply set the m_screen to 0 in that case since it will never happen anyway.

src/lib/app/RvCommon/DesktopVideoModule.cpp
pbergeron-adsk: no; both AppleArm and AppleIntel will use the Qt path; and both will use the same code for resolution switching. The apple implemenation using CoreGraphcis was a huge kludge relying on an external thread to trigger the swapbuffer, and it did other kludgey/ugly hacks to get this to work. 

It turns out that the Qt path works for both Mac Arm and Intel,  I'm pretty sure it can enumerate the displays and resolutions properly (and it can get rid of the naming "CGDisplay1/CGDisplay2" issues too).  The code that we have in the Arm implementation is also OK to cause the other displays to switch resolution. Because we use the Qt path, we will no longer need the CG implementation at all. We will have: 

DesktopVideoDevice.cpp, the generic QT path which works on all platforms but does not allow resolution switching

DesktopVideoDeviceMac.cpp,  derived from above but adds CG calls to enumerate and switch resolutions

and maybe (but not plans for now)
DesktopVideoDeviceWin.cpp, derived from above, but adds Win32 calls to enumerate and switch resolutions

and maybe (but no plans for now)
DesktopVideoDeviceLinux.cpp. same.

src/lib/app/RvCommon/DesktopVideoModule.cpp
pbergeron-adsk: that's fine I'll fix it.

src/lib/app/RvCommon/DesktopVideoDevice.cpp
pbergeron-adsk: Ok

src/lib/app/RvCommon/DesktopVideoDevice.cpp
cedrik-fuoco-adsk: I added that during my refactor and never got to it for clean up.

src/lib/app/RvCommon/DesktopVideoDevice.cpp
```diff
@@ -29,67 +36,187 @@ namespace Rv
...
+        fmt.setSwapInterval(m_vsync ? 1 : 0);
+
+        ScreenView* vw =
+            new ScreenView(fmt, 0, shareDevice()->widget(), Qt::Window);
+        setViewWidget(vw);
+
+        QTGLVideoDevice* vd = new QTGLVideoDevice(0, "local view", vw);
+        setViewDevice(vd);
+
+        const QList<QScreen*> screens = QGuiApplication::screens();
```
bernie-laberge, line 179: Please remove `screens` as it is no longer used (thanks to your new screenGeometry() method)

src/lib/app/RvCommon/DesktopVideoModule.cpp
bernie-laberge: Fantastic ! Thanks.

src/lib/app/RvCommon/RvCommon/CGDesktopVideoDeviceArm.h
```diff
@@ -0,0 +1,140 @@
+//
```
bernie-laberge, line 1: You could put the standard header:

//
//  Copyright (c) 2025 Autodesk, Inc. All Rights Reserved.
//
//  SPDX-License-Identifier: Apache-2.0
//

bernie-laberge (approved):
LGTM
Thank you @pbergeron-adsk !

src/lib/app/RvCommon/DesktopVideoModule.cpp
pbergeron-adsk: Ok; fixed. For now we just the qt path for all conditions.

src/lib/app/RvCommon/RvCommon/CGDesktopVideoDeviceArm.h
pbergeron-adsk: Fixed

src/lib/app/RvCommon/CGDesktopVideoDeviceArm.cpp
pbergeron-adsk: I need to keep it in the code to complete resolution switching later.

PR #738, opened 2025-04-02
SG-37377: Add Compare to Live Review

eloisebrosseau:
### Summarize your change.

- [X] Add Compare OTIO schema and hook
- [X] Add Opacity Compare mode in IPGraph and in OTIO compare hook
- [X] Add Wipe (angular mask) mode in IPGraph and in OTIO compare hook
- [X] Add Reverse Wipe shader to handle swapping versions
- [X] Update requirements.txt to use the main branch of the OTIO Github repo to get the latest fixes for OTIO V2d before the next release is available
- [X] `read_otio_string` was removed since the parsing of the OTIO Timeline is now down directly in the Live Review plugin
- [X] Handle OTIO GeneratorReference to use clips with plain coloured background in OTIO tests

In the web app, users can use four different compare modes when they are analyzing two versions. Since some of those compare modes were not, or only partially, present in RV, these PR contains all the changes required to add them. Opacity and Compare implementations are based on what is in [Create](https://git.autodesk.com/media-and-entertainment/shotgun-rv/pull/193). Note that Side by Side is entirely in the Live Review plugin in [this PR](https://git.autodesk.com/media-and-entertainment/rv/pull/342). The difference mode will need to be added in a following PR since some investigations need to be done to identify why the web app and RV have such different ways of implementing this mode.

The changes were tested on macOS during a Live Session between RV and the web app on [https://sg-create-prod.shotgunstudio.com/](https://sg-create-prod.shotgunstudio.com/).


src/lib/ip/IPBaseNodes/SequenceIPNode.cpp
```diff
@@ -6,6 +6,8 @@
 //
 //******************************************************************************
 
+#include "IPCore/ShaderCommon.h"
```
bernie-laberge, line 9: Not a big deal but I think it would best if these 2 includes were moved to line 16 and with the form
```
#include <IPCore/GroupIPNode.h>
#include <IPCore/ShaderCommon.h>
#include <TwkMath/Vec2.h>
#include <TwkFB/FrameBuffer.h>
```

src/lib/ip/IPCore/IPCore/ShaderCommon.h
```diff
@@ -7,6 +7,7 @@
 //
 #ifndef __IPCore__ShaderCommon__h__
 #define __IPCore__ShaderCommon__h__
+#include "TwkMath/Vec2.h"
```
bernie-laberge, line 10: Maybe move this include after #include <TwkMath/Mat44.h> below:
```
#include <TwkMath/Mat44.h>
#include <TwkMath/Vec2.h>
```

src/lib/ip/IPCore/ShaderFunction.cpp
```diff
@@ -767,30 +841,38 @@ namespace IPCore
                         Symbol::Type stype = Symbol::VoidType;
                         unsigned int qualifier = Symbol::Uniform;
 
-                        for (const SymbolTypeAssociation* a =
-                                 symbolTypeAssociations;
-                             a->glslName; a++)
+                        for (const auto& association : symbolTypeAssociations)
```
bernie-laberge, line 844: Nice ! I thought that this form was only working with actual containers like std::vector and  std::array. And not with plain vanilla arrays ! Thank you !

bernie-laberge (approved):
LGTM
Really nice addition. Thank you @eloisebrosseau !

src/lib/ip/IPBaseNodes/SequenceIPNode.cpp
eloisebrosseau: Done!

src/lib/ip/IPCore/IPCore/ShaderCommon.h
eloisebrosseau: Done!

src/lib/ip/IPCore/ShaderCommon.cpp
```diff
@@ -5,6 +5,11 @@
 //  SPDX-License-Identifier: Apache-2.0
 //
 //
+#include "IPCore/IPImage.h"
```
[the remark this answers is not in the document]
eloisebrosseau: Done!

Note that I went over all the files, and I think all the includes should follow the project convention now! I'll be more careful next time! Sorry about that!

src/lib/ip/IPCore/ShaderCommon.cpp
```diff
@@ -314,6 +322,9 @@ namespace IPCore
         static Function* Shader_StereoChecker = 0;
         static Function* Shader_StereoAnaglyph = 0;
         static Function* Shader_StereoLumAnaglyph = 0;
+        static Function* Shader_AngularMask = 0;
```
rogernelson, line 325: It's beyond the scope of this changelist, but at some point we should probably change these for `std::unique_ptr<Function>` instead to minimize the chance of leaks, especially in a multithreaded situation.

src/lib/ip/IPCore/IPCore/ShaderProgram.h
```diff
@@ -205,6 +205,7 @@ namespace IPCore
             std::string m_vertexCode;
             bool m_needOutputSize;
             bool m_needOutputST;
+            bool m_needFragmentPosition;
```
rogernelson, line 208: Could initialize here as `bool m_needFragmentPosition{false}` to save having to do it in each ctor.

src/lib/ip/IPCore/IPCore/ShaderFunction.h
```diff
@@ -282,6 +284,7 @@ namespace IPCore
             bool m_inline;
             bool m_usesOutputSize;
             bool m_usesOutputST;
+            bool m_usesFragmentPosition;
```
rogernelson, line 287: Could initialize here as `bool m_usesFragmentPosition{false}` to save having to do it in each ctor.

src/plugins/rv-packages/otio_reader/compare_hook.py
```diff
@@ -0,0 +1,141 @@
...
+
+
+def hook_function(
+    in_timeline: otio.schemadef.Compare.Compare, argument_map: dict | None = None
+) -> None:
+    """A hook for the compare schema"""
+    compare_mode = in_timeline.mode
+    is_swapped = in_timeline.is_swapped
+
+    match (compare_mode):
```
rogernelson, line 13: `compare_mode.lower` so it's case insensitive?

src/plugins/rv-packages/otio_reader/compare_hook.py
```diff
@@ -0,0 +1,141 @@
...
+        case "side_by_side":
+            # Only the bounds need to be handled for this mode, and this is done in
+            # the Live Review plugin, so nothing to do here
+            pass
+
+        case "over_with_opacity":
+            opacity = in_timeline.over_with_opacity["opacity"]
+            if not isinstance(opacity, float):
+                opacity = float(opacity)
```
rogernelson, line 22: You could probably do it unconditionally.  I don't think calling float() on a float would have any adverse effects.

rogernelson (approved):
This is really well done, nice work!

src/lib/ip/IPCore/ShaderCommon.cpp
eloisebrosseau: I totally agree! I felt it would change too much code unrelated to this PR, but I'll try to do a seperate one to address your suggestion when I have some time!

src/lib/ip/IPCore/IPCore/ShaderProgram.h
eloisebrosseau: Done!

src/lib/ip/IPCore/IPCore/ShaderFunction.h
eloisebrosseau: Done!

src/plugins/rv-packages/otio_reader/compare_hook.py
eloisebrosseau: I don't think we should ever get that string in anything else than lower case, but I added your suggestion to handle that case if it ever happen!

PR #748, opened 2025-04-14
SG-38731: Update annotation tool's undo/redo and clear to match Live Review's behaviour

eloisebrosseau:
### Summarize your change.

The keywords start and end were added as delimiters for a list of strokes cleared on the same frame in _clearPaint_. If the keywords are detected in _redoPaint_ all the strokes between the two words will be added to the undo stack and removed from the redo stack. Otherwise, only one stroke is added. Note that the keywords are not added to the undo stack since each stroke should be popped individually once that action is reached.

The current undo/redo implementation in the annotation tool needed to be updated to match the annotation behaviour in the Live Review plugin.

The changes were tested on macOS.

https://github.com/user-attachments/assets/dd8a1c4f-4667-486d-8c57-53a3a6cafce7

### DIFF
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -1205,8 +1205,34 @@ class: AnnotateMinorMode : MinorMode
             let u = getStringProperty(upropName),
                 r = getStringProperty(rpropName);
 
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
Excellent.
Thanks!

cedrik-fuoco-adsk (approved):

PR #749, opened 2025-04-14
Feature/Qt6 - Movie load acceleration

pbergeron-adsk:
### Summarize your change.

Note: This is a test PR for internal Autodesk review that needs to undergo testing.

This PR aims to dramatically improve the loading time of playlists, especially online playlists, which was quite painful for customers, sometimes having to wait several minutes.  For example, on my workstation, loading a playlist of 250 clips went from 3-4 minutes to about 10 seconds, which is pretty nice improvement.

Some years ago, RV introduced "progressive source loading" which aimed to achieve a similar result, but caused lots of unintended support issues, like scrips not working well and some random crashes sometimes, and other side effects. This PR is *not* an attempt to improve progressive loading, which had fundamental flaws, instead we use a different technique altogether.

Several changes were made in order to achieve this, but first, it's worthwhile to describe the cause of the issue to understand the fix.

The reason for the slowness is because after we receive the playlist (the list or URLs) we must create the RV graph, for which we create one FileSourceIPNode per movie, sequentially, in the main thread. For each FileSourceIPNode created sequentially, we must open the movie header (to get the number of frames, framerate, duration, etc) by calling FileSourceIPNode::openMovie(), eventually calling upon libav (ffmpeg), which initiates an http request (with cookies and headers), sends the request, downloads the response and the data, and then can open the movie's header. Unfortunately, doing this over the network, one file at a time, causes RV to block on I/O while the shotgrid server is serving only a single request. To give a rough idea, for a single clip, that network lag is about 0.5-to-1.0 seconds as I measured it, so multiply that by 250 and you've got "several minutes". 

Obviously, the shotgrid server can handle multiple concurrent requests at once, so we simply decided to create a preloader which loads all 250+ file headers in the background, using up to 32 concurrent threads to issue those http requests. In other words, the idea is to parallelize the I/O waiting time, hopefully by a factor of "up to" 32. 

So while the preloader is doing all of this work (well, waiting, mostly) in the background, the RV graph gets created in the main thread, just as before. But, by the time FileSourceIPNode::openMovie gets called -- lo-and-behold! -- the movie's header has already been opened (or so we're hoping) so the call completes almost immediately without having to pay that 0.5-to-1.0 second lag time.

Unfortunately, there's some post-fixup stuff going on after we load a clip, which causes RV to still need to spend about 0.05 second per node, but still, it's a much better outcome than before.

Another performance bug was fixed in this PR: we're no longer paying a O(n^2) (maybe O(n^3)?) loop on trying to determine if the UI name of a clip already exists, by searching for an existing name in the RV graph (which is NOT free), and instead we have an O(1) algorithm to figure out the next UI name for a node in case there's a duplicate.

Other little fixes were done here and there, but the crux of the problem is solved using the above approach, with other details provided herein below.

Loading long playlists was unusably slow; it was time to improve this for customers especially for the purpose of collaborative review.

I tested this on my Mac. When logged in shotgrid, I select a playlist of 250 clips, right-click, and then "play in RV". 

Also, using the Live Review workflow on shotgrid, we join a session in RV, and the clips load way faster than before, although this has made us uncover other issues we need to debug/fix separately.

**NOTE: For QA testing, please check that this continues to work as before even for static files or image sequences, even though we may not benefit much from preloading speed in some cases.**

A few things to mention here.

**(1) New methods to MovieReader's public API**

The class MovieReader's base public interface was modified to add two new methods, and all MovieReader-derived classes must implement two new functions: 

- preloadOpen() -- which gets executed in the background to open the movie file.
- postPreloadOpen() -- which gets executed in the main thread to perform work done in the FileSourceIPNode's current loading context at node creation.

Implementing the above two functions is actually quite easy -- you basically just need to break in two what was previously done in the open() methods. All of this is documented at length in MovieReader.h so if you're interested you may read the comments there.

When we did this for MovieFFMpegReader, it turns out that there's a thread-unsafe method that we would have liked to put in preloadOpen(). Indeed, findStreamInfo() calls libav's  avformat_find_stream_info() which _should_ be thread safe but isn't, and causes crashes when called concurrently. I haven't debugged this (yet?) but it's a long-time bug that was first reported like 15 years ago, but apparently remains, and I provided a link to one of the places I found online that mentions the issue (as a side note, it might have been a cause for random crashes with progressive loading years ago). In any case, the crashes went away when I moved findStreamInfo() to postPreloadOpen() (which is serialized and executed in the main thread) just like before.

**(2) You may not like the way the background thread scheduler works.**

It's perhaps possible that the background thread management and reader scheduler is a bit overly complex, although it's still a relatively simple thing overall. Nevertheless, it might be a worthwhile exercise to use a thread pool with each thread in the pool having a queue, instead of a thread scheduler that dispatches one new thread whenever we have pending files to read.  While a thread pool has the benefit of reusing threads, all threads remain in existence (but paused) after everything is done loading, consuming some memory and resources. On the other hand, my current approach launches/ends one thread per movie, and the thread ends when the movie's done reading, thus releasing the thread memory after the movie's header is read. We also maybe could have used a thread pool library, like the one provided by boost, but, at this time this custom code works well enough as it is, and it's tweakable to try new things, so why not.

**(3) Finding the next UI Name is done is a super simple helper class called UINameCache**

It is found in Session.h, and simply maintains a counter to a root name, and every time "nextUIName" is called, the counter increments.  While this is a super simple solution to an O(N^2-or-3) problem, it does have slightly different behavior than before, which is also documented in the header file.  These differences in behavior were discussed at length with Bernie, and we're ok with the changes.

**(4) There's a function called lookupFilenameInMediaLibrary which has moved.**

This is a function that not just gets the filename in the media library, but also fetches the cookies and headers used to create the http request (otherwise, fetching the file via http will fail due to not having session cookies).

The method used to be in FileSourceIPNode, but it's been moved to a namespace that is accessible by the preloader. It used no special member of FIleSourceIPNode, and mostly static functions at that, so it was easy to move elsewhere; it was mostly copy/pasted from one spot to another.

**(5) You may be wondering why some "if()" statements are now like "if (( ))"**
This is because of an assignment within the if statement. By adding a () around the assignment, the compiler doesn't complain about the if statement and doesn't issue a warning. I got tired of looking at the same verbose warnings constantly.

No screenshot to provide, but, you may imagine in your head a picture of a coffee mug with more coffee in it than before, by the time the playlist is done loading.


src/lib/app/RvApp/CommandsModule.cpp
```diff
@@ -1333,6 +1333,14 @@ namespace Rv
         NODE_RETURN(nodeName);
     }
 
+    NODE_IMPLEMENTATION(startPreloadingMovie, void)
```
bernie-laberge, line 1336: Please document this new command in commands.mud.

src/lib/app/RvApp/RvApp/RvSession.h
```diff
@@ -42,6 +44,7 @@ namespace Rv
 
     class RvSession : public IPCore::Session
     {
+    private:
```
bernie-laberge, line 47: Why is it required ? I'm just curious.

src/lib/app/RvApp/RvSession.cpp
```diff
@@ -496,6 +498,10 @@ namespace Rv
                     string fullname = filename + "/" + "emptyDirectory";
                     inputPatterns.push_back(fullname);
                 }
+                else // not a directory, so it's a movie, add to preloader.
+                {
+                    m_session->startPreloadingMovie(filename);
```
bernie-laberge, line 503: Have you tested with an image sequence ?
Do we want to pre-load image sequences ?

src/lib/app/RvApp/RvSession.cpp
```diff
@@ -581,18 +604,31 @@ namespace Rv
...
 
-            cout << "INFO: " << ((addLayer) ? "+ " : "") << file << endl;
+                m_loadState->nextPattern(pattern, sargs, addLayer);
+                const char* file = pattern.c_str();
+                readSource(file, sargs, true, addLayer, m_loadState->tag(),
+                           m_loadState->merge());
+
+                // cout << "INFO: " << ((addLayer) ? "+ " : "") << file << endl;
+
+                if (onlyOne)
```
bernie-laberge, line 629: Cool that you keep the old way.  I wasn't sure how it would impact the load dynamics if we were to load more than one source at a time here.

src/lib/app/RvApp/RvApp/RvSession.h
```diff
@@ -127,6 +130,11 @@ namespace Rv
 
         bool userHasSetViewSize() { return m_userHasSetViewSize; };
 
+        // Entry point for the media preloader. Used implicitly by addSources
+        // Used explicitly by otio reader
+        // Exposed to users via a Mu & Python command
+        void startPreloadingMovie(const std::string& filename);
```
bernie-laberge, line 136: Philosophical question: Are we just preloading movies ? or are we also going to preload images sequences and or individual image files. 
Depending on the answer, we might want to rename it `startPreloadingMedia()`

src/lib/app/RvApp/RvSession.cpp
pbergeron-adsk: Good question! I did not test this with a sequence of files, or individual image files. To the degree they derive from MovieReader, it should work, since I ported all MovieReader-derived classes, however not all of them will benefit from preloading since some ports basically do nothing in the preloadOpen() portion

src/lib/app/RvApp/RvGraph.cpp
```diff
@@ -287,6 +310,9 @@ namespace Rv
                         + std::string(") ") + nodeName)
                 .c_str());
 
+        //        std::cerr << "Adding source " << nodeName << " / " <<
```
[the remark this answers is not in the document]
pbergeron-adsk: I will remove this, it was for my own debug purposes.

src/lib/app/RvApp/RvSession.cpp
pbergeron-adsk: I kept the old way. It did "improve" the loading somewhat very slightly, but at the subjective perception of not knowing what's going on, so not worth it. The counter "1/2/3/4/n of N" is a useful visual cue of what's going and when to expect it to finish.  But I did leave the option to load everything in one go, for later debugging and testing, if someone wants to re-experiment with this.

src/lib/app/RvApp/RvApp/RvSession.h
pbergeron-adsk: We are preloading anything that derives from MovieReader. Please let me know what you'd like the function name to be.

src/lib/app/TwkMediaLibrary/Library.cpp
```diff
@@ -893,4 +893,108 @@ namespace TwkMediaLibrary
...
+
+                                if (!cookies[c].path.empty())
+                                    cookieStm << "; path=" << cookies[c].path;
+                                if (!cookies[c].domain.empty())
+                                    cookieStm << "; domain="
+                                              << cookies[c].domain;
+                            }
+
+                            /*
+                                                        if
```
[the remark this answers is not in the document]
pbergeron-adsk: I never thought about this at all!  I think we should not print the cookies, but, since RV is open source anyway, what's to prevent someone from creating a local build with printed cookies?  In any case, I dont think I'm the one who commented the cookie printing. I literally just moved the old code from one place to another.

src/lib/app/RvApp/CommandsModule.cpp
pbergeron-adsk: Done.  Also changed name to startPreloadingMedia

src/lib/app/RvApp/RvApp/RvSession.h
pbergeron-adsk: My mistake, removed this useless line.

src/lib/app/mu_rvui/extra_commands.mu
```diff
@@ -826,34 +831,44 @@ Set the user interface name of a node. May modify the name to make it unqiue.
 
 \: setUIName (void; string node, string val)
 {
-    string newName = val;
-    bool dupcheck = true;
+    bool useNextUIName = true;
```
bernie-laberge, line 834: I would get rid of the old implementation and only keep your new improved 
`newName = nextUIName(val);`
This is just for the UI anyway. It has no chance of changing the behaviour.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -1191,10 +1191,27 @@ namespace TwkMovie
...
+        m_request = request;
+
+        if (m_cloning)
+            return;
+
+        if (m_avFormatContext == 0)
+        {
+            openAVFormat(); // sets m_avFormatContext
+            //            findStreamInfo(); // findStreamInfo is not thread
+            //            safe. damnit.
```
bernie-laberge, line 1207: Please remove 'damnit' if you don't mind because it is borderline of a swear word.
:-)

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -4250,7 +4267,8 @@ namespace TwkMovie
             avCodecContext->time_base.den = m_timeScale;
             avCodecContext->codec_id = avCodec->id;
             avCodecContext->codec_type = AVMEDIA_TYPE_VIDEO;
-            avCodecContext->thread_count = m_request.threads;
+            avCodecContext->thread_count =
+                m_request.threads; // never seems to be set anywhere?
```
bernie-laberge, line 4271: Not used but intialized to zero. Note that that this is only for the Writer request:
https://github.com/AcademySoftwareFoundation/OpenRV/blob/b13dc6d5f469658783c4c5da48fb0d804d863638/src/lib/image/TwkMovie/TwkMovie/Movie.h#L267

src/lib/image/MovieMistika/MovieMistika.cpp
```diff
@@ -246,10 +246,18 @@ namespace TwkMovie
         return mov;
     }
 
-    void MovieMistika::open(const string& filename, const MovieInfo& info,
-                            const Movie::ReadRequest& request)
+    void MovieMistika::postPreloadOpen(const MovieInfo& /*unused*/,
+                                       const Movie::ReadRequest& /*unused*/)
+    {
+        // all the work was done in open() since info andrequest were unused.
```
bernie-laberge, line 252: Detail: typo. space missing 'andrequest'

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -48,6 +48,399 @@ namespace TwkMovie
...
+                setStatus(Reader::Status::LOADERROR);
+        }
+        catch (...)
+        {
+            setStatus(Reader::Status::LOADERROR);
+        }
+    }
+
+    GenericIO::Preloader::Preloader()
+        : m_maxThreads(32)
```
bernie-laberge, line 129: Should we put it in a constant at the beginning of the file ?
like `static const int PRELOADER_MAX_THREADS = 32;`

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -48,6 +48,399 @@ namespace TwkMovie
...
+    void GenericIO::Preloader::finalizeCompletedThreadsAndAdjustMaxThreads()
+    {
+        bool update = false;
+
+        // Finalize completed threads.
+        for (auto& reader : m_readers)
+        {
+            if (reader->isPriority())
+            {
+                /* Bump up max threads up to some higher upper limit?
```
bernie-laberge, line 179: I would remove this commented code.

src/lib/image/TwkMovie/TwkMovie/MovieIO.h
```diff
@@ -318,6 +330,99 @@ namespace TwkMovie
 
     class TWKMOVIE_EXPORT GenericIO
     {
+    public:
+        class TWKMOVIE_EXPORT Preloader
```
bernie-laberge, line 334: Should we move the Preloader code into its own files Preloader.h, and Preloader.cpp ?

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -48,6 +48,399 @@ namespace TwkMovie
...
+
+    MovieReader* GenericIO::Preloader::getReader(std::string_view filename,
+                                                 const MovieInfo& info,
+                                                 Movie::ReadRequest& request)
+    {
+        m_getCount++;
+
+        //        std::cerr << "PRELOADER GET " << filename << std::endl;
+
+        // There's no need to lock the scheduler mutex when accessing m_readers
```
bernie-laberge, line 274: I would lock to be safe. Because even if you lock in addReader(), it means nothing if you are not locking here to get access to the m_readers container that might have changed under your feet.
Perhaps a read-write-lock could be used to allow multiple reads and a single write:
**Read access to m_readers:**
    std::shared_lock<std::shared_mutex> lock(m_schedulerThread_mutex);
**Write access to m_readers:**
    std:: unique_lock <std::shared_mutex> lock(m_schedulerThread_mutex);

src/plugins/rv-packages/otio_reader/otio_reader.py
```diff
@@ -67,12 +67,20 @@ def read_otio_string(otio_string: str, host_prefix: str | None = None) -> object
...
+
+    print(">>> READ_OTIO_STRING called. host_prefix = {}".format(host_prefix))
+
     otio_obj = otio.adapters.read_from_string(otio_string)
     timeline = otio_obj["otio"]
-
     context = {"sg_url": host_prefix} if host_prefix else None
 
-    return create_rv_node_from_otio(timeline, context), timeline.global_start_time
+    print(context)
```
bernie-laberge, line 77: This should be commented out right ?

src/plugins/rv-packages/otio_reader/otio_reader.py
```diff
@@ -83,9 +91,16 @@ def read_otio_file(otio_file):
     Returns the top level node created that represents this otio
     timeline.
     """
+    print(">>> READ_OTIO_FILE called")
```
bernie-laberge, line 94: This should be commented out right ?

src/plugins/rv-packages/otio_reader/otio_reader.py
```diff
@@ -298,13 +313,21 @@ def _create_track(in_seq, context=None):
...
-def _get_global_transform(tl) -> dict:
+def _get_global_transform(tl, context) -> dict:
     # since there's no global scale in otio, calculate the minimum box size
     # that can contain all clips
     def find_display_bounds(tl):
         display_bounds = None
         for clip in tl.find_clips():
             try:
+                # get active clip and start prelodaing the movie in the background
+                if isinstance(clip.media_reference, otio.schema.ExternalReference):
```
bernie-laberge, line 324: How are you getting the active clip ?
Is it possible that you are pre-loading all the possible media representations here ?
The active media reference is specified in active_media_reference_key.
Like I wonder if you load both original_media and mp4 media representations here. But I might be wrong.

bernie-laberge (commented):
This is pretty amazing work considering that it was done in such a short period of time !
Respect (Je lève mon chapeau)
Thank you Patrick!

The only 2 things that we need to figure out I think are:
1. To lock or not to lock to access the m_readers in GenericIO::Preloader::getReader()
2. To make sure that we only preload the active media in otio_reader.py

src/lib/image/TwkMovie/MovieIO.cpp
eloisebrosseau: You should use `constexpr` instead of `const` since this can be evaluated at compile time. Also, usually all caps variables are for macros, not constants. It should simply be `preloaderMaxThreads` like any other variable names.

src/lib/image/TwkMovie/TwkMovie/MovieIO.h
```diff
@@ -23,6 +29,12 @@
 #define strcasecmp _stricmp
 #endif
 
+// my daughter dared me to leave this comment here saying
+// the code reviewers will never spot it.
+//
+// "The unhappy potato wuz here!!! I will take revenge for
+//  my falled bretherens that were slayed into fries"
+
```
eloisebrosseau, line 37: I found it ;)

src/plugins/rv-packages/otio_reader/otio_reader.py
```diff
@@ -67,12 +67,20 @@ def read_otio_string(otio_string: str, host_prefix: str | None = None) -> object
...
     timeline = otio_obj["otio"]
-
     context = {"sg_url": host_prefix} if host_prefix else None
 
-    return create_rv_node_from_otio(timeline, context), timeline.global_start_time
+    print(context)
+
+    commands.addSourceBegin()
+    ret = create_rv_node_from_otio(timeline, context), timeline.global_start_time
+    commands.addSourceEnd()
```
eloisebrosseau, line 81: FYI: I removed from my latest Pr that was just merged `read_from_string` from `otio_reader.py` and moved the logic in `live_review.py` since it was too specific to Live Review. `commands.addSourceBegin()` and `commands.addSourceEnd()` should be added [here](https://git.autodesk.com/media-and-entertainment/rv/blob/3716f767a411fbe8526bc49380c4ea1435e4eda1/src/plugins/packages/live_review/src/live_review.py#L1033) instead.

src/lib/app/mu_rvui/extra_commands.mu
pbergeron-adsk: I thought about this -- and I do want to keep it there. The reason is that if this turns out to generate some kind of trouble due to an unexpected use case, people won't necessarily know to look for the history. I'd leave it for the current release if that's ok.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
pbergeron-adsk: ok.

src/lib/image/TwkMovie/TwkMovie/MovieIO.h
pbergeron-adsk: Well, it's an internal class to GenericIO, so it kinda needs to be there unfortunately :(

pbergeron-adsk:
> This is pretty amazing work considering that it was done in such a short period of time ! Respect (Je lève mon chapeau) Thank you Patrick!
> 
> The only 2 things that we need to figure out I think are:
> 
> 1. To lock or not to lock to access the m_readers in GenericIO::Preloader::getReader()
> 2. To make sure that we only preload the active media in otio_reader.py

Thank you Bernard!  Concerns (1) and (2) have been addressed.

bernie-laberge (approved):
Beautiful. I can't wait to try it.
Thank you Patrick!

eloisebrosseau (approved):

pbergeron-adsk:
Done and done.

PR #752, opened 2025-04-16
Feature/Qt6 - Small tweak of the audio logic

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
I removed the conditions to check the state of the QAudioSink object because it does not make sense in Qt 6. In Qt 6, when we call the `start `method on the QAudioSink object, The code starts transferring audio data from the device to the system's audio output right away and changes the device state afterwards (while in Qt 5, the state was changed first, and then the audio data was read). Therefore, it does not make sense to check the state in the `readData` callback because even if the state is STOPPED, we want to start reading the audio data.

The audio cache/processing seems to behave just like the main branch with those changes. I tested it on Windows and MacOS.

Fix audio on Windows and the new logic makes more sense based on how Qt 6 works now.

Windows and MacOS

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -248,7 +248,7 @@ jobs:
           tools: 'tools_qtcreator'
           set-env: 'true'
           tools-only: 'false'
-          aqtversion: '==3.1.*'
+          aqtversion: '==3.2.*'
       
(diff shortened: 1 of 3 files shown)


bernie-laberge (approved):
LGTM

eloisebrosseau (approved):

PR #764, opened 2025-04-29
Qt6: Disable stereo shutter glass support 

pbergeron-adsk:
### Summarize your change.

Hardware Stereo Shutter mode gone since Qt6/QOpenGLWidget. 
(eg: nvidia glasses circa 2011)

There is apparently no customer demand for it, and review of stereo footage is apparently done using other methods regardless. We're leaving existing code in case customers want to revive it.

Tested on macOS,  simply commented out the few lines of code required to make this option available in the UI. Most of the code to enable the feature is found in DisplayStereoIPNode, which we have not touched.

n/a


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #781, opened 2025-05-16
Add hold and ghost

eloisebrosseau:
### Summarize your change.

This PR adds hold and ghost in the PaintIPNode based on the implementation in Create and the cherry-pick in [shotgun-rv](https://git.autodesk.com/media-and-entertainment/shotgun-rv/pull/1262) that was never merged. The following properties were added to enabled the feature in RV: `startFrame`, `duration`, `hold`, `ghost`, `ghostBefore` and `ghostAfter`. The Open RV documentation reflects this addition in the Paint node.

Hold and ghost can be enable on both polylines and text boxes. To hold annotations, a `duration` property was needed to know the number of frames on which the annotation should be held from the specified `startFrame`. To ghost annotations, a `ghostBefore` and `ghostAfter` properties were added to set the number of frames on which the ghosted annotation should be displayed before and after the actual annotation. The opacity to apply on ghosted annotations was also updated to match the web app instead of Create to make the annotations more visible the closer they are to the actual annotation, and the more transparent the further they are. Note that some code that was part of the cherry-pick was to fix some thread safeness issues, without being directly related to the hold and ghost feature, but more generally for the PaintIPNode. This code is also part of this PR to make sure everything is on par with what should have been merged back then, but I can remove it, if it makes this PR too hard to review.

Since this was added as a feature for Live Review, the annotation schema, the annotation hook and the otio reader are also updated in this PR. The `hold` and `ghost` properties are received for now as part of the annotation metadata. To access the values in the annotation hook, the context in the otio reader needed to be added the metadata propertyof OTIO effects. In the annotation hook, all the new properties to enable hold and ghost were added. The `duration` property is calculated so that a held annotation will stay on the frames until the next annotated frame or the end of the media. For ghosted annotations and "normal" annotations, the duration is simply the specified OTIO time range, which is normally 1, but it could be any other values.

To see the complete implementation of hold and ghost in the Live Review plugin, and some screen recordings, please refer to this other [PR](https://git.autodesk.com/media-and-entertainment/rv/pull/362). Also note that a playback issue is currently present when enabling `ghost` where the first frame without any annotation will completely freeze until the next annotated frame. However, the ghost feature is working as expected when scrubbing on the timeline. Held annotations and "normal" annotations are not affected by this issue.

RV is missing the hold and ghost feature that is needed for Live Review.

All the test were done on macos between RV and the web app during a Live Review session.

### DIFF
--- a/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
+++ b/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
@@ -1,17 +1,175 @@
 //
-//  Copyright (c) 2009 Tweak Software.
-//  All rights reserved.
+// Copyright (C) 2025 Autodesk, Inc. All Rights Reserved.
 //
-//  SPDX-License-Identifier: Apache-2.0
-//
+// SPDX-License-Identifier: Apache-2.0
 //
 #include <IPBaseNodes/PaintIPNode.h>
 #include <IPCore/Exception.h>
 #include <IPCore/IPGraph.h>
 #include <TwkGLText/TwkGLText.h>
+#include <mutex>
 #include <stl_ext/string_algo.h>
 #include <cstdlib>
 #include <IPCore/ShaderCommon.h>
+
+namespace
+{
+    using namespace IPCore;
+
+    float getGhostFactor(const int level, const int maxLevel)
+    {
+        constexpr float maxFactor = 0.5;
+        constexpr float minFactor = 0.1;
+        constexpr float factorRange = maxFactor - minFactor;
+        float visibility = 1.0;
+
+        if (level > 1 && maxLevel > 1)
+        {
+            visibility = 1.0f
+                         - (static_cast<float>(level - 1)
+                            / static_cast<float>(maxLevel - 1));
+        }
+
+        return minFactor + (visibility * factorRange);
+    }
+
+    PaintIPNode::LocalCommands
+    generateVisibleCommands(const PaintIPNode::LocalCommands& commands,
+                            const int frame, const int eye)
+    {
+        PaintIPNode::LocalCommands
+            allCommands; // visible commands, including hold and ghost
+        PaintIPNode::LocalCommands
+            CurrentFrameCommands; // visible commands, excluding hold and ghost
+
+        using PerFramePaintCommands = std::map<int, PaintIPNode::LocalCommands>;
+
+        PerFramePaintCommands beforeCommands;
+        PerFramePaintCommands afterCommands;
+
+        for (auto* localCommand : commands)
+        {
+            if (localCommand->eye != 2 && localCommand->eye != eye)
+            {
+                continue;
+            }
+
+            // Don't add polyLines with 0 points
+            if (auto* localPolyLine =
+                    dynamic_cast<PaintIPNode::LocalPolyLine*>(localCommand))
+            {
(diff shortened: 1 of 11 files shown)


src/lib/ip/IPBaseNodes/PaintIPNode.cpp
```diff
@@ -1,17 +1,212 @@
...
+                                // (ghostBefore)
+        {
+            ghostOpacity = static_cast<float>(duration)
+                               / static_cast<float>(frame - startFrame)
+                           + minOpacity;
+        }
+        else // Command starts after the current frame (ghostAfter)
+        {
+            ghostOpacity = static_cast<float>(duration)
+                               / static_cast<float>(startFrame - frame)
```
bernie-laberge, line 41: There is a possible division by zero error here I think if startFrame==frame.

src/lib/ip/IPBaseNodes/PaintIPNode.cpp
```diff
@@ -1,17 +1,212 @@
...
+    auto
+    separateCommandsByFrameGroup(const PaintIPNode::LocalCommands& commands,
+                                 const int frame, const size_t eye)
+    {
+        PaintIPNode::LocalCommands
+            currentFrameCommands; // visible commands, excluding hold and ghost
+        PerFramePaintCommands beforeCommands;
+        PerFramePaintCommands afterCommands;
+
+        for (auto* localCommand : commands)
```
bernie-laberge, line 57: Perhaps it would be helpful to the read to add the original comment here ? :
 ```
     // In this phase, we loop over all commands and we separate them in 3
      // containers.
      // 1. Commands that are visible on the current frame (based only on their
      //    visibility range settings)
      // 2. Commands that end *before* the current frame.
      // 3. Commands that start *after* the current frame.
      //
```

File: src/lib/ip/IPBaseNodes/PaintIPNode.cpp [hunk 3]
```diff
@@ -1,17 +1,212 @@
...
+
+    auto
+    separateCommandsByFrameGroup(const PaintIPNode::LocalCommands& commands,
+                                 const int frame, const size_t eye)
+    {
+        PaintIPNode::LocalCommands
+            currentFrameCommands; // visible commands, excluding hold and ghost
+        PerFramePaintCommands beforeCommands;
+        PerFramePaintCommands afterCommands;
+
```
bernie-laberge, line 56: Perhaps it would be helpful to the read to add the original comment here ? :
 ```
     // In this phase, we loop over all commands and we separate them in 3
      // containers.
      // 1. Commands that are visible on the current frame (based only on their
      //    visibility range settings)
      // 2. Commands that end *before* the current frame.
      // 3. Commands that start *after* the current frame.
      //
```

bernie-laberge (approved):
LGTM
Excellent work @eloisebrosseau and thank you for your constant effort of modernizing the code along the way !

src/lib/ip/IPBaseNodes/PaintIPNode.cpp
eloisebrosseau: The ghost opacity is always calculated for a frame that is either greater than or less than the current frame. I replaced the else condition with `frame < startFrame`

cedrik-fuoco-adsk (approved):

PR #786, opened 2025-05-21
Incorrect EDL in multi-fps playlists

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.

The calculation of the frame and out properties in the EDL was incorrect for playlists containing clips with varying frame rates. 

In `_get_in_out_frame`, the values for the `in` and `out` frames were not what RV is expecting. For all playlists, the in values must be the `start_time` of the clip at its rate - not the rate of the sequence/timeline (_which is the rate of the first clip_). 

For the out value:

- In multi-fps playlists, it should be [start_time](vscode-file://vscode-app/Applications/Visual%20Studio%20Code.app/Contents/Resources/app/out/vs/code/electron-sandbox/workbench/workbench.html) at the clip’s rate plus the duration at the sequence/timeline rate.
- In single-fps playlists, it should be the end time (inclusive) at the clip’s rate (no conversion needed).

For the frame value in the EDL, I based the implementation on what RV is doing here: https://github.com/AcademySoftwareFoundation/OpenRV/blob/673f2b7d37c222ab0400105163a466131102ab1f/src/lib/ip/IPBaseNodes/SequenceIPNode.cpp#L785-L800

The current implementation of the EDL calculations for the `frames` and `out` were wrong. The values were not what RV was expecting. The original issue was that the clip were stuck at the last frame of the clip because the EDL values are wrong.

**Note:** 
There are still issues with multi-fps playlists, but these will be tackle in subsequent PRs.

MacOS, Linux


pbergeron-adsk (changes requested):
Ok, but can you specify in your pr comment what are the other issues you intend to fix in other PRs?

cedrik-fuoco-adsk:
It's one specific issue. There are some cases in a multi-fps playlists where RV and CR does not interpret a frame calculation the same way. For example, RV will do a calculation similar to a floor() while CR will be a ceil(). There are no fix until we decide which one is right. @bernie-laberge and me decided to deal with that later.

bernie-laberge (approved):
LGTM
Please add the repro steps of SG-38200 for the post checkin test just to be safe.

PR #787, opened 2025-05-22
Update gc.cmake to work with non-RHEL distros

kenmcgaugh:
Unlike all the other dependency cmake files, this one was using RV_TARGET_LINUX rather than RHEL_VERBOSE to determine if the lib directory should be "lib64". This caused issues building under non-RHEL distros like ubuntu.


cedrik-fuoco-adsk (approved):
Good catch! Thank you!

PR #794, opened 2025-05-30
Fix crash when the pixel format is unknown in frames

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Save the pixel format when the media is opened and use that as fallback if the pixel format of some frames can not be detected. This assumes that there is a pixel format when opening the file and if it does not have one, it will fail at that point before getting to `decodeImageAtFrame`.

OpenRV was crashing on some AV1 media because the pixel format was unknown for all the frames.

MacOS


bernie-laberge (approved):
LGTM
That will make it more resilient. Thanks!

eloisebrosseau (approved):

PR #795, opened 2025-06-02
Add High DPI Display support to RV

bernie-laberge:
### Linked issues
NA

Taking advantage of high DPI monitor capabilities for nicer RV UI and better media resolution.

### Summarize your change.
This PR is leveraging Qt's high DPI capability in RV:
https://doc.qt.io/qt-6/highdpi.html

It is pretty straightforward: a high DPI scaling factor is available on a per screen basis via the QScreen::devicePixelRatio().
On non high DPI monitors or display resolution, this devicePixelRatio is 1.0f.
On high DPI monitors and compatible resolutions, QScreen::devicePixelRatio() returns a high DPI scaling factor that needs to be applied both in X and Y.

Note that this is on a per screen basis because a computer system could have multiple monitors of various high DPI capabilities and resolutions.

In this commit, the float GLView::devicePixelRatio() method was added which effectively returns the QScreen::devicePixelRatio() for the Qt screen where the GLView is currently being rendered.
A new devicePixelRatio() RV command was also added so that the different UI elements in RV can be adapted to take this new high DPI scaling into account.

With this commit, we are now supporting high DPI displays by default
Setting the following environment variable, effectively disable the high DPI support in RV and restore the previous non high DPI capable functionality:
`export RV_NO_QT_HDPI_SUPPORT=1`

Successfully tested on macOS with 4K external monitor set in 4K resolution to validate transitioning from a high DPI monitor (Retina Display) to a non high DPI monitor.

better_ui_and_media_resolution_with_high_dpi_support_in_rv

### DIFF
--- a/src/lib/app/RvCommon/QTGLVideoDevice.cpp
+++ b/src/lib/app/RvCommon/QTGLVideoDevice.cpp
@@ -5,8 +5,7 @@
 //  SPDX-License-Identifier: Apache-2.0
 //
 //
-// #include <QtGui/QtGui>
-// #include <QtOpenGL/QtOpenGL>
+
 #ifdef PLATFORM_WINDOWS
 #include <GL/glew.h>
 #include <GL/wglew.h>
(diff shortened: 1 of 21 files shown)


src/lib/app/RvCommon/QTGLVideoDevice.cpp
cedrik-fuoco-adsk, line 1: Looks like it is missing an include for QScreen (VFX2023 is complaining, but not VFX2024 builds)

bernie-laberge: OK I just pushed the change.

PR #801, opened 2025-06-12
SG-38113: Add stand-alone Clear All Frames

eloisebrosseau:
### Summarize your change.

Update the Annotation package to be able to clear all annotations on the timeline. To handle this new feature, a dropdown menu was added to the Clear button in the Draw panel to differentiate between clearing annotations on the current frame and clearing all annotations on the timeline. The Annotations menu was also updated to display both option. To prevent users from clearing all the annotations on the timeline by accident, an alert panel was added to confirm the user's intention.

This feature is required if we wish RV to be functionally on par with Creative Review Web with respect to Annotation related features.

This feature was tested on macOS 15.5.

https://github.com/user-attachments/assets/d1c7aaad-92c5-4b93-a696-368f2f976dbe

### DIFF
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -1619,7 +1619,7 @@ class: AnnotateMinorMode : MinorMode
 
     method: clearSlot (void; bool checked)
     {
-        clearPaint(_currentNode, sourceFrame(frame()));
+        clearPaint(_currentNode, _currentNodeInfo.frame);
         updateFrameDependentState();
         redraw();
     }
@@ -1639,6 +1639,7 @@ class: AnnotateMinorMode : MinorMode
                 let annotatedFrames = findAnnotatedFrames(node);
                 for_each(frame; annotatedFrames)
                 {
+                    clearPaint(node, frame);
                     clearPaint(node, sourceFrame(frame));
                 }
             }


src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1247,30 +1247,38 @@ class: AnnotateMinorMode : MinorMode
...
 
-    method: clearPaint (void; int frame)
+    method: clearPaint (void; int[] frames)
     {
-        let f = _currentNodeInfo.frame,
-            upropName = frameOrderName(_currentNode, f),
-            rpropName = frameOrderRedoStackName(_currentNode, f);
-
-        if (propertyExists(upropName))
+        for_each(node; nodes())
```
bernie-laberge: This is my only concern: if we no longer take _currentNode into account, then does it mean that if you have multiple sources, then this clearPaint() command will clear the annotations on every single source ?
(I tried in RV 2024.2.0, and if I load two clips, then select each clip individually and draw an annotation on the same frame, then the CleatFrame will only clear the annotations of the current source)

eloisebrosseau: You're right. I fix the code so that `Clear Frame` will call `clearPaint()` with the current node like it used to do. Let me know if you're fine with my fix!

bernie-laberge: All good now. Thanks !

cedrik-fuoco-adsk (approved):

pbergeron-adsk:
Looks cool! I approved.

pbergeron-adsk (approved):

PR #806, opened 2025-06-17
HTJ2K Support

richardssam:
### Pull Request HTJ2K Support

Adding HTJ2K support via a separate reader for J2C files, and also adding support for the MovieFFMpeg module, so that if it sees a HTJ2K stream embedded in a mxf or mov file, it will use the OpenJPH reader rather than the vanilla ffmpeg reader, which is quite a bit slower.

HTJ2K is a high quality extension of the Jpeg2000 codec, which is itself significantly faster than the original Jpeg2000 spec. OpenJPH adds native CPU acceleration on Intel platforms, and even without it, the performance improvements allow for 4K playback at > 60 FPS on Mac ARM hardware, at higher quality than alternative compressed codecs.

This does also support floating point htj2k support, although I think that will more typically be used within the new OpenEXR library which now has HTJ2K as a lossless compressed option.

While HTJ2K was supported previously via ffmpeg and OpenJpeg, the performance was not great. This adds the OpenJPH library to OpenRV (downloading it if necessary), which has a significant improvement in performance.

I have tested significantly on OSX, and windows.

I would definately look at the changes to the ffmpeg library, I've tried to minimize the impact


richardssam:
THis was a bad MR, it was cleaned up and resubmitted.

PR #807, opened 2025-06-17
SG-39280: Update UI on Hold and Ghost in RV session

eloisebrosseau:
### Summarize your change.

Update the state of the Hold and Ghost button in the bottom tool bar to match the annotations state in the RV session file loaded. If the annotations in the file are held or ghosted, the corresponding button will be checked.

When loading a RV session file, the Hold and Ghost button UI stated was not updated. If the file had annotations with held or ghosted annotations, the corresponding button was still dark grey instead of blue.

The fix was tested on macOS 15.5.


bernie-laberge (approved):
LGTM
On top of that this is all Annotation package agnostic ! Nice.

cedrik-fuoco-adsk (approved):

PR #808, opened 2025-06-17
HTJ2K Patches for FFMpeg and standalone IO mode.

richardssam:
Pull Request HTJ2K Support
Adding HTJ2K support via a separate reader for J2C files, and also adding support for the MovieFFMpeg module, so that if it sees a HTJ2K stream embedded in a mxf or mov file, it will use the OpenJPH reader rather than the vanilla ffmpeg reader, which is quite a bit slower.

HTJ2K is a high quality extension of the Jpeg2000 codec, which is itself significantly faster than the original Jpeg2000 spec. OpenJPH adds native CPU acceleration on Intel platforms, and even without it, the performance improvements allow for 4K playback at > 60 FPS on Mac ARM hardware, at higher quality than alternative compressed codecs.

This does also support floating point htj2k support, although I think that will more typically be used within the new OpenEXR library which now has HTJ2K as a lossless compressed option.

Describe the reason for the change.
While HTJ2K was supported previously via ffmpeg and OpenJpeg, the performance was not great. This adds the OpenJPH library to OpenRV (downloading it if necessary), which has a significant improvement in performance.

Describe what you have tested and on which operating system.
I have tested significantly on OSX, and windows.

Add a list of changes, and note any that might need special attention during the review.
I would definately look at the changes to the ffmpeg library, I've tried to minimize the impact

### DIFF
--- a/src/lib/app/ImGuiPythonBridge/ImGuiPythonBridge.cpp
+++ b/src/lib/app/ImGuiPythonBridge/ImGuiPythonBridge.cpp
@@ -7,8 +7,10 @@ namespace Rv
 {
     void ImGuiPythonBridge::PyObjectDeleter::operator()(PyObject* obj) const
     {
-        if (obj)
-            Py_DECREF(obj);
+        // seems to crash on exit if we have a callback registered.
+        // so patching this out for now.
+        // if (obj)
+        //     Py_DECREF(obj);
     }
 
     std::vector<ImGuiPythonBridge::PyObjectPtr> ImGuiPythonBridge::s_callbacks;
@@ -54,4 +56,9 @@ namespace Rv
 
     int ImGuiPythonBridge::nbCallbacks() { return (int)s_callbacks.size(); }
 
+    void ImGuiPythonBridge::clearCallbacks()
+    {
+        s_callbacks.clear();
+        // CAll DECREF for each element if we're not using the deleter for this?
+    }
 } // namespace Rv
--- a/src/lib/app/ImGuiPythonBridge/ImGuiPythonBridge.h
+++ b/src/lib/app/ImGuiPythonBridge/ImGuiPythonBridge.h
@@ -18,6 +18,7 @@ namespace Rv
         static void unregisterCallback(PyObject* callable);
         static void callCallbacks();
         static int nbCallbacks();
+        static void clearCallbacks();
 
         struct PyObjectDeleter
         {
--- a/src/plugins/rv-packages/otio_reader/annotation_hook.py
+++ b/src/plugins/rv-packages/otio_reader/annotation_hook.py
@@ -80,7 +80,9 @@ def hook_function(
                     "Unable to get the global scale, using the aspect ratio of the first media file"
                 )
                 try:
-                    first_source_node = commands.sourcesAtFrame(0)[0]
+                    media_switch = commands.nodeConnections("MediaTrack")[0][0]
+                    media_source_group = commands.nodeConnections(media_switch)[0][0]
+                    first_source_node = extra_commands.nodesInGroupOfType(media_source_group, "RVFileSource")[0]
                     media_info = commands.sourceMediaInfo(first_source_node)
                     height = media_info["height"]
                     aspect_ratio = media_info["width"] / height
--- a/src/plugins/rv-packages/otio_reader/compare_hook.py
+++ b/src/plugins/rv-packages/otio_reader/compare_hook.py
@@ -50,7 +50,9 @@ def hook_function(
             pivot_x = pivot.x
             pivot_y = pivot.y
 
-            source = argument_map["src"]
+            media_switch = commands.nodeConnections("MediaTrack")[0][0]
+            media_source_group = commands.nodeConnections(media_switch)[0][0]
+            source = extra_commands.nodesInGroupOfType(media_source_group, "RVFileSource")[0]
             transform = extra_commands.associatedNode("RVTransform2D", source)
             if commands.getIntProperty(f"{transform}.transform.active")[0] != 0:
                 global_translate_vec = otio.schema.V2d(0.0, 0.0)


src/lib/image/IOhtj2k/IOhtj2k.cpp
```diff
@@ -0,0 +1,385 @@
...
+                   const ReadRequest& request) const
+{
+    ojph::ui32 skipped_res_for_read = 0;
+    ojph::ui32 skipped_res_for_recon = 0;
+    ojph::j2c_infile j2c_file;
+    j2c_file.open(filename.c_str());
+
+    decodeHTJ2K(&j2c_file, &fb);
+    return;
+#if 0
```
pbergeron-adsk: Please remove dead code.

src/lib/image/IOhtj2k/CMakeLists.txt
```diff
@@ -0,0 +1,34 @@
+#
+# Copyright (C) 2022  Autodesk, Inc. All Rights Reserved.
```
pbergeron-adsk: Please fix year

cmake/dependencies/openjph.cmake
```diff
@@ -0,0 +1,116 @@
+#
+#
+# SPDX-License-Identifier: Apache-2.0
+#
```
pbergeron-adsk, line 4: Please fix header

```
#
# Copyright (C) 2025  Autodesk, Inc. All Rights Reserved.
#
#
```

src/lib/image/IOhtj2k/IOhtj2k.cpp
```diff
@@ -0,0 +1,385 @@
+//
+//  Copyright (c) 2009 Tweak Software. 
+//  All rights reserved.
+//  
+//  SPDX-License-Identifier: Apache-2.0
+//
```
pbergeron-adsk, line 6: Please fix header

```
#
# Copyright (C) 2025  Autodesk, Inc. All Rights Reserved.
#
#
```

src/lib/image/IOhtj2k/IOhtj2k/IOhtj2k.h
```diff
@@ -0,0 +1,48 @@
+//
+//  Copyright (c) 2009 Tweak Software. 
+//  All rights reserved.
+//  
+//  SPDX-License-Identifier: Apache-2.0
+//
```
pbergeron-adsk, line 6: Please fix header

```
#
# Copyright (C) 2025  Autodesk, Inc. All Rights Reserved.
#
#
```

src/lib/image/io_htj2k/CMakeLists.txt
```diff
@@ -0,0 +1,32 @@
+#
+# Copyright (C) 2022  Autodesk, Inc. All Rights Reserved.
```
pbergeron-adsk: Fix year

src/lib/image/io_htj2k/init.cpp
```diff
@@ -0,0 +1,34 @@
+//
+// Copyright (C) 2023  Autodesk, Inc. All Rights Reserved.
```
pbergeron-adsk: Fix year

cmake/dependencies/openjph.cmake
bernie-laberge: I don't we need to have Autodesk to be listed when it is an external contribution from the community.
Please correct me if I am wrong @compaga

src/lib/image/IOhtj2k/IOhtj2k.cpp
richardssam: Done.

bernie-laberge (approved):
Fantastic !
Thank you so much @richardssam for your incredible contribution !

richardssam:
To answer your question about ffmpeg, yes, getting the speedups that are in OpenJPH implemented in both OpenJPEG and/or natively in ffmpeg is certainly a possible solution, but certainly harder to implement. However, I think my main goal is progressive loading, which I dont think is going to be achieved via ffmpeg, since it will require a deeper integration with the player.

You mention opaque code, do let me know which areas you are thinking of? The ffmpeg code itself was pretty hard to dig through, but I'm certainly happy to add comments for the code I feel confident about.

I also reverted a change that was causing some of the earlier CI-runners to fail, so they are all running but one, which oddly is failing on a numpy install, which I dont think is related to my code.

pbergeron-adsk:
> You mention opaque code, do let me know which areas you are thinking of?

When I looked at the code originally, I only had a few moments to myself and I mostly glanced over it to get a sense of the type of change that was done, and as I wrote, I mostly commented on headers stuff until I had time to look it over once again. My comment was mostly about the low-level buffer copying code in decodeHTJ2K() etc, which I've now had time to look at properly, and it isn't actually opaque...

But speaking of which, I'd like to request a change there... 

As far as I can tell, the code handing the case for planar vs non-planar appears to be identical, differentiated only by the order of the outer and inner for-loops when copying scanline data.

Would you be kind enough to refactor this part such that there's only 1 body for both cases? I think the easiest thing would be for you to isolate the body and put it in a function that you call in both cases, instead of duplicating that code twice. 

eg:

```
if (is_planar) {
    for (a = 0; a < n; a++)
         for (b = 0; b < m; b++)
              copy_scanline_data(...)
} else  {
    for (b = 0; b < m; b++)
         for (a = 0; a < n; a++)
              copy_scanline_data(...)
}

```

Finally, shouldn't that byte_offset variable be named bit_offset?   

(and, yes: if I was as picky at home as I'm here, I'd no longer have a wife)

pbergeron-adsk:
> I also reverted a change that was causing some of the earlier CI-runners to fail, so they are all running but one, which oddly is failing on a numpy install, which I dont think is related to my code.

There's indeed an issue related to numpy we're looking into. Not related to you.

richardssam:
Addressed the byte-vs.bit variable name, and split out the copy_scanline code into its own function.

pbergeron-adsk:
> Addressed the byte-vs.bit variable name, and split out the copy_scanline code into its own function.

Thank you! I wanted to approve the changes, and my brain is probably not braining properly, but I thought I found some problematic areas as per my notes below. Can you have a look? Or maybe explain what I'm missing?

src/lib/image/IOhtj2k/IOhtj2k.cpp
```diff
@@ -0,0 +1,278 @@
...
+            bit_offset = 4;
+        case 16:
+            if (has_nlt && is_signed)
+                dtype = FrameBuffer::HALF;
+            else
+                dtype    = FrameBuffer::USHORT;
+            break;
+        case 32:
+            if (has_nlt && is_signed)
+                dtype = FrameBuffer::DOUBLE;
```
pbergeron-adsk: Humm... Double? A double is 64 bits. Did you mean the type to be FLOAT? or an INT? (signed 32-bit?)

src/lib/image/IOhtj2k/IOhtj2k.cpp
```diff
@@ -0,0 +1,278 @@
...
+    ojph::ui32 component, 
+    FrameBuffer::DataType dtype,
+    int bit_offset,
+    FrameBuffer *fb){
+
+    ojph::ui32 comp_num;
+    ojph::line_buf *line = codestream->pull(comp_num);
+    const ojph::si32* sp = line->i32;
+    assert(comp_num == component);
+    if (dtype == FrameBuffer::UCHAR){
```
pbergeron-adsk, line 133: Here, you seem to be handling dtype (destination type?) to be UCHAR, HALF or USHORT but you're not handling the DOUBLE/UINT cases? (or is it float, and/or int cases?)...

src/lib/image/IOhtj2k/IOhtj2k.cpp
```diff
@@ -0,0 +1,278 @@
...
+            break;
+        case 16:
+            if (has_nlt && is_signed)
+                fbi.dataType = FrameBuffer::HALF;
+            else
+                fbi.dataType = FrameBuffer::USHORT;
+            break;
+        case 32:
+            if (has_nlt && is_signed)
+                fbi.dataType = FrameBuffer::DOUBLE;
```
pbergeron-adsk: as per below also, is DOUBLE the correct type? 32-bits would be FLOAT, perhaps you mean signed int instead?

src/lib/image/IOhtj2k/IOhtj2k/IOhtj2k.h
```diff
@@ -0,0 +1,48 @@
...
+//  Copyright (c) 2009 Tweak Software.
```
pbergeron-adsk, line 2: pls fix year, and I'm also unsure who the (C) should belong to here. You? Autodesk?

src/lib/image/io_htj2k/CMakeLists.txt
```diff
@@ -0,0 +1,31 @@
+#
+# Copyright (C) 2025 Sam Richards
+# SPDX-License-Identifier: Apache-2.0
```
pbergeron-adsk, line 3: Not sure who the (C) belongs to here. Probably you indeed, but I'll have to check.

richardssam:
> > Addressed the byte-vs.bit variable name, and split out the copy_scanline code into its own function.
> 
> Thank you! I wanted to approve the changes, and my brain is probably not braining properly, but I thought I found some problematic areas as per my notes below. Can you have a look? Or maybe explain what I'm missing?

I'm not quite sure what you are refering to?

pbergeron-adsk:
>> [...] I thought I found some problematic areas as per my notes below. Can you have a look? Or maybe explain what I'm missing?
> 
> I'm not quite sure what you are refering to?

I am specifically referring to the unanswered comments/concerns related to what seems to be incomplete and/or incorrect type handling.

I refer to the following comments in the unresolved conversations::

"Humm... Double? A double is 64 bits. Did you mean the type to be FLOAT? or an INT? (signed 32-bit?)"

"Here, you seem to be handling dtype (destination type?) to be UCHAR, HALF or USHORT but you're not handling the DOUBLE/UINT cases? (or is it float, and/or int cases?)..."

"is DOUBLE the correct type? 32-bits would be FLOAT, perhaps you mean signed int instead?"

richardssam:
How odd. I dont see those unresolved conversations. Regardless yes,I see
your point, I started supporting a branch of OJPH that has never gotten
merged into the mainline (i.e. all the floating point support), and clearly
some of that code got removed. For now, I'm going to remove the floating
point support until it makes it into the main codebasse.

Thanks...

Sam.

richardssam:
Is it useful to provide test images that we support and/or some that we
dont?

Sam.

richardssam:
I think I've addressed the issues with the floating point and integer code, which was based on code that has yet to be merged into OpenJPH. Once it gets more widely adopted, we can revisit that. So for now we are only supporting 8-bit, 10-bit, 12-bit and 16-bit unsigned integer per component, which should address the most typical use-cases for this format.

bernie-laberge:
> HTJ2K

I believe this addresses all your concerns right @pbergeron-adsk ? (If you could remove the requested change flag that would be great)
Let's merge this great contribution.
Thank you @richardssam !

PR #811, opened 2025-06-19
Windows - Make sure the right Python is used in Windows Debug

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Added the build type as parameter for `get_python_interpreter_args` to make sure that the right Python is returned. The change has no effect on other platform because the other platforms does not have a debug prefix for Python Debug.

Currently, the Python Release executable is **always** used in Windows Debug to install `numpy` and that causes issue with Nanobind since Nanobind is using Python Debug. The issue has been there for a long time

Windows

### DIFF
--- a/src/build/make_pyside.py
+++ b/src/build/make_pyside.py
@@ -54,7 +54,7 @@ def test_python_distribution(python_home: str) -> None:
         print(f"Moving {python_home} to {tmp_python_home}")
         shutil.move(python_home, tmp_python_home)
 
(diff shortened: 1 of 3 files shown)


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk:
I will need to investigate again. The Windows Debug build is still **not** using `python_d.exe` to install `numpy`.

bernie-laberge (approved):
LGTM

eloisebrosseau (approved):

PR #813, opened 2025-06-19
SG-39413: Add BMD SDI support for 120Hz

eloisebrosseau:
### Summarize your change.

Add 120Hz SDI format support to the BMD SDI path for 1080p, DCI 2K, UHD 4K, and DCI 4K. Note that the formats were added at the end of the list to follow #652 comment about user preferences based on format indexes.

These missing formats were supported by the Blackmagic DeckLink SDK but had never beed added to RV until today.

The video formats were added based on the DeckLink SDK documentation. However, since we don't have access to the monitor needed to test them, no tests were performed.


cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
LGTM
Thank you @eloisebrosseau !

PR #814, opened 2025-06-23
CI - Remove caching of the dependencies step for Windows

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Remove the caching of the dependencies step in the Windows build.

It is causing more issues then benefits - especially when there is a lot of PRs.


bernie-laberge (approved):
LGTM

eloisebrosseau (commented):
I might be wrong, but it looks like the changes you removed only affect the Windows release build, but the PRs are also failing because of the debug build. Is the issue with the debug build unrelated to this one?

cedrik-fuoco-adsk:
> I might be wrong, but it looks like the changes you removed only affect the Windows release build, but the PRs are also failing because of the debug build. Is the issue with the debug build unrelated to this one?

I did not see the issue with the debug build at first, but it's two different issues. This PR fixes the issue with the Release builds.

The debug issue should have been fixed by : https://github.com/AcademySoftwareFoundation/OpenRV/commit/f85b7dafa0db85ab0c38633eed060e378ba524e1 but I'll investigate.

cedrik-fuoco-adsk:
The Windows debug should be fixed by https://github.com/AcademySoftwareFoundation/OpenRV/commit/f85b7dafa0db85ab0c38633eed060e378ba524e1. One PR from @bernie-laberge was not updated to the latest, so it could explain why.

PR #816, opened 2025-06-23
Windows - Make sure the right Python is used in Windows Debug - Part 2

cedrik-fuoco-adsk:
### Linked issues
N/a

### Summarize your change.
Fix typo for the `variant` variable

Window debug build wasn't using the right Python executable when installing numpy.

Windows


bernie-laberge (approved):
LGTM

PR #817, opened 2025-06-23
Added sample window if diagnostics are empty

pbergeron-adsk:
### Summarize your change.

When no python plugin registers diagnostics callbacks, and the callback window is open, show an ImGui window with instructions on how to make your first diagnostics info panel

An empty diagnostics window is really boring and pointless and full of question marks from users.

macOS.

image


bernie-laberge (approved):
Nice !

PR #818, opened 2025-06-25
fix playback regression

pbergeron-adsk:
### Linked issues
Fixes SG-39441

### Summarize your change.

In the annotate tool, if we click and we're in select mode, then we need to toggle the playback state but also return immediately, which the code didn't do. When combined with a small fix that made playback stop if we started annotating, this caused a regression where clicking to toggle the state didn't work anymore: we would set the play state to true, but then immediately stop it.

Regression noticed by ASWF members.

macOS.

Just added a missing return statement and moved the code below the filter call such that if we have passive participants they should not be able to start playback.

### DIFF
--- a/docs/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-five.md
+++ b/docs/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-five.md
@@ -227,6 +227,7 @@ Some events will originate from RV itself. These include things like new-source 
 | range-changed |  |  | The time range changed |
 | narrowed-range-changed |  |  | The narrowed time range changed |
 | margins-changed | left right top bottom |  | View margins changed |
+| output-video-device-changed | outputVideoModuleName/outputVideoDeviceName |  | The output video device changed |
 | view-resized | old-w new-w | old-h new-h |  | Main view changed size |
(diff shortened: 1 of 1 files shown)


src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -916,6 +916,7 @@ class: AnnotateMinorMode : MinorMode
         if (_currentDrawMode eq _selectDrawMode)
         {
             togglePlayVerbose(true);
+            return;
         }
 
         if (filterLiveReviewEvents()) {
```
bernie-laberge, line 922: As noted by @eloisebrosseau : While we are here, we should move the if (_currentDrawMode eq _selectDrawMode) section AFTER the filterLiveReviewEvents(). RV passive participants, should not be able to start playback by clicking in the view.

pbergeron-adsk: Fixed

eloisebrosseau (approved):

bernie-laberge (approved):

kenmcgaugh:
Hey, I just tried this and while it fixes the click-to-play it doesn't fix the frame scrubbing issue for me.

PR #825, opened 2025-07-02
SG-39476: Refactor Hold and Ghost as session wide properties

eloisebrosseau:
### Summarize your change.

The Hold and Ghost properties were moved from the PaintIPNode to the SessionIPNode. Instead of being applied to each strokes and text boxes, they are now considered as global properties since all strokes and text boxes of all paint nodes should have the same Hold and Ghost values. The behaviour of the Paint node is pretty much the same expect for the fact that it uses the values from the Session node instead of stocking a value per stroke. The documentation was updated to reflect that change. The Annotation mode package was reverted to its previous behaviour since having the properties applied to the whole session means that mode doesn't need to be active for the properties to be updated and for the annotations to be rendered properly. The Annotation hook used for an OTIO-based Live Review session was also updated to set the properties on the session node once before looping on each layer. The Hold and Ghost properties in RVIO and RVSession were also updated to use the values set at the session node level, but the behaviour was not changed.

The Hold and Ghost properties should not have been per stroke and per text box since we expect all annotations to be affected the same way as soon as Hold or Ghost is enabled. This was also causing an issue with the Hold and Ghost buttons in the bottom toolbar of RV since the Annotation mode needed to be active for the properties to be properly set on all annotations on each click. Moreover, having the annotation mode always active was causing a few conflicts with other packages. Making the Hold and Ghost properties global to the session is more representative of the expected behaviour of the feature while minimizing other possible conflicts that could have arise from having the Annotation mode always on. 

- [X] Enable Hold and Ghost before opening the Draw panel to add annotations
- [X] Enable and disable Hold and Ghost with the different drawing tools with the Draw panel opened
- [X] Enable and disable Hold and Ghost with the different drawing tools with the Draw panel closed
- [X] Join an OTIO-based Live Review session with Hold and Ghost enabled
- [X] Upload annotations through Screening Room with Hold and Ghost enabled
- [X] Load an RV session with Hold and Ghost enabled

https://github.com/user-attachments/assets/3c72ccff-e040-43c2-8564-6be790710c1b

### DIFF
--- a/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
+++ b/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
@@ -5,6 +5,7 @@
 //
 #include <IPBaseNodes/PaintIPNode.h>
 #include <IPCore/PaintCommand.h>
+#include <IPCore/SessionIPNode.h>
 #include <IPCore/Exception.h>
 #include <IPCore/IPGraph.h>
 #include <TwkGLText/TwkGLText.h>
@@ -108,7 +109,7 @@ namespace
     void addBeforeCommands(PaintIPNode::LocalCommands* allCommands,
                            PaintIPNode::LocalCommands* currentFrameCommands,
                            const PerFramePaintCommands& beforeCommands,
-                           const int frame)
+                           const PaintEffects& paintEffects, const int frame)
     {
         int levelIndex = 1;
         bool isHoldedCommandsInFirstLevel = false;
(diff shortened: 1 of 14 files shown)


src/lib/ip/IPBaseNodes/PaintIPNode.cpp
```diff
@@ -335,10 +323,10 @@ namespace IPCore
...
         p.startFrame = startFrame;
         p.duration = duration;
-        p.hold = hold;
-        p.ghost = ghost;
-        p.ghostBefore = ghostBefore;
-        p.ghostAfter = ghostAfter;
+        // p.hold = hold;
+        // p.ghost = ghost;
+        // p.ghostBefore = ghostBefore;
+        // p.ghostAfter = ghostAfter;
```
cedrik-fuoco-adsk: Should be removed if not needed

eloisebrosseau: Oh you're right! I removed it, thanks!

bernie-laberge (approved):
Beautiful !
Thank you @eloisebrosseau !

PR #829, opened 2025-07-04
SG-39535: Update UI live review

eloisebrosseau:
### Summarize your change.

Enable and disable the Hold and Ghost buttons during a Live Review session. If the user is a participant, the `checked` state will be updated to show the feature is activated while keeping the button disabled. If the user becomes a presenter, the two buttons will be enabled to control if the feature should be activated or deactivated for all the participants. Note that sending the events to update the UI is done in an other PR in Commercial RV. 

An RV participant of a Web to RV Live Review session could render the Hold and Ghost feature but its UI was not updating to show if the feature was activated or not like it is for the presenter.

Validating the UI is updated when:
- [X] Joining a Web to RV and RV to RV Live Review session when Hold and Ghost is activated
- [X] Enabling and disabling Hold and Ghost after joining a Web to RV and RV to RV Live Review session
- [X] Changing presenter in a RV to RV Live Review session


bernie-laberge (approved):
LGTM
Thanks!

cedrik-fuoco-adsk (approved):

PR #832, opened 2025-07-09
SG-39544: Fix Angular Mask Pivot Value on Compare

salolavyadsk:
### Summarize your change.

When using the wipe compare mode and changing the compare clip, we encountered an issue where the compare effect did not have access to the correct .otio.global_scale and transform properties. This happened because the compare_hook was retrieving the source node from the compare clip, which lacked these necessary fields, especially in certain compare workflows. To resolve this, we updated the compare_hook to always get the source node from the original clip, ensuring that the global scale and transform are consistently available for effect calculations. Additionally, we improved the annotation hook by replacing the use of first_source_node = commands.sourcesAtFrame(0)[0]—which could return an incorrect or missing source node—with logic that follows the same approach as in the compare hook. This makes both effect hooks more robust and ensures accurate property application in all compare and annotation scenarios.

The reason for this change is that relying on commands.sourcesAtFrame(0)[0] to determine the source node could result in incorrect or missing media information, especially in complex timelines or when using compare modes. This could cause the annotation effect to use the wrong aspect ratio or global scale, leading to inaccurate rendering of annotations. By updating the logic to consistently retrieve the correct source node, we ensure that the annotation effect always has access to accurate media properties, resulting in more reliable and visually correct annotations across all timeline and compare scenarios.

-  Joined a live-review session between RV and Creative Review on macOS.
- Changed the clip during the session to verify if the annotation pivot points update accurately.
- Observed that pivot points remained consistent and correctly positioned after switching clips.


bernie-laberge (approved):
LGTM
Thank you @salolavyadsk !

PR #834, opened 2025-07-10
SG-39510: Block options for Live Review participants

eloisebrosseau:
### Summarize your change.

Disable buttons, menus and related hotkeys that shouldn't be used by a participant of a Live Review session. For buttons, their corresponding tooltip is also updated to say "You must be presenter to use" when the user is a participant. If the user is a presenter or if they are not in a Live Review session anymore, all the disabled buttons and menus are enabled again and the tooltips are set back to their default description. When clicking on the media or using a hotkey, a widget is displayed with the string "You are not the presenter". Moreover, the Annotation panel will be closed when joining a Live Review session is it open at that moment.

During a Live Review session, an RV user was still capable of clicking on multiple buttons and menu options that could cause mismatch between the presenter's screen at the participant screen. A Live Review participant should only be allowed to click on options that don't change the ongoing session.

This is the list of disabled feature for a Participant of a Live Review session:

- Control, Edit, Color, View, Source, Stack, Sequence and Layout menu items
- Annotation, Session Manager and Default Views of the Tools menu
- All hotkeys for the above menu items
- Left-click and right-click on the timeline and the timeline magnifier
- All bottom toolbar options, except toggling image info, networking dialog/Live Review panel, timeline, timeline magnifier, and volume
- All the top toolbar options, except toggling full-screen mode, framing image in view, and selecting the background style

### DIFF
--- a/src/lib/app/mu_rvui/mode_manager.mu
+++ b/src/lib/app/mu_rvui/mode_manager.mu
@@ -213,11 +213,13 @@ class: ModeManagerMode : MinorMode
     {
         \: (int; )
         {
-            if entry.mode eq nil
-                 then UncheckedMenuState
-                 else if entry.mode._active
-                        then CheckedMenuState
-                        else UncheckedMenuState;
+            if (filterLiveReviewEvents())
+                then DisabledMenuState
+                else if entry.mode eq nil
+                     then UncheckedMenuState
+                     else if entry.mode._active
+                            then CheckedMenuState
+                            else UncheckedMenuState;
         };
     }
 
@@ -674,6 +676,11 @@ class: ModeManagerMode : MinorMode
             {
                 bind(m.event, \: (void; Event ev)
                 {
+                    if (filterLiveReviewEvents() && (m.name == "session_manager" || m.name == "annotate_mode"))
+                    {
+                        sendInternalEvent("live-review-blocked-event");
+                        return;
+                    }
                     toggleModeEntry(ev, m, this);
                 });
             }
--- a/src/plugins/rv-packages/multiple_source_media_rep/multiple_source_media_rep.py
+++ b/src/plugins/rv-packages/multiple_source_media_rep/multiple_source_media_rep.py
@@ -312,7 +312,9 @@ class MultipleSourceMediaRepMode(rvtypes.MinorMode):
         self._media_representation_btn = QtWidgets.QToolButton(self._bottomToolBar)
         self._media_representation_btn.setToolTip("Select media playback type")
         self._media_representation_btn.setProperty("tbstyle", "solo_menu")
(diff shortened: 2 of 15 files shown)


src/plugins/rv-packages/multiple_source_media_rep/multiple_source_media_rep.py
```diff
@@ -560,6 +562,29 @@ def _on_postpone_force_update_media_info(self, event):
...
         self._force_update_media_info = True
 
+    def _on_presenter_changed(self, event):
+        """
+        This method is bound to the internal-sync-presenter-changed event.
+        It disables the Multiple Media Representation button for all participants
+        that are not the presenter.
+        """
+        event.reject()
+        is_disabled = rvc.filterLiveReviewEvents()
```
bernie-laberge: We have to allow a passive Live Review participant to switch to an alternate media representation as this is a feature for the Web-2-RV Live Review session.
I think the multiple_source_media_rep.py file should be left untouched.

src/lib/app/mu_rvui/mode_manager.mu
```diff
@@ -213,11 +213,13 @@ class: ModeManagerMode : MinorMode
     {
         \: (int; )
         {
-            if entry.mode eq nil
-                 then UncheckedMenuState
-                 else if entry.mode._active
-                        then CheckedMenuState
-                        else UncheckedMenuState;
+            if (filterLiveReviewEvents())
```
bernie-laberge, line 216: Will this disable the RV Screeing Room functionality ?
(Because right now, a passive participant can submit notes to ShotGrid via the RV Screening Room Details pane)

bernie-laberge (approved):
LGTM
Wow ! This was a massive undertaking. Thank you @eloisebrosseau .
Please make sure to revert src/plugins/rv-packages/multiple_source_media_rep/multiple_source_media_rep.py as switching the media representation in a Web-2-RV or an RV-2-RV Live Review session is a required feature.
This allows RV participant to see a higher resolution media representation in a Web-2-RV or an RV-2-RV Live Review session.

src/lib/app/mu_rvui/mode_manager.mu
bernie-laberge: After discussing it, since that it opens the door for passive Live Review participants to double click on a version to open it, then let's block it and figure out a way to reintroduce the notes when we do multiuser control

src/plugins/rv-packages/multiple_source_media_rep/multiple_source_media_rep.py
eloisebrosseau: True! I reverted the changes made to that file!

cedrik-fuoco-adsk (approved):

PR #835, opened 2025-07-11
Live review diagnostics no muc broadcast python

pbergeron-adsk:

This adresses a very thorny issue where during a RV live review session, some Rv commands would generate a bajillion internal events that would get broadcast over the live review service. This issue would compound if supplied plugins or customer plugins did something nasty, like when they themselves would hook onto some events, and would themselves generate new events by deleting/creating/setting new properties, sometimes in a loop. In those cases, the live review service would choke from the sheer quantity of events, and performance would die.

We have so far identified only a few Mu commands that did this (below) but any client of those commands will cause the flood of events even if only a moderately sized playlist was managed by RV.

- clearSession()
- addSources()
- addSourcesVerbose()

(but there are probably other commands that do generate this event flood).

### Summarize your change.

Given this situation, we needed to find a way to drastically reduce the number of events that get pushed through the network event pipe. We came up with the following pattern: 

- From within the Mu stubs for the command, send a custom event "generic-rv-command") where the payload is the name of the rv command and its arguments (packed as a simple json payload)
- Set a flag (push-eat-broadcast-events) that will prevent broadcasting of events  until the flag is unset (popeat-broadcast-events).
- Wrap the functionality in a RemoteRvCommand class, which in its constructor sends the "generic-rv-command" event then blocks events from being broadcast over the network, and in the destructor releases the flag.

(The above approach means that the pattern can often be implemented with just one, or maybe a few lines of code in Rv's Mu command stubs)

Now, when the corresponding "remote-sync-generic-rv-command" is caught by the participant's live-review session, the command name and arguments are unpacked from the json payload, and the rv command is called with arguments.

Put in other words, this fix silences all events that originate from a mu command, and replaces it by a single event that tells the receiving client to call the mu command instead.

Tested on macOS.

* In the case of addSources() and addSourcesVerbose, I'm not sure (yet) if the filenames need to be sanitized first before packaging them as arguments.

* To test this fix, you need to fetch the branch with the same name in the private Autodesk RV git repo.

* In the RemoteRvCommand implementation, you'll notice that I'm forming the json payload manually with string operations. This is intended, for 3 reasons.  
   ** (a) I didn't find a small-enough FOSS json packager that worked AND had a compatible license. 
   ** (b) When I did try one of these 3rd-party options, the amount of code needed to use the json toolkit resulted in a lot more code than my homemade implementation
   ** (c) Rv doesn't have JSON-handling capability in its C++ code, and I didn't see the need to have yet-another 3rd-party dependency.

No screenshots to provide, only better performance.


src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
```diff
@@ -0,0 +1,199 @@
...
+
+            m_session->userGenericEvent("push-eat-broadcast-events", "");
+        }
+
+        ~RemoteRvCommand()
+        {
+            m_session->userGenericEvent("pop-eat-broadcast-events", "");
+        }
+
+        template <typename... Args>
```
rogernelson, line 44: Why not use something like rapidjson instead of rolling your own json writer?

src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
```diff
@@ -0,0 +1,199 @@
+//
+//  Copyright (c) 2012 Tweak Software.
```
bernie-laberge: Please use the following standard header instead:
```
#
# Copyright (C) 2025  Autodesk, Inc. All Rights Reserved.
#
#
```

src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
```diff
@@ -0,0 +1,199 @@
...
+//  All rights reserved.
+//
+//  SPDX-License-Identifier: Apache-2.0
+//
+//
+#ifndef __IPCore__RemoteRvCommand__h__
```
bernie-laberge: We can now use 
`#pragma once`

src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
```diff
@@ -0,0 +1,199 @@
...
+#define __IPCore__RemoteRvCommand__h__
+
+#include <iostream>
+#include <sstream>
+#include <string>
+#include <vector>
+#include <type_traits>
+#include <IPCore/Session.h>
```
bernie-laberge: Please move this #include <IPCore/Session.h> at the top:
we usually do higher level headers first, followed by lower level headers

src/lib/app/RvApp/CommandsModule.cpp
```diff
@@ -1402,10 +1409,25 @@ namespace Rv
...
+
                 filesAndOptions.emplace_back(
                     IPCore::Application::mapFromVar(arg));
             }
 
+            allFilesAndOptions.push_back(filesAndOptions);
+        }
+
+        // silences broadcasting events below this constructor
+        IPMu::RemoteRvCommand remoteRvCommand(s, "addSourceVerbose",
```
rogernelson: Not sure I like the name RemoteRvCommand, since it's more of a macro event, which could have other uses besides this particular case.

What about something like this?
```
IPMu::MacroEvent(s, "addSourceVerbose", allFilesAndOptions, tag);
```

Then in the implementation instead of having push/pop-eat-broadcast-events you have start/end-macro-event.

You could listen for the start/end macro-event in your live review plugin use those to set/unset your guards for eating the events.

Plus this way, we don't bake the concept of LiveReview into RV and keep it all confined to the plugin.

src/lib/app/RvApp/CommandsModule.cpp
```diff
@@ -1358,6 +1359,10 @@ namespace Rv
             sargs.push_back(IPCore::Application::mapFromVar(arg));
         }
 
+        // silences broadcasting events below this constructor
+        IPMu::RemoteRvCommand remoteRvCommand(s, "addSources", sargs, tag,
```
bernie-laberge, line 1363: Correct me if I am wrong but I don't think this startegy will work for addSources since the addSources command is asynchronous (See comment below on line 1368).
But your startegy should work fine for clearSession() + addSourcesVerbose() because they are both synchronous commands.

src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
pbergeron-adsk: See my PR comment. In a nutshell, using it was more trouble than not, adds yet another external dependency.

src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
pbergeron-adsk: Fixed

src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
pbergeron-adsk: Fixed

src/lib/app/RvApp/CommandsModule.cpp
pbergeron-adsk: It's not a macro event, it's really a Rv command that you execute remotely, especially if you look at the python implementation of how the event is handled in live review (eg: pseudo-snippet below)

As you can see, the python code looks for the existence of the _command_, puts the executable symbol in command_func, and calls it with expanded args if applicable. Even if you prefer the meaning of "macro event", it would still be reduced to be used in the context of calling a single mu/rv command.

            if hasattr(commands, command_name):   
                command_func = getattr(commands, command_name)
                    if args:
                        result = command_func(*args)
                    else:
                        result = command_func()

src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
rogernelson: No need to add another dependency, Qt has a JSON parser.

src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
pbergeron-adsk: Fixed

src/lib/ip/IPMu/IPMu/RemoteRvCommand.h
pbergeron-adsk: Rewrote it using Qt json parser. Cleaner code.

eloisebrosseau (approved):

PR #836, opened 2025-07-14
SG-39680: Fixed crash on exit with registered imgui callback (temp patch)

pbergeron-adsk:
### Linked issues

Fixes SG-39680

### Summarize your change.

I silenced the reference decrement in the python object deleter; I'll have to wait for Cedrik to create a proper fix for this. 
Ideally we want to call clearCallbacks() before exiting the app, there seems to be a weird race condition that prevents us from doing so. 

Furthermore, decrementing the refcount in the deleter appears to run counter to the fact that we're already decrementing the refcount when we unregister the callback.  I'm not sure if we're meant to keep the refcount derement in the deleter, or even if the deleter is needed at all at this point. 

Segfault on exit, need to send beta build.

macOS

Make sure we don't sefault on exit

n/a

### DIFF
--- a/src/lib/app/ImGuiPythonBridge/ImGuiPythonBridge.cpp
+++ b/src/lib/app/ImGuiPythonBridge/ImGuiPythonBridge.cpp
@@ -7,8 +7,10 @@ namespace Rv
 {
     void ImGuiPythonBridge::PyObjectDeleter::operator()(PyObject* obj) const
     {
-        if (obj)
-            Py_DECREF(obj);
+        // seems to crash on exit if we have a callback registered.
+        // so patching this out for now.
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):
LGTM
I agree with you @pbergeron-adsk : Let's fix the crash on exit first and foremost then @cedrik-fuoco-adsk can take a look when he gets back.
Thanks!

eloisebrosseau (approved):

pbergeron-adsk:
@bernie-laberge @eloisebrosseau I have merged this, but I'll assign the bug (not resolve it) to Cedrik so he can review upon his return.

PR #837, opened 2025-07-15
SG-39640: Add fast conversion for ProRes 4444 with alpha

eloisebrosseau:
### Summarize your change.

Add a dedicated fast conversion for packedAYUV64 to planarYUVA16 (AV_PIX_FMT_AYUV64LE -> AV_PIX_FMT_YUVA444P16LE). This conversion is for ProRes 4444 with an alpha value. Note that clang-format fixed some code formatting issue from #806. The only change related to the conversion is in `copyFrame()`

Fast conversions for ProRes 422 and ProRes 4444 without alpha were added in #732. However, we were missing a fast conversion implementation for ProRes 4444 with alpha. With a 4K 60 fps media file of that format, playback was buffering a lot because the standard conversion applied by FFmpeg was too slow. Adding a dedicated conversion between the source and destination format for that ProRes format fixed that issue.

Loading ProRes 4444 and ProRes 4444XQ with an alpha value on macOS 15.5 with an M3 chip.


bernie-laberge (approved):
Superb
Thank you @eloisebrosseau !

pbergeron-adsk (approved):
Approved. Looks good.

PR #838, opened 2025-07-17
Channel flood / reordering and dither support when using OpenColorIO

markreidvfx:
### Summarize your change.

If the OpenColorIO plugin is enabled, OpenRV does not support channel selecting, reordering or dithering.
This PR allows the TopViewToolBar menus to function the same way it does when OCIO is disabled. The main menu items work too for channel reordering. The code added to the OCIOIPNode node is very similar to the non-OCIO DisplayIPNode. The identical shader code is utilised for dithering, and the extra properties are named the same. 

```
color.dither
color.channelOrder
color.channelFlood
```

https://github.com/user-attachments/assets/b8962e1c-c0df-496e-bdc3-fcbb4b546846

The PR also includes 2 additional fixes
- Fixes a Qt issue where screen names end up being blank. I was running into this on macOS where Qt is using deprecated apis to determine the screen name. 
- Fixes error message that can occur with recent changes to the OCIO node. The fix temperraly disables the OCIONode during display/view property changes. This Prevents it from rebuilding shaders while it may be in an invalid state.

Rocky 9, CentOS, MacOS


src/lib/app/RvCommon/RvTopViewToolBar.cpp
```diff
@@ -901,12 +903,15 @@ namespace Rv
...
+        bool hasDisplay = hasStandardDisplayPipeline();
+        bool hasOCIODisplay = hasOCIODisplayPipeline();
+
+        if (hasDisplay || hasOCIODisplay)
         {
 
             IntPropertyEditor editor(m_session->graph(),
                                      m_session->currentFrame(),
-                                     "@RVDisplayColor.color.channelFlood");
+                                     hasOCIODisplay ? "#OCIODisplay.color.channelFlood" : "#RVDisplayColor.color.channelFlood");
```
bernie-laberge, line 914: Should we keep using the @ instead of # ?

Based on Reference Manual Chapter 2:
By addressing properties using their node’s type name, you can affect only active nodes with that type without needing search for the exact node(s).There is an additional shorthand using “@” in front of a type name:

@RVDisplayColor.color.brightness 
The above would affect only the first RVDisplayColor node it finds instead of all RVDisplayColor nodes of depth 1 like “#” does. This is useful with presentation mode for example because setting the brightness would be confined to the first RVDisplayColor node which would be the one associated with the presentation device. If “#” was used, all devices would have their brightness modified. The utility of the “@” syntax is limited compared to “#” so if you are unsure of which to use try “#”

src/plugins/rv-packages/ocio_source_setup/ocio_source_setup.py
```diff
@@ -186,8 +186,13 @@ def F(event):
         groupName = "RVDisplayPipelineGroup"
         dpipeline = groupMemberOfType(group, groupName)
         dOCIO = groupMemberOfType(dpipeline, "OCIODisplay")
+        # Both 'display' and 'view' must be set together.
+        # Disable the OCIONode during display/view propety changes.
+        # Prevents node from rebuilding shaders while it may be in an invalid state.
+        commands.setIntProperty(dOCIO + ".ocio.active", [0], True)
         commands.setStringProperty(dOCIO + ".ocio_display.display", [display], True)
         commands.setStringProperty(dOCIO + ".ocio_display.view", [view], True)
+        commands.setIntProperty(dOCIO + ".ocio.active", [1], True)
```
bernie-laberge, line 195: Great catch ! Thanks!

bernie-laberge (approved):
Great contribution @markreidvfx !
Thank you so much !

src/lib/app/RvCommon/RvTopViewToolBar.cpp
markreidvfx: This was actually intentional. Previously to get channel select with OCIO you needed to use the channel_select plugin as a workaround. Many of our users are used to it and I did this to replicate the plugin's behaviour 

https://github.com/AcademySoftwareFoundation/OpenRV/blob/2ed0a1eb8d61a932b58be2e27b8d07361870aaf5/src/plugins/rv-packages/channel_select/channel_select.py#L76

In review sessions it's common to want to look at a particular channel (looking at film grain for example), having it not reflected on the presentation device is frustrating IMO.

pbergeron-adsk (approved):
Approved!

PR #842, opened 2025-07-22
Minor update to boost, the older API has now been retired.

richardssam:
### Boost::filesystem update

This is a pretty trivial change, just updating a boost API change, that C++ on OSX was complaining about.

Seems like the older API is being retired.


bernie-laberge (approved):
LGTM
Thanks @richardssam !

PR #846, opened 2025-07-24
(Enhancement) - Directory filter for drag and drop.

richardssam:
### Adding Directory Filter Event.

We have some sidecar files which are often in directories of images, so if we drag the directory of images into RV, RV will attempt to load the sidecar files too. This directory filter allows you to filter out the un-needed sidecar files so they will never be loaded.

```
def filter_directory (event):
  file = event.contents();
  if file.endswith(".info"):
  event.setReturnContent("0");
  return
  event.setReturnContent("1");

commands.bind("default", "global", "directory-filter", filter_directory, "Doc string");
```

I was looking for a way to filter out files that should never be treated as loadable media from a directory scan that happens when a drag and drop event occurs with a directory. I couldnt see another easy approach. 

Windows and OSX

I dont totally know if there are other areas that this filter should also be run, for now its just doing the drag and drop event, it feels like the file-loader GUI encourages better behaviour.

Also, is there somewhere I should add the above documentation? (and possibly flesh it out).
I also dont quite understand the bind, I'm assuming you would want a "global" bind, but am not 100% sure.

### DIFF
--- a/docs/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-five.md
+++ b/docs/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-five.md
@@ -303,3 +303,22 @@ Sometimes an image is not available on disk when RV tries to read. This is often
 
 bind("missing-image", missingImage); 
 ```
+
+#### 5.6.4 Drag and Drop Filtering.
+
+Sometimes dragging in a directory into RV, the directory might contain files that RV cannot load (e.g. sidecar files), the directory-filter allows you to filter out certain file types.
+
+The Filter should return "1" if you want the file, "0" if you dont.
+
+A python example of using this is:
+
+```
+def filter_directory (event):
+  file = event.contents();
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
LGTM
Thank you @richardssam !

eloisebrosseau:
Hi @richardssam! This PR is being tested by our QA team right now so we should be able to merge it soon! To answer the questions in your PR:

- I think it would be great if you could add the documentation in [Chapter 5 - Event Handling](https://aswf-openrv.readthedocs.io/en/latest/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-five.html)! The `directory-filter` event should be added to the `Internal Events` table and you could add a new sub section to explain how to use it with the code snippet!
- Since in your case, you just want to show the basic approach of binding the `directory-filter` event, I think "global" is the appropriate choice

richardssam:
Added documentation. Let me know if you need anything else.

PR #847, opened 2025-07-27
Fix 2x RemoteRvCommands

bernie-laberge:
### Linked issues
NA

### Summarize your change.

Note that this commit is only impacting Open RV when using the Open RV Live Review functionality.

This commit adds addSourceVerbose() to the list of the RemoteRvCommands.

It also solves an issue with the addSourcesVerbose() command where the RemoteRvCommands was not executed before it was defined after before-progressive-loading which was preventing the generic-rv-command event from being broadcasted to the participants.

addSourcesVerbose() and addSourceVerbose() were not behaving as expected when used in RV Live Review.

Successfully tested on macOS 

### DIFF
--- a/src/lib/app/RvApp/CommandsModule.cpp
+++ b/src/lib/app/RvApp/CommandsModule.cpp
@@ -1303,6 +1303,13 @@ namespace Rv
 
         IPGraph::GraphEdit edit(s->graph());
 
+        // silences broadcasting events below this constructor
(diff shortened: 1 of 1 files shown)


pbergeron-adsk:
LGTM !!

pbergeron-adsk (approved):

cedrik-fuoco-adsk (approved):

PR #849, opened 2025-07-27
Fix Y-Inverted Color Picker

markreidvfx:
### Summarize your change.

Fixes the color inspector from calculating the wrong coordinate when a RVFileSources has multiple media paths with different orientations

We use multiple media paths to add audio to images sequences from a proxy mov file.

```
rv.commands.addSource(["/path/to/image.####.exr", "/path/to/proxy_with_audio.mov"])
```

The current logic is to use the `ImageStructureInfo` of last movie in the list and not what is currently visible. In some cases the second path might have a different coordinate system then the image sequence. In our cases the Y coordinate is opposite causing the color picker sample location to be upside down.

This change reuses the exisiting `getMediaFromContext` method to get the same media that is displayed. 

Linux, macOS

### DIFF
--- a/src/lib/ip/IPBaseNodes/FileSourceIPNode.cpp
+++ b/src/lib/ip/IPBaseNodes/FileSourceIPNode.cpp
@@ -724,7 +724,7 @@ namespace IPCore
 
     FileSourceIPNode::MediaPointer
     FileSourceIPNode::getMediaFromContext(ImageComponent& selection,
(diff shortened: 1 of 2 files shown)


eloisebrosseau (approved):

bernie-laberge:
Testing successfully completed. Ready to merge.

bernie-laberge (approved):
LGTM

PR #850, opened 2025-07-28
SG-40371: Presentation Mode Enhancements

markreidvfx:
### Summarize your change.

This PR add some features to aid in automating Presentation mode

#### Add command to set the presentation device

Add new command `setPresentationDevice`. You can do this with the writeSettings command but that requires a restart of RV or opening of the preference window to trigger the the preference being applied.

#### Fix Video Format Overlay

View->Presentation Settings->Video Format overlay was disable because it was causing a crash in `RV-4.0+`. I fixed the crash and it works again. It is useful for debugging presentation devices.

image

image

#### Presentation device additional options

This is pretty funny, **disable** the rich text option on this widget.   don't think this widget needs bold, underline text, colors etc.  Can happen if you are  copy pasting text into widget 

image

#### Add command to apply preferences

`rv.commands.writeSetting` command does not immediately apply the preferences you set. It only saves them to the configuration file. A restart of RV or opening of the preference window is required to trigger the changes to take effect. The `loadSettingsIntoOptions` command addresses this issue by allowing preferences to be applied. I used the same name as the c++ function for the command. This name might not be ideal, I’m open to suggestions for a better name. This command has proven to be very useful, for setting additional preferences like modifying `Treat Stylus Events as Mouse Events` which currently not possible.

Linux, macOS


pbergeron-adsk (approved):
LGTM

src/lib/app/mu_rvui/presentation_mode.mu
```diff
@@ -234,6 +234,79 @@ class: PresentationControlMinorMode : MinorMode
...
+            sep = (height / 4.0) * 0.1;
+
+        int y = height - hm;
+
+        drawRoundedBox(m, y - _totalHeight - sep * 4,
+                        width - m, y,
+                        20,
+                        Color(0,0,0,.75), Color(0,0,0,1));
+
+        for_each (t; [_deviceText, _formatText, _dataFormatText, _audioFormatText])
```
pbergeron-adsk, line 293: We may have to scale dimensions or do something special here for HDPI.... Will look at this during view testing, not holding PR for this

bernie-laberge: Note that that the original code was not properly handling High-DPI displays. So this is not @markreidvfx's fault. I will create a ticket for it and I will take care of fixing it.

bernie-laberge:
View Testing was completed successfully. Thank you @markreidvfx !

PR #851, opened 2025-07-28
SG-40372: Add Basic File Operations to the RVFileDialog

markreidvfx:
### Summarize your change.

Adds support for basic file operations within the `RVFileDialog` window. Please note I've not had time to test this on Windows.

file_operations2

Adds a context menu with the following options
- create folders
- delete files/folders
- rename files/folders
- refresh

#### New folder button
The `New Folder` button was already in the ui files. It was just hidden because it was never implemented

image

Linux, macOS


eloisebrosseau (approved):

pbergeron-adsk:
@markreidvfx We have found two issues with this PR:

**Issue 1:**

* In a folder FolderRoot that already has folders, select a folder, let's say FolderABC.
* In the empty space, right click, create new folder FolderDEF.
* User is navigated to FolderABC.

Note that if you don't select FolderABC first, the bug doesn't appear.

Expected result: User should stay in FolderRoot, optionally goes to FolderDEF, but definitely not FolderABC.


**Issue 2:**

* In a folder FolderRoot that already has folders, right-click in the empty space, hit Refresh, nothing happens, which is expected.

but

* Click on a folder, let's say FolderABC.
* Right-click in the empty space, hit Refresh, user is navigated to FolderABC.


They are probably different manifestations of the same bug (I'd expect that when you click on a folder, that sets the "current folder", and when you're done creating a folder, it calls refresh, which navigates to the previously-selected folder)

PR #852, opened 2025-07-30
SG-39510: Close annotation panel when joining Live Review

eloisebrosseau:
### Summarize your change.

When joining a Live Review session, the Annotation panel should be closed if it was still open. Since a participant shouldn't be allowed to annotate the media in a standard Live Review session, there is no reason for the panel to be available.

In a previous PR, options and menus related to annotations were blocked for participants of a Live Review session. However, if the Annotation panel was opened when joining a session, it would stay that way. This can be confusing to the participant since all other options are disabled. 

Testing joining a Live Review session with the Annotation panel opened and closed was perform on macOS

### DIFF
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -1804,6 +1804,14 @@ class: AnnotateMinorMode : MinorMode
     method: nextSlot (void; bool b) { nextAnnotatedFrame(); }
     method: prevSlot (void; bool b) { prevAnnotatedFrame(); }
 
+    method: onPresenterChanged(void; Event event)
+    {
+        if (_active && filterLiveReviewEvents())
+        {
+            toggle();
+        }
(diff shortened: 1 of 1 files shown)


cedrik-fuoco-adsk (approved):
LGTM!

bernie-laberge:
Are we sure we want to do that ? That might be annoying for users if they pass the Presenter baton from one another...
That being said: we could give it a try and if there is a user revolt :-) we could revert it.

eloisebrosseau:
> Are we sure we want to do that ? That might be annoying for users if they pass the Presenter baton from one another... That being said: we could give it a try and if there is a user revolt :-) we could revert it.

This is what the web app is doing when a participant joins with the Annotation panel open. Since a participant is not allowed to use the Annotation panel once in a session and all other menus related to annotations are blocked or disabled, I think it would be counter-intuitive to keep the panel open

bernie-laberge (approved):
LGTM

PR #855, opened 2025-07-31
SG-39832: Block Zoom and Pan to participants of Live Review

eloisebrosseau:
### Summarize your change.

Block the Zoom and Pan hotkeys if RV is a participant of a Live Review session.

Multiple menus and options are now disabled or blocked when RV is a participant of a Live Review session. However, it was still possible for a user to enable Zoom and Pan through their hotkeys. As a participant, RV should only be a viewer and display exactly what's on the presenter's screen.

Zoom and Pan outside and inside a Live Review session was tested on macOS.

https://github.com/user-attachments/assets/c5004e74-f102-4edd-8c1d-f6adf7469bd0


bernie-laberge (approved):
LGTM

PR #858, opened 2025-08-05
Fix issue when importing pyimgui and pyimplot in a Python plugin on Windows

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
PyImGUI and PyImPlot could not be imported on Windows.

Make sure that the pyd module is generated and copy it to the right location.

Windows


bernie-laberge (approved):
LGTM
Thanks!

eloisebrosseau (approved):

PR #862, opened 2025-08-06
Refactor diagnostics view

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.

The primary issue was an always-on 25fps timer that consumed CPU/GPU resources even when the diagnostics window was never opened, along with eager initialization of ImGui/ImPlot contexts and Python bridge overhead. 

The solution was to implemented a lazy initialization where all expensive operations (ImGui context creation, timer connections, Python bridge setup) are deferred until the diagnostics window is first shown.

Additionally, I improved the shutdown sequence to prevent the application from hanging and crashing during exit.

Finally, I removed duplicated code in DiagnosticsView that was not used.

- Segmentation fault when leaving the application because the code was trying to delete Python references that did not exist anymore
- Slowness and High CPU usage seen during integration tests with Live review plugin. There were a lot of different initialization that was done before showing the diagnostics windows and it was slowing the normal operation of the application.

### DIFF
--- a/src/lib/app/RvCommon/DiagnosticsView.cpp
+++ b/src/lib/app/RvCommon/DiagnosticsView.cpp
@@ -16,6 +16,8 @@
 
 #include "DiagnosticsView.Roboto_Regular"
 
+#include <iostream>
+
 namespace Rv
 {
     class DiagnosticsModule
@@ -34,7 +36,6 @@ namespace Rv
 
         if (_initCount == 1)
         {
-            _initCount++;
             ImGui::CreateContext();
             ImPlot::CreateContext();
             ImGui::GetIO().ConfigFlags |= ImGuiConfigFlags_DockingEnable;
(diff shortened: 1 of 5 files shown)


src/lib/app/RvCommon/DiagnosticsView.cpp
```diff
@@ -51,72 +52,63 @@ namespace Rv
 
             if (_initCount == 0)
             {
+                int callbackCount = Rv::ImGuiPythonBridge::nbCallbacks();
```
bernie-laberge: Leftover perhaps ? You could get rid of that line I guess ?

cedrik-fuoco-adsk: Good catch! I think I had a std::cout in there before. I'll remove it!

bernie-laberge (approved):
LGTM

src/lib/app/RvCommon/DiagnosticsView.cpp
cedrik-fuoco-adsk: Fixed

eloisebrosseau (approved):

PR #863, opened 2025-08-07
SG-39830: Add refreshOutputVideoDevice to CommandsModule

eloisebrosseau:
### Summarize your change.

Add refreshOutputVideoDevice in CommandsModule.cpp and rv_commands_setup.py to access the function in Python.

When loading a media file, RV calls systematically `setOutputVideoDevice` in `readGTO` of RVSession.cpp. However, when joining an OTIO-based Live Review session, this path is never used and the output video device will not be set properly after quitting and joining back a session. To replicate the normal behaviour the output video device should be set when joining an OTIO-based Live Review session. To do so, that C++ function needs to be available in Python through the commands module. Note that this behaviour was not affecting RV to RV Live Review session since in that case, `readGTO` was being called.

Quitting and joining back an OTIO-based Live Review session with the presentation mode enabled through NDI in RV.

### DIFF
--- a/src/lib/ip/IPMu/CommandsModule.cpp
+++ b/src/lib/ip/IPMu/CommandsModule.cpp
@@ -5128,6 +5128,14 @@ namespace IPMu
         NODE_RETURN(stype->allocate(idstr));
     }
 
+    NODE_IMPLEMENTATION(setOutputVideoDevice, void)
+    {
+        Session* s = Session::currentSession();
+        auto outputDevice = s->outputVideoDevice();
+
+        s->setOutputVideoDevice(outputDevice);
+    }
+
     NODE_IMPLEMENTATION(audioTextureID, int)
     {
         Session* s = Session::currentSession();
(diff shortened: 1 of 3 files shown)


src/lib/ip/IPMu/CommandsModule.cpp
```diff
@@ -5128,6 +5128,14 @@ namespace IPMu
         NODE_RETURN(stype->allocate(idstr));
     }
 
+    NODE_IMPLEMENTATION(setOutputVideoDevice, void)
```
bernie-laberge: Perhaps we should add a short description of the new command in OpenRV/src/lib/app/mu_rvui/commands.mud

bernie-laberge (approved):
LGTM
Thanks!

markreidvfx:
does this have any overlap with the `setPresentationDevice` command I'm hoping to add in my PR?

https://github.com/AcademySoftwareFoundation/OpenRV/pull/850/commits/85863c5a900519093b982fea0d8f9df0139aefd0

bernie-laberge:
Oups ! Yes it does overlap ! Sorry about that.
FYI: @eloisebrosseau

src/lib/ip/IPMu/CommandsModule.cpp
bernie-laberge: Thanks!

PR #864, opened 2025-08-08
Fix Window>Fit (w shortcut) for high dpi screen

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Take the DPI scaling into account when calculating the width and height in the function.

There are one other issues that was discovered during testing:
Intermittent:  Sometime the scaling of the UI get fuzzy when moving OpenRV from a non-high-dpi screen to a high dpi screen.

Pressing W multiple time (or using the menu) would crash OpenRV after 3 times.

MacOS

### DIFF
--- a/cmake/dependencies/imgui/CMakeLists_PyImGUI.cmake
+++ b/cmake/dependencies/imgui/CMakeLists_PyImGUI.cmake
@@ -36,6 +36,11 @@ ENDIF()
 
 NANOBIND_ADD_MODULE(pyimgui ${IMGUI_NB_SOURCES})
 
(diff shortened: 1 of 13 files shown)


bernie-laberge (approved):
LGTM
If you could validate that your fix still works when disbling the High-DPI support in RV that would be great:
export RV_NO_QT_HDPI_SUPPORT=1

eloisebrosseau (approved):

cedrik-fuoco-adsk:
New commit to fix an issue when RV_NO_QT_HDPI_SUPPORT=1

PR #865, opened 2025-08-08
Fix the path to copy pyimgui and pyimplot to the right location on all platforms

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Make sure that the path are correct for all platforms.

Fixes an issue cause by https://github.com/AcademySoftwareFoundation/OpenRV/pull/858.
I forgot to take into account the other platforms. Part of the changes was not limited to windows only.

Windows


cedrik-fuoco-adsk:
The path for MacOS and Linux should be the same as before my previous PR, and Windows is using the new path.

bernie-laberge (approved):
LGTM

PR #866, opened 2025-08-11
Do not build libpng as a framework on MacOS

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Prevent libpng from building as a framework on MacOS.

Since the update to libpng 1.6.48, libpng builds as a framework on MacOS. `PNG_FRAMEWORK` will be passed to keep the same behavior as before the version update.

MacOS


bernie-laberge (approved):
LGTM

PR #868, opened 2025-08-12
SG-39984: Fix duration and startFrame paint properties

eloisebrosseau:
### Summarize your change.

Change fallback values of duration and startFrame when the properties are not defined on the paint node. For duration, that value should have been 1 instead of 0. For startFrame, the frame value is access through the pen component that is format as follows: `"pen:strokeId:frameNumber:user"`. Even if it wasn't causing an issue, I changed the `ghostBefore` and `ghostAfter` values to 5 instead of 0 to match what the SessionIPNode is setting since 0 doesn't make sense here.

The `duration` and `startFrame` properties were added to know when and for how long to display held and ghosted annotations as well as calculate the opacity level for ghosted annotations. However, these two properties doesn't exist in previous version of RV. This is causing an issue, because annotations saved in an older RV session won't have these new properties and the hold and ghost feature will not be displayed properly.The fallback values are now taking care of that case.

Enabling hold and ghost on an RV session from RV 2024.2.1 in RV 2025.0.0 was tested on macOS.

### DIFF
--- a/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
+++ b/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
@@ -488,22 +488,27 @@ namespace IPCore
                 sessionNode->property<IntProperty>("paintEffects",
                                                    "ghostAfter");
 
-            m_paintEffects.hold =
-                (holdProperty != nullptr && holdProperty->size() != 0)
-                    ? holdProperty->front()
-                    : 0;
-            m_paintEffects.ghost =
-                (ghostProperty != nullptr && ghostProperty->size() != 0)
-                    ? ghostProperty->front()
-                    : 0;
-            m_paintEffects.ghostBefore = (ghostBeforeProperty != nullptr
-                                          && ghostBeforeProperty->size() != 0)
-                                             ? ghostBeforeProperty->front()
-                                             : 5;
-            m_paintEffects.ghostAfter = (ghostAfterProperty != nullptr
-                                         && ghostAfterProperty->size() != 0)
-                                            ? ghostAfterProperty->front()
-                                            : 5;
+            if (holdProperty != nullptr && holdProperty->size() != 0)
+            {
+                m_paintEffects.hold = holdProperty->front();
+            }
+
+            if (ghostProperty != nullptr && ghostProperty->size() != 0)
+            {
+                m_paintEffects.ghost = ghostProperty->front();
+            }
+
+            if (ghostBeforeProperty != nullptr
+                && ghostBeforeProperty->size() != 0)
+            {
+                m_paintEffects.ghostBefore = ghostBeforeProperty->front();
+            }
+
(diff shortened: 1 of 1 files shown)


src/lib/ip/IPBaseNodes/PaintIPNode.cpp
```diff
@@ -9,17 +9,47 @@
...
+
+        std::string property;
+        std::vector<std::string> properties{};
+
+        while (std::getline(stringStream, property, ':'))
+        {
+            properties.push_back(property);
+        }
+
+        if (properties.size() > 2)
```
bernie-laberge, line 38: It compiles even if not all paths return a value ?

src/lib/ip/IPBaseNodes/PaintIPNode.cpp
```diff
@@ -467,11 +497,11 @@ namespace IPCore
             m_paintEffects.ghostBefore = (ghostBeforeProperty != nullptr
                                           && ghostBeforeProperty->size() != 0)
                                              ? ghostBeforeProperty->front()
-                                             : 0;
+                                             : 5;
```
bernie-laberge: Perhaps we could attach this numerical constant to the IPCore::PaintEffects struct ?
```
    struct PaintEffects
    {
        static constexpr int DEFAULT_GHOST_FRAMES = 5;
```

src/lib/ip/IPBaseNodes/PaintIPNode.cpp
eloisebrosseau: Oups sorry about that. I accidentally removed `return 0` when I added the check. It was compiling, but there was a warning I missed

bernie-laberge: Crazy that this would even compile :-) 
I had never realized that it could compile successfully. My mind is blown !

According to chatGPT:
```
If you omit the return statement, like this:
int exampleFunction() {
    // Code logic here
    // No return statement
}
This will lead to undefined behavior
```

src/lib/ip/IPBaseNodes/PaintIPNode.cpp
eloisebrosseau: `ghostBefore` and `ghostAfter` are already defined with a default value in the struct:

```
struct PaintEffects
{
        int hold = 0;
        int ghost = 0;
        int ghostBefore = 5;
        int ghostAfter = 5;
};
```

I can change the check to only set the value if the property is set to something instead of a ternary if you prefer

bernie-laberge: Sure. Good idea. Whatever works to prevent the duplication of the numerical constant 5 in two different places in the code. 
Thank!

bernie-laberge (approved):
LGTM
Thank you @eloisebrosseau !

PR #873, opened 2025-08-14
SG-40023: Fix Clone Sync

eloisebrosseau:
### Summarize your change.

Fix formatting issue in the print statement for debugging in sync.mu

This line was changed a few times in the last months with the addition of Qt6 in RV. The resulting line is a mix of two previous fixes and a warning of a missing '"' was now present on this line. This format issue was causing syntax errors preventing Clone Sync from being enabled correctly. \" are not necessary and this is actually how the line was before the previous changes.

Launching Clone Sync on macOS


bernie-laberge (approved):
LGTM

PR #877, opened 2025-08-18
SG-39966: Fix annotation color selection

eloisebrosseau:
### Summarize your change.

Replace QColorDialog `currentColorChanged` signal with `colorSelected`.

`currentColorChanged` is a signal emitted whenever the color is changed in the dialog. This means that "Ok" and "Cancel" had no effect since the color was already changed as soon as it was selected. `colorSelected` is emitted just after "Ok" is clicked to set the color to use. That way, the color is only set after the user has clicked on "Ok", quitting the dialog or clicking on "Cancel" will not save the selected color.

Selecting colors through the QColorDialog was tested on macOS.


cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
LGTM

Issue #881, opened 2025-08-26
[Bug]: rvcfg command fails to enable non-free codec ac3


williamwira:
### What happened?

Building RV with the rvcfg option is supposed to enable the non-free audio codec `ac3`, but it doesn't:
```rvcfg -DRV_FFMPEG_NON_FREE_DECODERS_TO_ENABLE="ac3"```

### List all the operating systems versions where this is happening

Linux RHEL 9

### On what computer hardware is this happening?

HP Z8 G4 Workstation, Graphics: nVidia Quadro P5000, Procs: 72 Intel Xeon Gold 6154 3.0G , RAM: 191 GB

### Relevant console log output

```shell
When specifying other codecs to enable in the `rvcfg` command, e.g. `aac`, this is printed in stdout when building:

-- FFmpeg decoder aac enabled

However, there is no equivalent message for `ac3`
```

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

PR #882, opened 2025-08-26
Fix rvcfg process to allow enabling of ac3 non-free codec (#881)

williamwira:
### Summarize your change.
Added ac3 to the list of available non-free codecs in NON_FREE_DECODERS_TO_DISABLE that could
be overridden by rvcfg command

Enabling ac3 via this command did not enable ac3 at build time: rvcfg -DRV_FFMPEG_NON_FREE_DECODERS_TO_ENABLE="ac3"

Tested building with ac3 support on Linux RHEL 9.4

### DIFF
--- a/cmake/dependencies/ffmpeg.cmake
+++ b/cmake/dependencies/ffmpeg.cmake
@@ -279,6 +279,7 @@ IF(NOT RV_FFMPEG_CONFIG_OPTIONS)
       "aac_at"
       "aac_fixed"
       "aac_latm"
+      "ac3"
       "bink"
       "binkaudio_dct"
       "binkaudio_rdft"


bernie-laberge:
Hello @williamwira,

I am really sorry for not looking at it earlier but we have been swamped with the RV 2025 release which is done now.
Now I tested it (without your fix) and I don't think that your fix is required. I believe that this rvcfg is sufficient to enable AC3 audio decoding:
`rvcfg -DRV_FFMPEG_NON_FREE_DECODERS_TO_ENABLE="ac3" 
`
I tested it on macOS and it works. ie the rvcfg command line ☝️ is sufficient to enable AC3 decoding in Open RV.

**Rationale:**
That's because the AC3 audio decoding is enabled by default in FFmpeg. 
In RV (even in commercial the comercial version), the AC3 codec is disabled but not at the FFmpeg level, it is blocked at this level:
https://github.com/AcademySoftwareFoundation/OpenRV/blob/87b2024f71e854be295279bac6b7cb4c5b6d340c/src/lib/image/mio_ffmpeg/init.cpp#L36

I believe the original RV creators did that because at the time AC3 was a patent protected codec and required expensive royalties to be paid. It looks like the patent protection expired in 2017. I will try to get come legal confirmation on that so that we can enable AC3 audio by default in Open RV.

Are you sure that your PR is required ?

williamwira:
Hi @bernie-laberge , I can confirm on MacOS I do see `--FFmpeg decoder ac3 enabled` when running the rvcfg command, so I agree this is not needed for Mac.

I don't see that enabled message on Linux RHEL9 without this update, however -- so that's why I made this update. I'm using this command:
```
rvcfg -DRV_FFMPEG_NON_FREE_DECODERS_TO_ENABLE="aac;ac3;hevc;mpeg2video;prores;prores_aw;prores_ks;prores_lgpl"
```
and I only see this in the Linux output:
```
-- FFmpeg decoder aac enabled
-- FFmpeg decoder prores enabled
```

Could you let me know if it works for you there? Thanks!!

bernie-laberge:
You are absolutely right: the message is missing in action for ac3 (even though it is not doing anything since ac3 is enabled by default).
Let's go with your PR then. Everything will be much more explicit.
Thanks!

eloisebrosseau (approved):

PR #884, opened 2025-08-27
SG-40116: Fix removal of EnvVar prefix to filenames by broadcasting unmodified paths

pbergeron-adsk:
### Fix removal of environment variable prefix by broadcasting unmodified paths.

This is a tentative fix for SG-40116

Labeled as "**DO NOT MERGED - BLOCKED**" until customer validates this PR.

Customer reported that they prefix their filenames by an environment variable, but that when broadcasting the filenames the env var had been replaced by the expanded version. Remote computers receiving the filenames thus can't resolve the files locally.

### Summarize your change.

In addSources and addSourcesVerbose, send the raw, unmodified filename to the broadcast remote command. 

Let's hope this fixes it!

Not tested; just built. Needs customer validation.

N/A

### DIFF
--- a/src/lib/app/RvApp/CommandsModule.cpp
+++ b/src/lib/app/RvApp/CommandsModule.cpp
@@ -1377,15 +1377,17 @@ namespace Rv
         string tag = (muTag) ? muTag->c_str() : "";
 
         vector<string> sargs;
+        vector<string> sargsRemote;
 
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
I believe that should do the trick.
Thank you @pbergeron-adsk !

cedrik-fuoco-adsk (approved):

PR #887, opened 2025-08-28
Add fallback to search for timecode in FFmpeg format context metadata

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
When the streams metadata do not have the timecode information, I added a fallback mechanism that looks at the format context metadata and try to find the `timecode` information.

During testing, there were some MXF media files that had the wrong start time.

MacOS

### DIFF
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -1582,6 +1582,40 @@ namespace TwkMovie
         return true;
     }
 
+    void MovieFFMpegReader::populateTimecodeMetadata(const AVDictionaryEntry* tcEntry, const AVTimecode& avTimecode)
+    {
+        ostringstream tcStart, tcFR, tcFlags;
+        tcStart << tcEntry->value << " (" << avTimecode.start << ")";
+        m_info.proxy.newAttribute("Timecode/Start", tcStart.str());
+        tcFR << avTimecode.fps;
+        m_info.proxy.newAttribute("Timecode/FrameRate", tcFR.str());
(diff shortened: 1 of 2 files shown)


src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -1631,31 +1659,34 @@ namespace TwkMovie
...
+        }
 
+        // Fallback: check format-level metadata for timecode if not found in any stream
+        if (!foundTimecode && m_avFormatContext->metadata)
+        {
+            AVDictionaryEntry* fmtTcEntry = av_dict_get(m_avFormatContext->metadata, "timecode", NULL, 0);
+            if (fmtTcEntry && fmtTcEntry->value)
+            {
+                // Try to get a valid frame rate the same way as the standard mechanism above.
+                AVRational tcRate = {25, 1}; // Default to 25fps
```
bernie-laberge: Perhaps we should change the default to 24 as it is the default fps used by RV:
https://aswf-openrv.readthedocs.io/en/latest/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-six.html#aside-what-is-the-fps-of-an-rvfilesource-object-anyway
`If there is no playback FPS on startup either via the command line or preferences, it will default to 24 fps.`

And according to AI:
`The standard "cinematic" frame rate in movie production is 24 frames per second (fps)`
:-)

bernie-laberge (approved):
LGTM
(I think it would be nice to change the default fps to 24)

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
cedrik-fuoco-adsk: done!

PR #895, opened 2025-09-03
SG-40198: Prep work for testing automation of concurrent annotations

pbergeron-adsk:
### Linked issues

This PR Addresses: SG-40198

### Summarize your change.

This PR addresses 2 issues found in OpenRV to support Multiuser control of annotations.  

First, when running/testing locally, the username is the same and the annotation ID is potentially the same, causing potential prop name conflicts. The prop name has the process id attached, to help differentiate such properties. There used to be a "computer" tag that went unused (a bug) which at the time aimed to do the same thing, but was never implemented properly.

Second, we now support events that draw on the viewport in NDC (normalized display coordinates) space. This allows annotation coords to be specified in a normalized [-1,1] space, which is much easier to use from a plugin (I'll spare the unfortunate details of trying to use absolute "pixel" or "event" coordinates). These functions aren't meant to be used for real live production purposes, they are there mainly for testing automation and scripting simultaneous strokes done in multiple RVs running locally on a single machine, because we don't have 8 hands and 4 mice ;)

Third, again for the purpose of drawing shapes automatically, we allow setting a color, eiher RGB or HSV.  HSV is simpler to use because we can just change the hue and leave the color quite bright and visually distinctive.

Automating the testing and bug fixing of concurrent strokes for live review.

Tested multiple clients on macos

Not at this stage

### DIFF
--- a/src/lib/ip/IPMu/CommandsModule.cpp
+++ b/src/lib/ip/IPMu/CommandsModule.cpp
@@ -1807,7 +1807,8 @@ namespace IPMu
 
     NODE_IMPLEMENTATION(ndc2event, Mu::Vector2f)
     {
-        // Convert from NDC space [-1, 1] to event space coordinates.
+        // Convert from Normalized Device Coordinates (NDC) space [-1, 1] to
+        // event space coordinates.
         // This bypasses image transforms and works purely in viewport space.
         // Height spans 2 NDC units, and assumes the viewport fully contains the
         // height of the image (eg: has vertical bars, no horizontal bars).
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -905,12 +905,10 @@ class: AnnotateMinorMode : MinorMode
         }
     }
 
-    method: push (void; Event event)
+    method: pushStrokeCommon (void; string name, Point ip, Event event = nil, bool supportPressure = true)
     {
-        if (filterLiveReviewEvents()) {
-            sendInternalEvent("live-review-blocked-event");
-            return;
-        }
+        if (name == "") return;
+
         runtime.gc.disable();
 
         // Stop playback if we were playing.
@@ -923,10 +921,7 @@ class: AnnotateMinorMode : MinorMode
         }
 
         updateCurrentNode();
-        let (name, ip) = pointerLocation(event),
-            d = _currentDrawMode;
-
-        if (name == "") return;
+        let d = _currentDrawMode;
 
         if (_syncWholeStrokes) beginCompoundStateChange();
 
@@ -957,13 +952,16 @@ class: AnnotateMinorMode : MinorMode
 
         bool useWidth = false;
 
-        case (_currentDrawMode.pressureMode)
+        if (supportPressure)
         {
-            Size -> { useWidth = true;}
-            _ -> { useWidth = false; }
+            case (_currentDrawMode.pressureMode)
+            {
+                Size -> { useWidth = true;}
+                _ -> { useWidth = false; }
+            }
         }
 
-        if (useWidth)
+        if (useWidth && event neq nil)
         {
             let p = event.pressure(),
                 p2 = math_util.clamp(p, .01, 1.0),
@@ -975,6 +973,17 @@ class: AnnotateMinorMode : MinorMode
         {
             addToStroke(pei);
         }
+    }
+
+    method: push (void; Event event)
+    {
+        if (filterLiveReviewEvents()) {
+            sendInternalEvent("live-review-blocked-event");
+            return;
+        }
+
+        let (name, ip) = pointerLocation(event);
+        pushStrokeCommon(name, ip, event, true /*supportPressure*/);
     }
 
     method: checkDragFilter (bool; Event event, Point ip)
@@ -1004,24 +1013,16 @@ class: AnnotateMinorMode : MinorMode
         return true;
     }
 
-    method: drag (void; Event event)
+    method: dragCommon (void; string name, Point ip, Event event, bool supportPressure )
     {
-        if (filterLiveReviewEvents()) {
-            sendInternalEvent("live-review-blocked-event");
-            return;
-        }
-
         let d = _currentDrawMode;
         _pointerGone = false;
 
-        if (_currentDrawObject eq nil)
-        {
-            //
-            //  This can happen on the mac when "clicking through"
-            //
-            push(event);
-        }
-        else
+        if (name == "") return;
+
+        let pei = eventToImageSpace(name, ip, true);
+
+        if (supportPressure)
         {
             bool useWidth = false;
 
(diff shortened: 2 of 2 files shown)


src/lib/ip/IPMu/CommandsModule.cpp
```diff
@@ -1805,6 +1805,48 @@ namespace IPMu
         NODE_RETURN(mp);
     }
 
+    NODE_IMPLEMENTATION(ndc2event, Mu::Vector2f)
+    {
+        // Convert from NDC space [-1, 1] to event space coordinates.
```
bernie-laberge: Detail: Perhaps we could just define the acronym once here : "Normalized Device Coordinates (NDC)". 
You know what I think of acronyms. :-)

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1038,6 +1072,158 @@ class: AnnotateMinorMode : MinorMode
         if (_syncWholeStrokes) endCompoundStateChange();
         undoRedoUpdate();
         redraw();
+
+    }
+
+    method: color_rgb (void; Event event)
```
bernie-laberge: setColorRGB() ?
To conform with the rest of the file since it uses camel case for methods if I am not mistaken.

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1038,6 +1072,158 @@ class: AnnotateMinorMode : MinorMode
...
+        {
+            let r = float(parts[0]),
+                g = float(parts[1]),
+                b = float(parts[2]);
+
+            setColor(Color(r,g,b,1));
+        }
+    }
+
+    method: color_hsv (void; Event event)
```
bernie-laberge: setColorHSV() ?
To conform with the rest of the file since it uses camel case for methods if I am not mistaken.

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1038,6 +1072,158 @@ class: AnnotateMinorMode : MinorMode
...
+    //
+    // These functions mimic the regular push/drag/release handlers, but
+    // are used to test concurrent drawing of annoations in RV in a live 
+    // review session. We draw in a normalized [-1,1] space where (0,0) 
+    // is the center of the image.
+    // 
+    method: ndc_push (void; Event event)
+    {
+        let parts = event.contents().split(";;");
+        if (parts.size() >= 2)
```
bernie-laberge, line 1117: early return maybe instead ?

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1038,6 +1072,158 @@ class: AnnotateMinorMode : MinorMode
...
+            let x = float(parts[0]),
+                y = float(parts[1]);
+            
+            // Get source name and image point using our new method
+            let (name, ip) = pointerLocationFromNDC(Point(x, y));
+            
+            if (name == "") 
+                return;
+            
+            runtime.gc.disable();
```
bernie-laberge: It is unfortunate that we are duplicating the code of push() from this point.
Because every time we modify push(), we will have to remember to modify this one as well...
Plus we are not really exercising the same code as the one used by the real annotation.
Perhaps we could extract the push internal functionality inside its own sub routine and call it at both places ?

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1038,6 +1072,158 @@ class: AnnotateMinorMode : MinorMode
...
+        let parts = event.contents().split(";;");
+        if (parts.size() >= 2)
+        {
+            let x = float(parts[0]),
+                y = float(parts[1]);
+            
+            // Get source name and image point using our new method
+            let (name, ip) = pointerLocationFromNDC(Point(x, y));
+            
+            let d = _currentDrawMode;
```
bernie-laberge: code duplication... Evil :-)

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1038,6 +1072,158 @@ class: AnnotateMinorMode : MinorMode
...
+                let pei = eventToImageSpace(name, ip, true);
+                addToStroke(pei);
+            }
+        }
+    }
+
+    method: ndc_release (void; Event event)
+    {
+        let parts = event.contents().split(";;");
+        if (parts.size() >= 2)
```
bernie-laberge, line 1200: early return would be nice

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1038,6 +1072,158 @@ class: AnnotateMinorMode : MinorMode
...
+        {
+            let x = float(parts[0]),
+                y = float(parts[1]);
+            
+            // Get source name and image point using our new method
+            let (name, ip) = pointerLocationFromNDC(Point(x, y));
+            
+            runtime.gc.enable();
```
bernie-laberge: code duplication will cause us headaches in the long run

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1816,7 +2002,7 @@ class: AnnotateMinorMode : MinorMode
     method: AnnotateMinorMode (AnnotateMinorMode; string name)
     {
         _textPlacementMode = false;
-        _machine           = "computer";
+        _processId         = QApplication.applicationPid();
```
bernie-laberge, line 1971: Is this going to work with rvio ?
As long as it is also supported for QCoreApplication then we should be fine:
https://github.com/AcademySoftwareFoundation/OpenRV/blob/87b2024f71e854be295279bac6b7cb4c5b6d340c/src/bin/imgtools/rvio/main.cpp#L1042

src/plugins/rv-packages/annotate/annotate_mode.mu
pbergeron-adsk: I tried this first indeed and it was really kinda messy, but I could try again.

src/plugins/rv-packages/annotate/annotate_mode.mu
pbergeron-adsk: yes

bernie-laberge (approved):
LGTM !
Thank you @pbergeron-adsk !

src/lib/ip/IPMu/CommandsModule.cpp
pbergeron-adsk: done

src/plugins/rv-packages/annotate/annotate_mode.mu
pbergeron-adsk: fixed

src/plugins/rv-packages/annotate/annotate_mode.mu
pbergeron-adsk: fixed

src/plugins/rv-packages/annotate/annotate_mode.mu
pbergeron-adsk: fixed

src/plugins/rv-packages/annotate/annotate_mode.mu
pbergeron-adsk: fixed

Issue #894, opened 2025-09-03
[Bug]: Export to Snapshot does not work


jakubjezek001:
### What happened?

Currently, when an export to Snapshot is used, an empty black frame is saved with the size of the current viewport geometry. This is not suitable for automation. We are currently working on a tool that will automatically snap the current viewport if an annotation is drowned in an image. The reason for this is that we are unable to use OpenRV 3.0.0. 

Maybe there's a different way to deal with this problem?

I was already considering using Export Annotated Frames to only get the frame where the playhead is located, but it is too slow to be used as an event processor. All annotated frames are always saved and each frame takes some time to render.

We are only able to see this bug from version 3.0.0.

### List all the operating systems versions where this is happening

windows 11, MacOS Sonoma

### On what computer hardware is this happening?

Windows Intel

### Relevant console log output

```shell
No console output.
```

### Extra information
Tested on our own builds:
OpenRV_3.0.0+d5e4a1e_win64
OpenRV_3.0.0+f85b7da-win64

Test output from MacOS
<img width="1865" height="850" alt="Image" src="https://github.com/user-attachments/assets/8ac57a27-1d26-4983-923e-0a3aab9ac52e" />

Test output from Windows:
Notice that we are using `rv.commands.exportCurrentFrame` python function which is doing the same as `File/Export/Snapshot`. This is used to create temporarily the image every time an drawing tool is uses and user stops drawing. This way we are attaching the image to an annotation post you can see at the end of arrows chain. 

<img width="1550" height="924" alt="Image" src="https://github.com/user-attachments/assets/458c03e5-884c-48fe-a96f-642e16ee2d1c" />

### Code of Conduct

- [x] I agree to follow this project's Code of Conduct

philippe-ynput:
I would like to add this is blocking our and our clients' adoption of OpenRV 3.0.0.
Thanks for your time.

cedrik-fuoco-adsk:
@philippe-ynput @jakubjezek001 Letting you know that there is a PR to fix your issue with the snapshot feature. It will be merged once the review and testing are done.

https://github.com/AcademySoftwareFoundation/OpenRV/pull/901

cedrik-fuoco-adsk:
Fixed by #901 

If it is possible, could you rebuild from the main branch and test it out @jakubjezek001 @philippe-ynput ?

Feel free to re-open if any other issues occurs.

PR #901, opened 2025-09-09
Fix Export > Snapshot feature

cedrik-fuoco-adsk:
### Linked issues
#894 

### Summarize your change.

Fixed OpenGL framebuffer state management in `exportCurrentFrame` by removing unused FBO creation and adding explicit framebuffer binding. The function now properly saves and restores Qt's framebuffer state, and explicitly binds the device's default framebuffer for glReadPixels() to ensure consistent behavior.

The Export > Snapshot menu option was broken with Qt 6.

The original implementation was causing `GL_ERROR: invalid framebuffer operation errors` in GLView.cpp due to improper OpenGL framebuffer state management. The code was creating a custom FBO via renderer->newOutputOnlyImageFBO() but never actually using it for pixel reading. Instead, it read pixels from whatever framebuffer happened to be bound at the time, then called fbo->unbind() on the unused custom FBO.

This was cause issue because the context was bind to the framebuffer id 0, and we cannot use that with QOpenGL* widgets. We have to use the default framebuffer object (which is not 0 with QOpenGL* widgets).

MacOS

### DIFF
--- a/src/lib/app/RvApp/CommandsModule.cpp
+++ b/src/lib/app/RvApp/CommandsModule.cpp
@@ -550,11 +550,16 @@ namespace Rv
             d->width() - m.left - m.right, d->height() - m.bottom - m.top, 4,
             TwkFB::FrameBuffer::UCHAR);
 
-        const ImageRenderer::GLFBO* fbo =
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
LGTM

PR #903, opened 2025-09-11
Add support for up to 32 channels audio channels

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Added support for 9, 11, 12, 13, 14 and 15 audio channels.

**Review the channels layout carefully as I am not an expert on audio layouts.**

Added more audio channels count because they were not handled.

I tested on MacOS with an 9-channels audio file.


src/lib/audio/TwkAudio/AudioFormats.cpp
```diff
@@ -273,12 +291,30 @@ namespace TwkAudio
         case 8:
             lv.push_back(Surround_7_1);
             lv.push_back(Back_7_1);
-            break;
+            // break;
```
bernie-laberge: Good catch ! 
Let's remove the commented out break altogether

cedrik-fuoco-adsk:
Added support to up to 32 channels.

eloisebrosseau (approved):

PR #904, opened 2025-09-12
SG-40182: Finer granularity over Multiuser Control Filtering

pbergeron-adsk:
### Linked issues

Adresses core requirement of SG-39429

### Summarize your change.

In the context of Multiuser Control for Review, need to allow more granular control over what events are allowed to flow through. It used to be that when joining a session as a participant, a setLiveReviewFilter(true) was called to disable multiple RV menu items, commands, and functionality. Participants were observers only. 

A new requirement is that the presenter (aka the host for now) may grant participants permission to perform annotations, and to play with transport controls, and in the future, multiple other permission types may be granted (eg: allowed to change the view mode or rendering mode). For this reason, a global filter flag was no longer sufficient, we needed arbitrary categories for what permissions may be granted to the participant.

This is what this PR aims to do.

It removes the mu function (set)filterLiveReviewEvents() and replaces it with (add)FilterEventCategory to define arbitrary filtered categories. The PR also reclassifies all calls to filterLiveReviewEvents() to the new filterEventCategory(category_name, notify=true) call.  At the same time, we've removed the explicit broadcast of event that is caught by the live review package to show "you are not the presenter", and unless notify is set to false, the event is dispatched within the filterEventCategory() call.

As per above. Needed to categories event filters for multi user control.

MacOS, two RVs together; behaves like it used to when also running modified live review package.

NOTE: This breaks backwards compatibility with older RVs live review packages, since a crucial mu function has disappeared and has been replaced.

n/a

### DIFF
--- a/src/lib/ip/IPCore/Session.cpp
+++ b/src/lib/ip/IPCore/Session.cpp
@@ -423,6 +423,7 @@ namespace IPCore
         , m_waitingOnSync(false)
         , m_avPlaybackVersion(DEFAULT_AVPLAYBACK_VERSION)
         , m_lastDrawingTime(0)
+        , m_filteredEventsEnabled(false)
     {
         if (!m_graph)
             m_graph = new IPGraph(App()->nodeManager());


src/lib/app/RvApp/CommandsModule.cpp
```diff
@@ -1377,6 +1394,9 @@ namespace Rv
     NODE_IMPLEMENTATION(addSources, void)
     {
         RvSession* s = RvSession::currentRvSession();
+        if (filterMediaChanges(s))
```
bernie-laberge: I think we should add the same protection for the clearSession() command in OpenRV/src/lib/ip/IPMu/CommandsModule.cpp.
Otherwise participants could wipe the session with the keyboard shortcut CMD+SHIFT+N

pbergeron-adsk: YES! I had this noted for in a 2nd phase. For this PR, I wanted to just "port" the existing state. In the 2nd phase, I will want to address all of the other cases that cause problems, including clear session.  (but, I'll fix it here since you asked)

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
```diff
@@ -502,7 +502,8 @@ namespace Rv
                 bool isChecked = (contents == "1");
                 m_ghostAction->setChecked(isChecked);
 
-                if (m_session->filterLiveReviewEvents())
+                if (m_session->filterEventCategory("review_nonhost",
```
bernie-laberge: "hold_and_ghost" instead of "review_nonhost"
I wonder if we shouldn't use some generic category here. The review_nonhost is too closely linked with the business logic of the Live Review session with respect to participant.
Proposition: How about we use "hold_and_ghost" insted of "review_nonhost" here ?
This way if the Live Review plugin wants to restrict access to hold_and_ghost then it will add this filter explictely. The Open RV code does not have to know (and shouldn't know) the Live Review business logic.

src/lib/app/RvCommon/RvTopViewToolBar.cpp
```diff
@@ -566,7 +567,8 @@ namespace Rv
             else if (name == "internal-sync-presenter-changed"
                      || name == "sync-session-ended")
             {
-                bool isDisabled = m_session->filterLiveReviewEvents();
+                bool isDisabled = m_session->filterEventCategory(
```
bernie-laberge: This whole section should be removed and handled at the plugin level. 
Open RV should not know about the Live Review plugin

src/lib/app/mu_rvui/extra_commands.mu
```diff
@@ -317,8 +317,7 @@ require system;
 
 \: toggleForwardsBackwards (void;)
 {
-    if (filterLiveReviewEvents()) {
-        sendInternalEvent("live-review-blocked-event");
+    if (filterEventCategory("review_transport")) {
```
bernie-laberge: I would rename it from "review_transport" to "transport".
Open RV only needs to know that this operation is in the "transport" category. So if a plugin wants to filter these events then they can do so. Even if the plugin is not a review related plugin.

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
```diff
@@ -512,15 +513,17 @@ namespace Rv
                 bool isChecked = (contents == "1");
                 m_holdAction->setChecked(isChecked);
 
-                if (m_session->filterLiveReviewEvents())
+                if (m_session->filterEventCategory("review_nonhost",
```
bernie-laberge: "hold_and_ghost" instead of "review_nonhost"
This way it is what it is. If a plugin wants to filter then they they can do so.

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
```diff
@@ -512,15 +513,17 @@ namespace Rv
...
+                                                   false /*notify*/))
                 {
                     m_holdAction->setDisabled(true);
                 }
             }
             else if (name == "internal-sync-presenter-changed"
                      || name == "sync-session-ended")
             {
-                bool isDisabled = m_session->filterLiveReviewEvents();
+                bool isDisabled = m_session->filterEventCategory(
```
bernie-laberge: This whole section should be removed and handled at the plugin level.
Open RV should not know about the Live Review plugin

src/lib/app/mu_rvui/extra_commands.mu
```diff
@@ -327,8 +326,7 @@ require system;
 
 \: toggleRealtime (void;)
 {
-    if (filterLiveReviewEvents()) {
-        sendInternalEvent("live-review-blocked-event");
+    if (filterEventCategory("review_transport")) {
```
bernie-laberge: I would rename it from "review_transport" to "transport".

src/lib/app/mu_rvui/extra_commands.mu
```diff
@@ -395,8 +393,7 @@ require system;
 
 \: toggleFilter (void;)
 {
-    if (filterLiveReviewEvents()) {
-        sendInternalEvent("live-review-blocked-event");
+    if (filterEventCategory("review_nonhost")) {
```
bernie-laberge: I would rename it from "review_transport" to "transport".
etc...

src/lib/app/RvCommon/RvTopViewToolBar.cpp
pbergeron-adsk: why?

src/lib/app/mu_rvui/extra_commands.mu
pbergeron-adsk: so do you want me to remove the review_ prefix from all categories?

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
pbergeron-adsk: "nonhost" est le token pour dire: "not the host" (presenter). Autrement dit, le host a le droit de faire l'action. ca n'a pas rapport a hold and ghost.

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
pbergeron-adsk: je crois ici que le but etait que le BottomViewToolbar puisse faire quelque chose d'intelligent en reponse a un changement de presentateur. 

Mon but ici n'est pas de tout revoir le code qui existait avant, c'est juste de mettre des categories. Enlever ce check ici devrait etre une autre story de maintenance.

src/lib/app/RvApp/CommandsModule.cpp
pbergeron-adsk: fixed, but also changed approach; closing conversation/comment please redo another code review pass next week (this is still in a wip state)

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
pbergeron-adsk: changed approach; closing conversation/comment please redo another code review pass next week (this is still in a wip state)

src/lib/app/mu_rvui/extra_commands.mu
pbergeron-adsk: fixed, but also changed approach; closing conversation/comment please redo another code review pass next week (this is still in a wip state)

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
pbergeron-adsk: changed approach; closing conversation/comment please redo another code review pass next week (this is still in a wip state)

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
pbergeron-adsk: changed approach; closing conversation/comment please redo another code review pass next week (this is still in a wip state)

src/lib/app/mu_rvui/extra_commands.mu
pbergeron-adsk: fixed, but also changed approach; closing conversation/comment please redo another code review pass next week (this is still in a wip state)

src/lib/app/mu_rvui/extra_commands.mu
pbergeron-adsk: fixed, but also changed approach; closing conversation/comment please redo another code review pass next week (this is still in a wip state)

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
```diff
@@ -512,15 +512,15 @@ namespace Rv
...
+                if (m_session->isFilteredEvents(false /*notify*/))
                 {
                     m_holdAction->setDisabled(true);
                 }
             }
             else if (name == "internal-sync-presenter-changed"
                      || name == "sync-session-ended")
             {
-                bool isDisabled = m_session->filterLiveReviewEvents();
+                bool isDisabled = m_session->isFilteredEvents(false /*notify*/);
```
bernie-laberge, line 523: How do you discriminate between the Live Review with Multiuser Control (where the Annotation tool must be enabled) and without Multiuser Control (where the Annotation tool must be disabled ?)

src/lib/app/mu_rvui/extra_commands.mu
```diff
@@ -317,8 +317,7 @@ require system;
 
 \: toggleForwardsBackwards (void;)
 {
-    if (filterLiveReviewEvents()) {
-        sendInternalEvent("live-review-blocked-event");
+    if (isFilteredEventsAllow("transport")) {
```
bernie-laberge, line 320: I think we need to rename isFilteredEventsAllow() because by first reading it I see:
if the filtered events allow "transport" then return true.
Which is the reverse of what it is here.

bernie-laberge: I would just use isFilteredEvent() (instead of isFilteredEventsAllow()) and pass the category as the first parameter:
isFilteredEvent(string category="", bool notify=true)
Which would read:
If this is a filtered event then return right away.

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -801,6 +817,11 @@ global Configuration globalConfig =
 
 \: canExportOTIOState (int;)
 {
+    if (isFilteredEvents(false /*notify*/))
```
bernie-laberge, line 820: We should be able to export the OTIO since it wil not impact the current session. This will prove usefull when debugging CR<->RV Live Review

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -747,6 +758,11 @@ global Configuration globalConfig =
 
 \: singleSourceState (int;)
 {
+    if (isFilteredEvents(false /*notify*/))
```
bernie-laberge, line 761: Good catch !

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -1076,8 +1102,9 @@ global let toggleFlip = toggleIntProp("#RVTransform2D.transform.flip"),
 {
     \: (void;)
     {
-        if (filterLiveReviewEvents()) {
-            sendInternalEvent("live-review-blocked-event");
+        // Bernard/Eloise. Is this a transport func, or a general func
```
bernie-laberge, line 1105: Setting in and out points is NOT part of the Multiuser Control spec. So maintain the restriction for all Live Review participants and remove the comment please. Thanks!

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -1087,8 +1114,8 @@ global let toggleFlip = toggleIntProp("#RVTransform2D.transform.flip"),
 \: incN ((void;); int i)
 {
     \: (void;) {
-        if (filterLiveReviewEvents()) {
-            sendInternalEvent("live-review-blocked-event");
+        // Bernard/Eloise same question as above.
```
bernie-laberge, line 1117: This is a transport event.

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -1481,8 +1505,7 @@ global let gammaMode      = startParameterMode("#RVColor.color.gamma", 4.0, Defa
 
 \: setFPSValue (void; string v)
 {
-    if (filterLiveReviewEvents()) {
-        sendInternalEvent("live-review-blocked-event");
+    if (isFilteredEventsAllow("transport")) {
```
bernie-laberge, line 1508: I don't think participants should be allowed to change the frame rate. 
It is NOT part of the multiuser control spec so I would filter this event for all.

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -1759,7 +1781,9 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: markedState (int;)
 {
-    if (filterLiveReviewEvents()) {
+    // Bernard/Eloise: do we allow this for muc?
```
bernie-laberge, line 1784: Marking frames is not in the multiuser control spec so I would maintain the original restriction for both multiuser control on and off.

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -1779,7 +1803,9 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: sequenceState (int;)
 {
-    if (filterLiveReviewEvents()) {
+    // Bernard/Eloise: is the sequence state menu item based on
+    // transport permissions or nonhost permissions?
+    if (isFilteredEventsAllow("transport", false /*notify*/)) {
```
bernie-laberge, line 1808: Marking sequence is not in the multiuser control spec so I would maintain the original restriction for both multiuser control on and off.

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -2268,11 +2296,6 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: wipeShown (int;)
 {
-    if (filterLiveReviewEvents())
```
bernie-laberge, line 2271: Why aren't we filtering wipes anymore ?

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -3195,8 +3205,7 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
     // then we are restoring the old values.
     //
 
-    if (filterLiveReviewEvents()) {
-        sendInternalEvent("live-review-blocked-event");
+    if (isFilteredEventsAllow("transport")) {
```
bernie-laberge, line 3208: Setting/Resetting in and out points is not in the multiuser control spec so I would maintain the original restriction for both multiuser control on and off.

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -3279,8 +3288,7 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 {
     \: (void;)
     {
-        if (filterLiveReviewEvents()) {
-            sendInternalEvent("live-review-blocked-event");
+        if (isFilteredEventsAllow("transport")) {
```
bernie-laberge, line 3291: Changing the FPS is not in the multiuser control spec so I would maintain the original restriction for both multiuser control on and off.

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -3897,6 +3905,10 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: clearEverything (void;)
 {
+    // Prevent participants from clearing the session
+    if (isFilteredEvents())
```
bernie-laberge, line 3909: Good !

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -4531,6 +4565,9 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: save (void; Event ev)
 {
+    if (isFilteredEvents())
```
bernie-laberge, line 4568: I think we should allow it

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -4545,6 +4582,9 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: saveAs (void; Event ev)
 {
+    if (isFilteredEvents())
```
bernie-laberge, line 4585: I think we should allow it

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -4658,6 +4701,9 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: exportAs (void; Event ev, string requiredExt, string outputType)
 {
+    if (isFilteredEvents())
```
bernie-laberge, line 4704: I think we should allow it

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -4736,6 +4782,9 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: exportFrame (void; Event ev, (void;string) outputFunc)
 {
+    if (isFilteredEvents())
```
bernie-laberge, line 4785: I think we should allow it

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -4751,6 +4800,9 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: exportAttrs (void; Event ev)
 {
+    if (isFilteredEvents())
```
bernie-laberge, line 4803: I think we should allow it

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -4775,6 +4827,9 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: exportNodeDefinition (void; bool all, Event ev)
 {
+    if (isFilteredEvents())
```
bernie-laberge, line 4830: I think we should allow it

src/lib/app/mu_rvui/rvui.mu
```diff
@@ -4932,7 +4985,8 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 
 \: isStackMode (int;)
 {
-    if (filterLiveReviewEvents()) {
+    // Bernard/Eloise: What's stack mode? is part of the transport family?
```
bernie-laberge, line 4988: It tells you if the current view node is looking at a stack. Now it is used in this context to cycle through the different sources in the stack, so it is effectively changing the current source.
This functionality is not in the multiuser control spec so I would maintain the original restriction for both multiuser control on and off.

src/lib/ip/IPCore/IPCore/Session.h
```diff
@@ -814,8 +816,18 @@ namespace IPCore
         void setGlobalAudioOffset(float, bool internal = false);
         void setGlobalSwapEyes(bool);
 
-        void setFilterLiveReviewEvents(bool shouldFilterEvents = false);
-        bool filterLiveReviewEvents();
+        // Old category-based API (to be deprecated)
```
bernie-laberge, line 819: Why can't we simply remove those 3 funcitons if they have never been merged ?

src/lib/ip/IPCore/IPCore/Session.h
```diff
@@ -814,8 +816,18 @@ namespace IPCore
...
+        void enableFilterEventCategory(const std::string& category);
+        void disableFilterEventCategory(const std::string& category);
+        bool filterEventCategory(const std::string& category, bool notify);
+
+        // New global + allow-list API
+        void setFilteredEvents(bool enabled);
+        bool isFilteredEvents(bool notify);
+        void allowFilteredEventCategory(const std::string& category);
+        void disallowFilteredEventCategory(const std::string& category);
+        bool isFilteredEventsAllow(const std::string& category,
```
bernie-laberge, line 829: As mentioned earlier, the name of this function is not natural at all. It does the reverse of what it says.

src/lib/ip/IPCore/Session.cpp
```diff
@@ -720,12 +721,100 @@ namespace IPCore
...
+    {
+        auto it = std::find(m_allowedEventCategories.begin(),
+                            m_allowedEventCategories.end(), category);
+        if (it != m_allowedEventCategories.end())
+        {
+            m_allowedEventCategories.erase(it);
+        }
+    }
+
+    bool Session::isFilteredEventsAllow(const std::string& category,
```
bernie-laberge, line 797: As mentioned earlier I would simply rename it to isFilteredEvent(const std::string& category, bool notify)

src/lib/ip/IPCore/Session.cpp
```diff
@@ -4620,29 +4709,34 @@ namespace IPCore
...
-                        || contents == "extra_commands.stepBackward(1)"
-                        || contents == "commands.setPlayMode(PlayOnce)"
-                        || contents == "commands.setPlayMode(PlayPingPong)"
-                        || contents == "commands.setPlayMode(PlayLoop)"))))
-        {
-            GenericStringEvent event("live-review-blocked-event", this, "");
-            sendEvent(event);
-            return "";
+        // Check for general events first
+        if (isFilteredEvents(false /*notify*/))
```
bernie-laberge, line 4713: I don't think this is what the code was doing before :-)
I think this will create tons of notifications for Live Review participants.

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
bernie-laberge: After going through the whole code review, I think you just need to make sure that the annotations ICON in the bottom RV toolbar is enabled for multiuser control=ON participants.

pbergeron-adsk: Good catch, I didn't notice that we had an annotation icon at the bottom. I'll fix this properly.

src/lib/ip/IPCore/IPCore/Session.h
pbergeron-adsk: good catch, old code from an old commit.

src/lib/ip/IPCore/IPCore/Session.h
pbergeron-adsk: I agree with you on the naming but I've been struggling to find good naming. 
It used to be "filteredLiveReviewEvent()" which returns true/false, with return true being filtered-out and blocked. 

To be peffectly explicit of what I want, is " **_isFilteredEventButLetPassThroughAllowedCategory("transport")_** but that name seems clumsy just as well.  

Or I could do:   **_isFilteredEventExcept("transport") { ... }_**

Note that I cannot do:   **_if (!allowEvent("transport") && isFilteredEvent()) { ... }_**

Because of the default notify parameter. if I put isFilteredEvent() to the right of the &&, then it wont even get evaluated if the left portion returns false (and it won't notify) and if I put it to the left of the && then it will notify even though I dont want it to because the 2nd part could also return true.

pbergeron-adsk: The naming may be ambiguous because "filtered" might evoke both "allow" and "disallow" depending if you think about passing through the filter (allow) or being blocked by the filter (disallow). I may just rename this isBlockedEventExceptCategory("transport") which is already better.

Finally, note that this is still only for live review. I know you asked me to remove "live review" notions from the naming,  but I'm not sure leaving it generic will convey information any better. An OpenRV contributor-only would look at this code and go: this is completely useless and unused stuff" 

At least with "live review" in the name, that ambiguity doesn't exist.

bernie-laberge: How about:
isFilteredEvent(const std::string& event_category="")
isFilteredEventNoNotify(const std::string& event_category="")
What do you think @eloisebrosseau ?
Naming things is the hardest thing to do. :-)

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: Fixed.

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: Done

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: Fixed

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: Fixed

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: Fixed

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: mistake. good catch. fixed.

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: Fixed

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: fixed

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: Fixed

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: Fixed

src/lib/ip/IPCore/IPCore/Session.h
pbergeron-adsk: Fixed

src/lib/app/mu_rvui/rvui.mu
pbergeron-adsk: Thanks

src/lib/ip/IPCore/Session.cpp
pbergeron-adsk: lets discuss

pbergeron-adsk:
close because not taking this approach

PR #907, opened 2025-09-15
SG-40348: Fix print crashing when called from non-main thread; add warning for other mu commands

pbergeron-adsk:
### Summarize your change.

Remove redirecting of python's print() command to mu's print().

When we're running live review with debug traces on, we get a ton of stuff in the console (as expected) but also sometimes RV mysteriously crashes. It crashes with the message "Unexpected State" (but may crash elsewhere in mu-ish related handlers - it may crash with "bus error", or "segmentation fault")

This "Unexpected State" message is printed from the Mu garbage collector, which gets into an invalid state, because Live Review is running multiple threads, and Mu isn't thread safe.

But what does calling print() from python have to do Mu's garbage collector? Well, it's because when we issue a python print() statement, it's actually redirected to Mu's print() command, which calls the garbage collector (again, not  thread safe), and (sometimes) crashes, but 100% crashes if you let it live review run long enough with print statements.

(The code that redirects python's print() to Mu's print() is in rv_commands_setup.py)

We looked into the possibility of just removing the redirection of python print, but, this is not an option since we need to show print statements into the RV console. 

As an alternative, we instead detect mu being called from a non-main thread, and if so, prints out a warning message (see screenshot) that Mu isn't thread safe and that this should not be allowed.

Nevertheless, we will special-case "print()" being called from any thread, because it's often used to print traces to debug python stuff (even multi-threaded python stuff). However, for print(), we don't let the mu code execute. We call cout() directly via a helper function.

Tested on macOS, but the problem appears on any OS. 

See that we detect an unsafe print() from python, and that the unsafe print (with a prefix) appears both in the console and in the terminal.

See that other mu commands (like sendInternalEvent) are detected in the wrong thread and a warning appears both in the console and in the terminal.

image

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -222,8 +222,29 @@ jobs:
       - name: Install other system dependencies
         if: ${{ matrix.rocky-version == '9' }}
         run: |
-          dnf install -y perl-CPAN
-          cpan FindBin
+          retry_count=0
+          max_retries=5
+
+          until [ $retry_count -ge $max_retries ]; do
+            echo "Attempt $((retry_count + 1)) of $max_retries"
+
+            if dnf install -y perl-CPAN && \
+              cpan FindBin; then
+              echo "Dependencies installed successfully"
+              break
+            else
+              retry_count=$((retry_count + 1))
+              if [ $retry_count -lt $max_retries ]; then
+                echo "Attempt $retry_count failed. Retrying in 5 seconds..."
+                sleep 5
+              fi
+            fi
+          done
+          
+          if [ $retry_count -ge $max_retries ]; then
+            echo "Failed to install dependencies after $max_retries attempts"
+            exit 1
+          fi
 
       - name: Check out repository code
         uses: actions/checkout@692973e3d937129bcbf40652eb9f2f61becf3332 # 4.1.7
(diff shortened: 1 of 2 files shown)


cedrik-fuoco-adsk (approved):

src/lib/app/PyTwkApp/PyMuSymbolType.cpp
```diff
@@ -45,14 +45,40 @@
...
 
+    // Thread safety tracking
+    // Default constructor creates "not-a-thread" ID
+    static boost::thread::id s_mainThreadId;
+
+    // Helper function for direct cout printing
+    static PyObject* unsafe_mu_print(PyObject* args)
+    {
+        // const char *prefix = "Unsafe Mu print: ";
+        const char* prefix = "";
```
bernie-laberge: Why do we declare prefix if it's not really used (empty string) ?

bernie-laberge (approved):
LGTM
(apart from the unused prefix which could be removed I think)

markreidvfx:
Thanks @pbergeron-adsk, I think this is a bug I've been trying to track down for ages. I have been getting random crashes if calling print from a python thread and thought it might be something like this.

cedrik-fuoco-adsk (approved):

src/lib/app/PyTwkApp/PyMuSymbolType.cpp
pbergeron-adsk: bah it was there as a helper to test if indeed it was this version of Print that was being called. No worries I removed it.

pbergeron-adsk:
> Thanks @pbergeron-adsk, I think this is a bug I've been trying to track down for ages. I have been getting random crashes if calling print from a python thread and thought it might be something like this.

No problem!

pbergeron-adsk:
Note: I will bypass view testing because this is an intermittent bug.

PR #910, opened 2025-09-16
Pin OpenTimelineIO to specific commit in python build

lithorus:
This pins the OTIO python module to a specific commit since the latest commit doesn't work with the compiled python interpreter.

### DIFF
--- a/src/build/requirements.txt
+++ b/src/build/requirements.txt
@@ -3,7 +3,7 @@
 pip                     # License: MIT License (MIT)
 setuptools              # License: MIT License
 # We need to use the main branch of the OTIO Github repo since the latest release available as a Pypi package is missing some features we need for Live Review
-git+https://github.com/AcademySoftwareFoundation/OpenTimelineIO@main#egg=OpenTimelineIO # License: Other/Proprietary License (Modified Apache 2.0 License)
+OpenTimelineIO==0.17.0  # License: Other/Proprietary License (Modified Apache 2.0 License)
 PyOpenGL                # License: BSD License (BSD)
 
 # MacOS only - PyOpenGL_accelerate is built from source in python3.cmake as a workaround until the fix
--- a/src/build/requirements.txt
+++ b/src/build/requirements.txt
@@ -3,7 +3,8 @@
 pip                     # License: MIT License (MIT)
 setuptools              # License: MIT License
 # We need to use the main branch of the OTIO Github repo since the latest release available as a Pypi package is missing some features we need for Live Review
-OpenTimelineIO==0.17.0  # License: Other/Proprietary License (Modified Apache 2.0 License)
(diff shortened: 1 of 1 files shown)


lithorus:
Looks like the Windows builds are randomly crashing for no good reason. Can you make it retry the single errored job?

cedrik-fuoco-adsk:
The windows debug build is expected to fail, but a fix was commit today. You can "Update branch" and it will restart the CI @lithorus 

I can do it if you do not see the "Update branch" button

cedrik-fuoco-adsk:
@lithorus Thanks for the contribution! But we can not pin to 0.17 because some newer commits are mandatory for the Live Review plugin. 

Do you know which commit introduce the issue you are talking about?

The best we can do is to pin to the the commit that we need from the main branch for Live review plugin.

lithorus:
> @lithorus Thanks for the contribution! But we can not pin to 0.17 because some newer commits are mandatory for the Live Review plugin.
> 
> Do you know which commit introduce the issue you are talking about?
> 
> The best we can do is to pin to the the commit that we need from the main branch for Live review plugin.

No, the only reason I discovered it was because I was re-building 2 internal version that both had the issue.

Can do some testing, but the weird thing is that it seems the OTIO has their own testing pipeline that hasn't picked up the error. Could be something to do with how OpenRV is building Python.

lithorus:
Alternatively requirements.tx would just use pypi instead of git directly. (without the pinning)

lithorus:
@cedrik-fuoco-adsk I found out that it was the absolute latest commit that breaks it.
https://github.com/AcademySoftwareFoundation/OpenTimelineIO/commit/7f1ee9ce5fab803320e849b55885249583784835

The one before doesn't have the issue and I have pinned it to that instead.

Also, I should mention that if I use my system python and install the otio module from latest master it works fine. It only breaks with the custom python build.

bernie-laberge (approved):
LGTM
Thanks you @lithorus !

lithorus:
Has anyone else been able to reproduce the import error?
Just wondering if it's a fault on my setup (using ubuntu 24.04)

bernie-laberge:
Yes we're having the exact same issue now in our internal CICD on Windows.
What we don't understand is that the OpenRV's github actions Windows build does seem to work so we are wondering if this could be fixed by using a more recent CMake (we have a build currently in progress to validate this hypothesis). If the CMake update is not successful when we'll go with your PR. Thanks !

lithorus:
FYI, I also tested with a newer version of nanobind (since it seemed related to pybind11) which didn't help either.

bernie-laberge:
Updating CMake to version 3.31.6 did not work either. So we are merging your PR @lithorus.
Thank you for your help !

PR #913, opened 2025-09-17
SG-40385: Better prompt, aliases and command-line environment for building rv

pbergeron-adsk:
### Linked issues

SG-30485 
https://jira.autodesk.com/browse/SG-40385

### Summarize your change.

This PR streamlines the usage of the rv commands at the bash or zsh shell. 

In particular:
* we no longer need "d" versions of aliases.
* after sourcing rvcmds.sh, we always have a command prompt that tells us if we're in debug or release
* if the _build(_debug) directory doesn't exists, the command prompt suggests to run rvbootstrap
* if the _build(_debug) directory does exist, the command prompt suggests to run rvcfg or rvmk
* after running rvcfg, the (.venv) command prompt also has a prefix to indicate if we're in debug or release.
* to switch from one build type to another, just call "rvrelease" or "rvdebug". The build paths and aliases will auto adjust, and the environment and command prompt will remain consistent.

**Seeing it in action**
----------------------------------------------------------------

After calling "source rvcmds.sh" - note the initial state of the command prompt.

image

----------------------------------------------------------------

Switching configs (calling rvdebug or rvrelease) back and forth will auto adjust the prompt (and aliases internally).

image

----------------------------------------------------------------

And finally, after calling rvcfg, the prompt changes to .venv keeping the context for release or debug

image

----------------------------------------------------------------

While I was at it, I added three other new rv aliases: 
* rvappdir
* rvhomedir
* rvrun

These will "cd" into the directory where the rv binary sits eg ($HOME/ARV/_build_debug/stage/app/RV.app/Contents/MacOS) or home dir ($HOME/ARV). it makes it easy to navigate to the bin dir to start rv from the command line. Obviously, "rvappdir" selects the correct _build or _build_debug dir, depending on the currently selected configuration.

as for rvrun, this will call rvappdir and start RV.

image

----------------------------------------------------------------

Well, this is a fix for a pet-peeve of mine that has been bothering me since I started working on RV. It was like a little itch too small to do anything about it, but that constantly itches. Well, I fixed the itch.

I wanted a visual cue to know if my rv cfg was release or debug, because I often need to switch to/from release/debug builds, and I often lose track of which environment I'm in at the command prompt because I have multiple shells and I don't know which I've done source rvcmds, which app dirs are release or debug (I have to do "pwd" to know), etc.

I also often mix up "rvbuild" vs "rvbuildd", triggering the wrong build for the wrong environment. It was driving me a bit nuts.

With these changes, the context for the environment, aliases, and paths are always clear. It's difficult to make a mistake.

macOS. I'd like others to test this on Windows and Linux.

see above

see above

### DIFF
--- a/docs/build_system/config_macos.md
+++ b/docs/build_system/config_macos.md
@@ -29,7 +29,11 @@ From macOS System Settings > Privacy & Security > App Management, allow Terminal
 (install_xcode)=
 ## Install Xcode
 
-From the App Store, download Xcode. Make sure that it is the source of the active developer directory.
+**Heads Up:**
+Xcode 26 isn't supported yet because it needs the fix for [QTBUG-137687](https://bugreports.qt.io/browse/QTBUG-137687). This fix is available in Qt 6.5.10, but there isn't a PySide6 version that matches this Qt version as of now.
+In the meantime, you can use Xcode 16.4 on the latest macOS Tahoe 26 to build Open RV.
+
+From the App Store, download Xcode 16.4. Make sure that it is the source of the active developer directory.
 
 `xcode-select -p` should return `/Applications/Xcode.app/Contents/Developer`. If that is not the case, run `sudo xcode-select -s /Applications/Xcode.app`
 
--- a/src/build/requirements.txt
+++ b/src/build/requirements.txt
@@ -3,7 +3,8 @@
 pip                     # License: MIT License (MIT)
 setuptools              # License: MIT License
 # We need to use the main branch of the OTIO Github repo since the latest release available as a Pypi package is missing some features we need for Live Review
-git+https://github.com/AcademySoftwareFoundation/OpenTimelineIO@main#egg=OpenTimelineIO # License: Other/Proprietary License (Modified Apache 2.0 License)
+# It is pinned to a specific commit since the latest breaks the module with the OpenRV python build.
+git+https://github.com/AcademySoftwareFoundation/OpenTimelineIO@7a76cf3c33212f3fc56464e2c016a80f6a264819#egg=OpenTimelineIO # License: Other/Proprietary License (Modified Apache 2.0 License)
 PyOpenGL                # License: BSD License (BSD)
 
 # MacOS only - PyOpenGL_accelerate is built from source in python3.cmake as a workaround until the fix
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -2311,7 +2311,7 @@ class: AnnotateMinorMode : MinorMode
         connect(_drawDock, QDockWidget.topLevelChanged, topLevelChangedSlot);
 
         connect(_colorButton, QPushButton.clicked, chooseColorSlot);
-        connect(_colorDialog, QColorDialog.currentColorChanged, newColorSlot(,true,true));
+        connect(_colorDialog, QColorDialog.colorSelected, newColorSlot(,true,true));
 
         _undoAct = QAction(auxIcon("undo_64x64.png"), "Undo", m);
         _redoAct = QAction(auxIcon("redo_64x64.png"), "Redo", m);


rvcmds.sh
```diff
@@ -105,69 +105,184 @@ else
...
+alias rvbuildt='rvenv && cmake --build ${RV_BUILD_DIR} --config ${RV_BUILD_TYPE} -v --parallel=${RV_BUILD_PARALLELISM} --target '
+alias rvbuild='rvenv && rvbuildt main_executable'
+alias rvtest='rvenv && ctest --test-dir ${RV_BUILD_DIR} --extra-verbose'
+alias rvinst='rvenv && cmake --install ${RV_BUILD_DIR} --prefix ${RV_INST_DIR} --config ${RV_BUILD_TYPE}'
+alias rvclean='__rv_clean_build'
+alias rvappdir='cd ${RV_APP_DIR}'
+alias rvhomedir='cd ${RV_HOME}'
+alias rvhome='cd ${RV_HOME}'
+alias rvmk='rvcfg && rvbuild'
+alias rvbootstrap='rvsetup && rvmk'
```
bernie-laberge, line 260: Aren't the rvdebug and rvrelease aliases missing ?

bernie-laberge: The documention needs to be changed since rvcfgd and rvbuildd no longer exist:
https://aswf-openrv.readthedocs.io/en/latest/build_system/config_common_build.html

rvcmds.sh
```diff
@@ -105,69 +105,184 @@ else
...
-echo "If this is your first time building RV try rvbootstrap (release) or rvbootstrapd (debug)"
-echo "To build quickly after bootstraping try rvmk (release) or rvmkd (debug)"
+echo "Use 'rvrelease' (default) or 'rvdebug' to switch between build configurations."
+echo
+echo "Call 'rvbootstrap' if its your first time building or after calling rvclean."
+echo "After 'rvbootstrap', use 'rvbuild' or 'rvmk' for incremental builds."
+echo
+
+# Initialize with appropriate build mode (after sourcing, if we were in debug, stay in debug)
+if [ "$INIT_BUILD_TYPE" = "Debug" ]; then
```
bernie-laberge, line 284: How do you detect the current build type (debug/release) ? Is there any _build_debug path detection ?
Or do you always assume release when sourcing rvcmds.sh ?
Can we still have both _build and _build_debug ? 
Because this is what I currently do: I have both _build and _build_debug

pbergeron-adsk: yes of course you can have _build and _build_debug. You just need to switch between them using rvdebug or rvrelease.

if you source twice, it will keep the same build type (Debug or Release)

as for the default, if _build_debug exists but not _build, it will default to debug.

rvcmds.sh
pbergeron-adsk: yes I changed the doc

the rvdebug and rvrelease aliases dont exist but they are bash commands/functions and thus accessible by the prompt

rvcmds.sh
```diff
@@ -105,69 +105,189 @@ else
...
+
+# Update all path variables based on current RV_PATH_SUFFIX
+__rv_update_paths() {
+  RV_BUILD_DIR="${RV_HOME}/_build${RV_PATH_SUFFIX}"
+  RV_INST_DIR="${RV_HOME}/_install${RV_PATH_SUFFIX}"
+  RV_APP_DIR="${RV_BUILD_DIR}/stage/app"
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    RV_APP_DIR="${RV_APP_DIR}/RV.app/Contents/MacOS"
+  fi
+}
```
cedrik-fuoco-adsk, line 146: RV_APP_DIR is not right for Linux and Windows. it has to point to `${RV_APP_DIR}/stage/app/bin`

cedrik-fuoco-adsk:
```suggestion
__rv_update_paths() {
  RV_BUILD_DIR="${RV_HOME}/_build${RV_PATH_SUFFIX}"
  RV_INST_DIR="${RV_HOME}/_install${RV_PATH_SUFFIX}"
  RV_APP_DIR="${RV_BUILD_DIR}/stage/app"
  if [[ "$OSTYPE" == "darwin"* ]]; then
    RV_APP_DIR="${RV_APP_DIR}/RV.app/Contents/MacOS"
  else
    RV_APP_DIR="${RV_APP_DIR}/bin"
  fi
}
```

docs/build_system/config_common_build.md
```diff
@@ -93,8 +98,8 @@ rvmk
 rvmkd
```
bernie-laberge: rvmk since rvmkd no longer exists

docs/build_system/config_common_build.md
```diff
@@ -93,8 +98,8 @@ rvmk
...
 
 # Note: rvmkd is just an alias to these two commands:
```
bernie-laberge: rvmk since rvmkd no longer exists

rvcmds.sh
pbergeron-adsk: fixed

PR #925, opened 2025-09-25
SG-37413: Update BMD DeckLink SDK to 15.0

eloisebrosseau:
### Summarize your change.

Some important changes were made to the SDK, which means some refactoring in RV was necessary. All changes were made following the examples and instructions provided in the [14.3 Migration Guide](https://documents.blackmagicdesign.com/UserManuals/DeckLinkSDKMigrationGuide.pdf?_v=1732780810000):

- [X] Replace bytesPerRow with the new available helper method `IDeckLinkOutput::RowBytesForPixelFormat` to
calculate row bytes for pixel format
- [X] Remove HDRVideoFrame and replace it with the new available interface `IDeckLinkVideoFrameMutableMetadataExtensions` directly in DeckLinkVideoFrame
- [X] Migrate from IDeckLinkVideoFrame::GetBytes to access the underlying video frame buffer with the IDeckLinkVideoBuffer interface
- [X] Refactor StereoVideoFrame to be use through a Provider class with the new method `IDeckLinkMutableVideoFrame::SetInterfaceProvider`. The provider object is able to provide access to the custom interface when its IUnknown::QueryInterface method is called. We now only need to define our StereoVideoFrame as a custom class of IDeckLinkVideoFrame3DExtensions

RV should be able to use the latest version of the BMD SDK.

I tested broadcasting some media files from my macOS with the BMD UltraStudio 4K and BMD SmartScope Duo 4K monitor.

### DIFF
--- a/src/plugins/output/BlackMagicDevices/StereoVideoFrame.cpp
+++ b/src/plugins/output/BlackMagicDevices/StereoVideoFrame.cpp
@@ -44,30 +44,17 @@ HRESULT StereoVideoFrame::Provider::QueryInterface(REFIID iid, LPVOID* ppv)
         ->QueryInterface(iid, ppv);
 }
 
-ULONG StereoVideoFrame::Provider::AddRef()
-{
-#ifdef PLATFORM_WINDOWS
-    return _InterlockedIncrement((volatile long*)&m_refCount);
-#else
-    return ++m_refCount;
-#endif
-}
+ULONG StereoVideoFrame::Provider::AddRef() { return ++m_refCount; }
 
-ULONG StereoVideoFrame::Provider::Release()
+ULONG
+StereoVideoFrame::Provider::Release()
 {
-#ifdef PLATFORM_WINDOWS
-    ULONG newRefValue = _InterlockedDecrement((volatile long*)&m_refCount);
-    if (!newRefValue)
-        delete this;
-    return newRefValue;
-#else
     ULONG refCount = --m_refCount;
     if (refCount == 0)
     {
         delete this;
     }
     return refCount;
-#endif
 }
 
 StereoVideoFrame::StereoVideoFrame(IDeckLinkMutableVideoFrame* owner,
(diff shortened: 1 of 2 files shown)


src/plugins/output/BlackMagicDevices/StereoVideoFrame.cpp
```diff
@@ -1,170 +1,163 @@
...
-ULONG StereoVideoFrame::AddRef(void)
+ULONG StereoVideoFrame::Provider::AddRef()
 {
 #ifdef PLATFORM_WINDOWS
     return _InterlockedIncrement((volatile long*)&m_refCount);
 #else
-    ScopedLock lock(m_refMutex);
-    m_refCount++;
-    return m_refCount;
+    return ++m_refCount;
```
bernie-laberge: Nice cleanup ! 
I would get rid of the _InterlockedIncrement altogether and replace it with the same std::atomic based implementation that you added for all the platforms.

bernie-laberge: I asked AutodeskGPT and it also agrees :-)
`Yes, you can use std::atomic in place of _InterlockedIncrement. The std::atomic library in C++ provides atomic operations that are designed to be safe to use in concurrent environments, making it a suitable replacement for the Interlocked functions provided by the Windows API.`

src/plugins/output/BlackMagicDevices/DeckLinkVideoDevice.cpp
```diff
@@ -919,8 +950,10 @@ namespace BlackMagicDevices
         m_firstThreeCounter = 0;
         m_frameCount = 0;
         m_totalPlayoutFrames = 0;
-        m_lastPboData = NULL;
-        m_secondLastPboData = NULL;
+        m_lastPboData = nullptr;
+        m_secondLastPboData = nullptr;
+        m_lastPboData = nullptr;
```
bernie-laberge: Duplicate ?

src/plugins/output/BlackMagicDevices/DeckLinkVideoDevice.cpp
```diff
@@ -919,8 +950,10 @@ namespace BlackMagicDevices
...
+        m_secondLastPboData = nullptr;
```
bernie-laberge: Duplicate ?

bernie-laberge (approved):
Beautiful Cleanup !
Thank you @eloisebrosseau !

src/plugins/output/BlackMagicDevices/DeckLinkVideoDevice.cpp
eloisebrosseau: Oops! I had an issue when applying a stash I had at some points and though I had removed all the duplicate code blocks that got added because of it. Thank you catching that!

eloisebrosseau: Fixed!

src/plugins/output/BlackMagicDevices/DeckLinkVideoDevice.cpp
eloisebrosseau: Fixed!

src/plugins/output/BlackMagicDevices/StereoVideoFrame.cpp
eloisebrosseau: I agree! I didn't want to touch too much the Windows only code blocks, but there's no reason to keep this old way of handling atomic operations on Windows! I'll change it! Thank you!

PR #926, opened 2025-09-25
SG-40472 - Disable events filter for participants when MultiUser is ON 

cedrik-fuoco-adsk:
SG-40472 - Disable events filter for participants when MultiUser is ON 

### Linked issues
n/a

### Summarize your change.
Commented out the part where OpenRV disable the some toolbars buttons based on the Live Review filtering. It will be rework soon.

Before merging this PR, a PR for the Live Review plugin needs to be merged.

We need the toolbar actions to be enabled on all time for now.

MacOS


eloisebrosseau (approved):

bernie-laberge (approved):
LGTM

Issue #610, opened 2024-10-28
[Bug]: RV Crash w/ Audio Scrubbing On


skyyred:
### What happened?

RV crashes.

repro step:
1) File > Open... load a .mov with embeded audio (pcm_s16le in my case)
2) Set noMovieAudio for the source to 1 (on)
`from rv import commands as rvc`
`rvc.setIntProperty("sourceGroup000000_source.group.noMovieAudio", [1])`
3) File > Clear (Crashes RV) also loading another .rv session file would crash RV

Audio Output Module: ALSA(Safe)
newest version repro'd on RV-2024.1.0 (9f48e90 ([Open RV] 6a328a5))

### List all the operating systems versions where this is happening

cent7

### On what computer hardware is this happening?

Intel Xeon, quadro p5000

### Code of Conduct

- [X] I agree to follow this project's Code of Conduct

tjjackson:
I have a fix for this and plan to contribute it for ASWF Dev Days, if you want to assign it to me

bernie-laberge:
This is fantastic !
It's all yours.

Thank you @tjjackson !

tjjackson:
Per @markreidvfx, his PR #848 may address this issue

pbergeron-adsk:
@tjjackson can you follow-up on this? Has PR #848 fixed the issue? I believe we've never seen a PR or update from you...

tjjackson:
@pbergeron-adsk I will need to rebuild OpenRV with the latest changes to confirm, but I believe PR #848 fixes the crash.  I think there still be a bug with the audio buffering progress glyph that I had addressed in PR #927.  I'll test if it is still an issue with the changes in #848 and if so, I'll either re-open #927 or open a new PR for the audio buffering glyph bug.

pbergeron-adsk:
Thanks. Let us know how it goes, so that we can close or reopen the issue as appropriate.

tjjackson:
I was able to build the latest w/ the changes from #848 and confirmed I am still encountering an issue with the audio progress glyph because of the audio buffering ending prematurely.  Things worked again after I applied my fix for that, so have re-opened #927 and will update it with just the audio progress glyph fix

tjjackson:
#927 has been updated

pbergeron-adsk:
Thanks we'll have a look

pbergeron-adsk:
Closing as fixed by @bernie-laberge -- thank you @tjjackson

PR #927, opened 2025-09-25
Fix audio buffering crash

tjjackson:
### Linked issues

[#610](https://github.com/AcademySoftwareFoundation/OpenRV/issues/610)

### Summarize your change.

This update adds a check to audioCacheProgressGlyph() to confirm that the total frames is not less than or equal to 0.

If audio buffering ends prematurely (ex: the EDL data is changed while a source's audio is being actively cached), this can cause audioCacheProgressGlyph to calculate the total frames as less than or equal to 0, which would cause the code to either divide by zero or draw a negative percentage glyph, which causes OpenRV to freeze.  I added a check that the total frames are greater than 0.

Successfully tested on Rocky Linux 9.5.

### DIFF
--- a/src/lib/app/mu_rvui/rvui.mu
+++ b/src/lib/app/mu_rvui/rvui.mu
@@ -5575,8 +5575,11 @@ global let enterFrame = startTextEntryMode(\: (string;) {"Go To Frame: ";}, goto
 \: audioCacheProgressGlyph (void; bool outline)
 {
     let (_, u, _) = cacheUsage(),
-        total     = (frameEnd() - frameStart()) / fps(),
-        pcent     = u / total;
+        total     = (frameEnd() - frameStart()) / fps();
+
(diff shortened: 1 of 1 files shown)


markreidvfx:
I think my PR addresses this issue too and a few other issues
 https://github.com/AcademySoftwareFoundation/OpenRV/pull/848

tjjackson:
Thanks for the heads up Mark.  I'm going to close this out

markreidvfx:
I didn't do the `audioCacheProgressGlyph` bit though. Maybe you could make a PR for that

tjjackson:
I was able to build the latest w/ the changes from #848 and confirmed I am still encountering an issue with the audio progress glyph. so I will update this PR just to address that

bernie-laberge (approved):
LGTM
Thank you @tjjackson !

PR #929, opened 2025-09-26
added bind to event "clear-annotations-current-frame"

pbergeron-adsk:
### Summarize your change.

Added two events that can be sent internally, to clear the current frame of annotations.  I want to call this internal event from rv_tests.py, when we auto-draw-annotations (circles).

This is necessary for auto-draw-annotations in rv_test.py, where I'd like to clear the frame every 10 annotations (otherwise, in stress tests, annotations keep acumulating, and the render gets slower and slower and slower and slower.

rv_tests.py, which is in Commercial RV. 


bernie-laberge (approved):
LGTM

PR #933, opened 2025-09-30
SG-40541: fix broken qt qmessagebox 6.5+ on macOS Tahoe

pbergeron-adsk:
### Linked issues
Fixes SG-40541

### Summarize your change.
on macOS Tahoe, QMessageBox which is used for the mu alertPanel command produces a correct-looking messagebox, but it is unresponsive to mouse clicks, keyboard TABs, button focus changes, etc.

* We tried to strongFocus the buttons: didn't work
* We tried modifying the plist to disable liquid glass macOS look: didn't work
* We tried a few other things, didn't work.

In the end, we just decided to implement QAlertPanel for the purpose of responding to the mu alertPanel command, respecting the same contract and API of QMessageBox, but without the bugs.

broken Qt QMessageBox which we used to display error/info/warning messages.

macOS, Tahoe. This remains to be tested on Sequoia, Linux and Windows.

image

image


src/lib/app/RvCommon/QAlertPanel.cpp
```diff
@@ -0,0 +1,316 @@
+//******************************************************************************
+// Copyright (c) 2024 Autodesk.
```
cedrik-fuoco-adsk: 2025

src/lib/app/RvCommon/QAlertPanel.cpp
```diff
@@ -0,0 +1,316 @@
...
+    {
+        // Get the text size with word wrapping at a reasonable width
+        QFontMetrics fm(m_textLabel->font());
+
+        // Try different widths to find the best text wrapping
+        float bestFitRatio = 1e100;
+        int bestWidth = 100; // Start with a reasonable width
+        int bestHeight = 50;
+
+        // float targetBestFitRatio = 8.0/2.0;
```
cedrik-fuoco-adsk: Can you remove this if it is not needed

PR #934, opened 2025-09-30
Add Ruff linter and formatter

eloisebrosseau:
### Summarize your change.

Note that multiple files had too many issues that would need to manually be reviewed to follow best practices and pass Ruff linting rules. The idea here was not to fix everything but rather to add tools to make working on this project easier. For that reason, some rules were ignore in specific files in `ruff.toml` that is added in another PR. Since some of these files were added recently, we should to take the time to fix them properly and make sure the code is following Python's conventions.

This PR adds a `Lint and Format` step in our Github actions CI and Ruff as a `pre-commit` hook to lint and format our Python code. I also added some documentation on how to use these tool to contribute to the project when working on Python files.

The files to review are:

- [docs/build_system/config_common_build.md](https://github.com/AcademySoftwareFoundation/OpenRV/pull/934/commits/34b7e9d8444afa85ece243ea0cb1443bef690f18#diff-65b028884e1988b5969e57e20e487c25afe94b933b9df3e69a07f5953b15be23)
- [ruff.toml](https://github.com/AcademySoftwareFoundation/OpenRV/pull/934/commits/e8d09fbf31721e8f4e1c351bc451c2aa6d93cb81#diff-26f436cb29eecc7bb4f9066747b43df8da81ac808955ce95c109f5d7aa778989)
- [.github/workflows/ci.yml](https://github.com/AcademySoftwareFoundation/OpenRV/pull/934/commits/ba5ba30702862e35073775c982ee2203adf1a313#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03f)

The project had a formatter but no linter. Ruff is a tool that can do exactly what Black was doing as a formatter while also being able to lint and fix the code following PEP8 at the same time. We were also not enforcing the use of a linter and a formatter for Python in our CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -49,8 +49,11 @@ jobs:
           python3 -m pip install --upgrade pip
           pip install ruff
       - name: Run Ruff check
+        # Check if all the files are linted correctly
+        # Output errors as inline annotations
         run: ruff check --output-format=github
       - name: Run Ruff format
+        # Check if all the files are formatted correctly
         run: ruff format --check
   rocky-linux:
     if : ${{ github.repository_owner == 'AcademySoftwareFoundation' }}
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,23 @@ env:
   WINDOWS_QT6_ARCHIVES: 'd3dcompiler_47 opengl32sw qtbase qtdeclarative qtsvg qttools qttranslations'
 
 jobs:
+  lint-format:
+    name: "Lint and Format"
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v4
+      - name: Install Python
+        uses: actions/setup-python@v5
+        with:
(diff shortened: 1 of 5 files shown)


.github/workflows/ci.yml
```diff
@@ -35,6 +35,23 @@ env:
...
+        with:
+          python-version: "3.11"
+      - name: Install dependencies
+        run: |
+          python3 -m pip install --upgrade pip
+          pip install ruff
+      - name: Run Ruff check
+        run: ruff check --output-format=github
+      - name: Run Ruff format
+        run: ruff format --check
```
cedrik-fuoco-adsk, line 57: Could you add a comments on what are the differences between these two command? Are they both just doing a check? @eloisebrosseau

eloisebrosseau: They are just doing a check, and failing the step if the check doesn't pass. I added comments, let me know if I should add more!

cedrik-fuoco-adsk (approved):

PR #935, opened 2025-09-30
Lint Python code with Ruff

eloisebrosseau:
### Summarize your change.

Run `ruff check --fix` on all Python files with the default linting rules.

Note that multiple files had too many issues that would need to manually be reviewed to follow best practices and pass Ruff linting rules. The idea here was not to fix everything but rather to add tools to make working on this project easier. For that reason, some rules were ignore in specific files in `ruff.toml` that is added in another PR. Since some of these files were added recently, we should to take the time to fix them properly and make sure the code is following Python's conventions.

The project had a formatter but no linter. Ruff is a tool that can do exactly what Black was doing as a formatter while also being able to lint and fix the code following PEP8 at the same time.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -4,11 +4,6 @@ repos:
     hooks:
       - id: cmake-format
 
-  - repo: https://github.com/psf/black
-    rev: 24.8.0
-    hooks:
-      - id: black
-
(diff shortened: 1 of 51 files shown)


bernie-laberge (approved):
LGTM

PR #936, opened 2025-09-30
Format Python code with Ruff

eloisebrosseau:
### Summarize your change.

Run `ruff format` on all Python files with the default formatting rules.

The formatting was done following the default rules but the line length was increased to 120 in `ruff.toml` that is added in another PR.

Ruff is a tool that can do exactly what Black was doing as a formatter while also being able to lint and fix the code following PEP8 at the same time. Using the same tool for both makes the steps of linting and formatting easier to run locally, with pre-commit and on the CI.

### DIFF
--- a/docs/conf.py
+++ b/docs/conf.py
@@ -3,19 +3,19 @@
 # -- Project information
 import re
 
-project = 'Open RV'
-author = 'Open RV contributors'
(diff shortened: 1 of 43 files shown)


bernie-laberge (approved):
LGTM

PR #937, opened 2025-09-30
SG-n/a: Mu syntax highlighting for vscode or Cursor.

pbergeron-adsk:
Fixes SG-40568

Tonight I had enough, and I fixed a forever-irritant. 

Did you ever get tired of looking at Mu in the editor as if it was plain, white-on-black text?  Are you tired of bleeding eyes, not finding that missing semicolon, or parenthesis, or stray comma, in that not-so-intuitive-sometimes-cryptic scripting langauge?

Well, no longer.  Mu can be recognized as an actual language with proper syntax highlighting. 

Finally your eyes can rest.

### Summarize your change.

Added support for vscode and cursor syntax highlighting of Mu language (it's not perfect, but it's certainly better than nothing!)

To install this IDE extension, I provided two shell scripts: install_vscode.sh, or install_cursor.sh.  Install the one that applies to you. 

Added a folder called mu_language_extension in OpenRV. 
Because this extension is very niche, it should not be published to the microsoft or cursor marketplaces. 

Does not affect code at all. It' possible some syntax keywords, tokens, or constructs are incorrect, but the result is quite pleasing to the eye.

### If possible, provide screenshots (before, and after) 

image

image

image


markreidvfx:
Amazing looking forward to trying this! I've been using obj-c syntax highlighting in vscode for mu files and its been doing the job.

pbergeron-adsk:
won't merge since we can just use "CPP" as the code style.

PR #939, opened 2025-10-02
Mu categories

pbergeron-adsk:
### Summarize your change.

This is a PR that adds support for adding the code infrastrure needed to specify categories of actions/events. Categories aren't actually assigned to menu actions or binds yet, this will come in a separate PR where the product management team will be able to easily pick-and-choose what functions go with what categories.

You'll notice the following:

**Better way to build menus:**  The way we build the RV main menu is completely redone. Instead of calling Mu-to-CPP contructors directly to create the nested hierarchy of MenuItem objects, we now have helper functions like "subMenu(), menuItem(), menuSeparator()" etc.

**Better way to bind a key shortcut to menu item:** It used to be that in the menu item, we'd specify the key combination, but the actual binding of the key to the function needed to be done with a separate bind() call (there were ~100 of them).  Not only was this quite redundant, but they required similar function pointers to the execution function, and had no validators like menu items do (therefore, we could call a bound function even if it was disabled in the menu item).

Those problems are gone: If in the menuItem() we specify an event name that begins with "key-down--", we now convert the eventPattern to the correct menu label (so that it appears correctly in the menu item entry) and then we issue the bind() call directly within menuItem() to associate the key down event to the same handler. We have thus commented-out the ~100 now-unnecessary bind() calls.  

**Validation before calling key-bound event funcs:** As alluded to just above, we could not previously specify validators on bound keys before. Because in menuItem() we already had a validator state function to enable/disable the menu item, we can create a composite mu function that first checks for the enabled/disabled state before calling the actual bound function. So, if it's disabled by the menu, it's also now disabled by the key shorcut.

**Category-specific enabled/disabled features**: The entire goal of the work above was to be able to logically categorize actions and events, and to be able to disable entire categories of actions because they would interfere with a plugin. We have this issue with LiveReview, where participant were completely prohibited from doing almost anything because they were passive observers (so a global flag was used for this purpose) but with upcoming multi-user controls for live review sessions, only certain types of events should be disabled for participants and reserved to the host/presenter (notably, those that affect the source media, but I digress). In any case, we needed a way to classify/disable such functionality groups, so we made categories.  In this PR, you'll notice that we don't actually specify which categories are assigned to which menu items/binds, but the code to do so is there.

To handle categories, the new menuItem() function creates a composite mu function that first checks to see if the category is disabled before calling the actual handler. If it's disabled, the composite mu function broadcasts an "event blocked" internal event and doesn't call the handler. 

**Try-catch statements when building the menus**: Mu syntax can sometimes be quite obscure, and building a meta-recursive function like the menu-building function can become extremely hard to debug if there's a problem with one of the parameters of if there's a problem with the underlying code.  I've lost an insane amount of time with all sorts of obscure errors that could have been caught if exceptions had been present, so for the benefit of all in the future, I've added exception handling.

**A few (missing?!?) menu items for pixelRelativeScale were added:** During my code conversion activities, I noticed some menu items appered to be missing because they still had keyboard binds for them, so.... I added the menu items (back?).

**New string functions!** Well, this one's a bit of a surprise, isn't it? When needing to convert the eventPattern to a key shortcut string, I needed to do some string manipulations but neither stirng.find(), or string.contains() or string.replace() existed.  I considered implementing them in mu itself with the functions that actually did exist, but the code was becoming quite convoluted, so for the sake and sanity of all, I added these two additional string functions for everyone to use.

**Whats next?** In a soon-to-be-PR, we will categorize events, fix the LiveReview to specify disabled categories instead of using a global flag, and remove the global flag check that was peppered throughout mu code to block certain types of events. Again, this will need to be a separate (upcoming) PR.

It was getting ludicrously hard and ugly to implement category-handling with the current code organisation, so I before I did any of that, I reworked the way we do binds and create menus so that they are cohesive and coherent, so that I had a proper place to do categories handling in a way that is centralized.

macOS.

With the current PR, complete compatibility should exist with the previous implementation. Things should work exactly as before.


src/lib/app/mu_rvui/app_utils.mu
```diff
@@ -24,6 +25,15 @@ MenuStateFunc   := (int;);
     \: (void; Event ignored) { f(); };
 }
 
+\: makeCategoryEventFunc (EventFunc; string category, EventFunc f) {
+    \: (void; Event ev) {
+        if (!commands.isEventCategoryEnabled(category))
+            sendInternalEvent("action-blocked");
```
bernie-laberge: Nice !
I think it would be informative to add the category if sone code wants to know.
sendInternalEvent("action-blocked", category);

src/lib/app/py_rvui/rv_commands_setup.py
```diff
@@ -301,6 +301,9 @@
     "isDebug",
     "setFilterLiveReviewEvents",
     "filterLiveReviewEvents",
+    "enableEventCategory",
```
bernie-laberge, line 304: Beautiful!

src/lib/app/mu_rvui/app_utils.mu
pbergeron-adsk: Fixed. Also, the event name was wrong, it was supposed to be "category-event-blocked", also fixed.

src/lib/app/py_rvui/rv_commands_setup.py
pbergeron-adsk: yay

cedrik-fuoco-adsk (approved):

pbergeron-adsk:
PR closed because bigger enclosing PR was merged instead

pbergeron-adsk:
PR closed because bigger enclosing PR was merged instead

PR #941, opened 2025-10-06
SG-40626: Garbage returned by Mu's exception handling if returning value inside catch block

pbergeron-adsk:
### Linked issues

SG-40626

Under certain circumstnaces, returning a value from within a mu catch block returned garbage data.

### Summarize your change

This is one of those changes that requires some background and a thorough explanation before describing the fix.... so here goes!

While testing a PR for menu refactoring and event categorization, I noticed that when RV started (without loaded media),  that certain menu items are enabled when they shouldn't be. For example, Flip/Flop should have been disabled, but they were not! But, my code seemed to be correct. 

image

For these items, the state callback passed as a validator to the menu item is declared as an alias to a function that returns a lambda, in rvui.mu:

_toggleFlipState = toggleIntPropState("#RVTransform2D.transform.flip")_

image

In our PR that exhibits the bug, we take the state func (toggleFlipState) and combine it with another category-validating function to create a composite function that tests both the category and the state. This composite function is called by the menu item handler to determine if the item should be enabled/disabled. It's when using this composite function that the bug manifested itself.

The bug showed up when _toggleIntPropState()_ calls getIntProperty(...) inside a Mu try block. If the property doesn't exist (which is the case when no media is loaded at RV startup), the C++ code for getIntProperty in CommandsModule.cpp throws an exception, which is caught by the Mu catch block, and should return DisabledMenuState (aka: -1) from within the Mu catch block. 

That should have worked just fine (and it did work just before my refactoring), but it didn't anymore after I refactored: our menu item stayed enabled. My composite function looked fine, too, so I had to dig deeper.

To debug this, I added traces in toggleIntPropState() to see when if enter the function, if we enter the catch, if we return the correct value, and if anything executes after the return in the Mu catch block. I also printed the returned value in the composite function that used the propStateFunc.

image

and

image

And we got interesting results: 

First, we saw the traces appear when we entered the propStateFunc, and when we entered the catch block, as expected. Thus, the code did indeed "return DisabledMenuState" (-1) from within the catch block.   BUT the composite function that consumed the return value from said propStateFunc showed a garbage value (e.g., 187387624), which is definitely not the expected -1.

image

To confirm that the problem only existed when returning from within a catch block, I just moved the return statement to just outside the catch block, and lo and behold, the problem was fixed! (Flip/Flop menu items were now disabled, as expected)

Conclusion: There's a problem in Mu's exception handling when returning from inside a catch block.

=================

Further problem analysis/debugging/solving.

When Mu calls getIntProperty() (in CommandsModule.cpp) and the property doesn't exist, the C++ code throws an exception via throwBadPropertyType(), which is supposed to be caught by the catch block in Mu. The C++ catch code for mu is implemented in ExceptionType::mu__catch(), which contains some code that looks like black magic invocations, but I digress.

Anyway, after a lot of digging, the problem is that in Mu, return statements are implemented via longjmp, while try/catch is implemented via C++ exceptions. When a return occurs inside a catch block, the longjmp must traverse C++ exception handling code, which is not allowed (in fact it's undefined behaviour as per the C++ standard).

Indeed, in ExceptionType::mu__catch, there was simply no mechanism whatsoever to properly preserve the return value when doing return from the catch block; and the longjmp call does not appear to be set up in a way that it traverses the C++ exception handling code safely. Its actually unclear to me why this code used to work at all, I think it might have been because the stack was left in an "undisturbed" state, or at least, still in a state where the return value happened to remain at the right place in memory, but I don't know for sure.

Anyway, to fix this, we need to modify mu__catch() and mu__catch_all() to Install a setjmp that captures return statements from the catch body, and if a return is detected, throws an internal C++ exception that carries the returned value with it. Then, mu__try() catches this exception, and converts it to a proper longjmp _after_ the exception was handled, which allows the function to exit properly with the returned value (whew, holy code path, batman!).

macOS

Due to the very unusual type of fix this is, I don't expect anyone to be able to look at this code change and go "LGTM!" right away.  In fact, I expect people to be more like : "what the hell am I even looking at?".

For this reason, it's probably the only time I will suggest that code reviewers who look at this actually pull the branch locally, and ask AI the following: 

_"What are these changes about, why are they necessary, what do the changes do, how does it work, is the fix correct and if so, why is it correct, does it work with all return types, even complex types?"_

### DIFF
--- a/src/lib/mu/Mu/Mu/Exception.h
+++ b/src/lib/mu/Mu/Mu/Exception.h
@@ -138,7 +138,7 @@ namespace Mu
 
         const Value& returnValue() const noexcept { return _returnValue; }
 
-        const char* what() override const noexcept
+        const char* what() const noexcept override
         {
             return "return from Mu catch block";
         }


src/lib/mu/Mu/Mu/Exception.h
```diff
@@ -117,6 +117,36 @@ namespace Mu
...
+    //  This exception is thrown when a return statement occurs inside a Mu
+    //  catch block. Using longjmp in this case causes undefined behavior
+    //  because it jumps through C++ exception handling code. Instead, we
+    //  throw a C++ exception that carries the return value, which mu__try
+    //  can catch and handle properly.
+    //
+    class MuReturnFromCatchException : public std::exception
+    {
+    public:
+        MuReturnFromCatchException(const Value& returnValue) throw()
```
rogernelson: It's probably worth adding an explicit here.  I think doing something like `throw value` could lead to some pretty wonky behaviour if there are other exception classes that allow implicit conversion.

src/lib/mu/Mu/Mu/Exception.h
```diff
@@ -117,6 +117,36 @@ namespace Mu
...
+    //
+    //  MuReturnFromCatchException
+    //
+    //  This exception is thrown when a return statement occurs inside a Mu
+    //  catch block. Using longjmp in this case causes undefined behavior
+    //  because it jumps through C++ exception handling code. Instead, we
+    //  throw a C++ exception that carries the return value, which mu__try
+    //  can catch and handle properly.
+    //
+    class MuReturnFromCatchException : public std::exception
```
rogernelson: Maybe its a bit pedantic here, but you probably want final ineritance (in conjuction with public). I can't see a use case where you'd want to allow someone deriving from this.

src/lib/mu/Mu/Mu/Exception.h
rogernelson: minor, but you can use move semantics here to save yourself a copy of returnValue.

rogernelson: Also `throw()` has long been deprecated, you should use `noexcept` instead (and same applies for all these class methods).

src/lib/mu/MuLang/ExceptionType.cpp
```diff
@@ -243,8 +275,26 @@ namespace Mu
 
     NODE_IMPLEMENTATION(ExceptionType::mu__catch_all, bool)
     {
-        NODE_ARG(0, void);
-        NODE_THREAD.setException(0);
+        // Install a setjmp to catch return statements from the catch block
+        // body.
+        NODE_THREAD.jumpPointBegin(JumpReturnCode::Return);
+        int returnCode = SETJMP(NODE_THREAD.jumpPoint());
```
rogernelson: Might as well make that const

src/lib/mu/Mu/Thread.cpp
```diff
@@ -176,6 +176,13 @@ namespace Mu
         {
             _returnValue = _rootNode->eval(*this);
         }
+        catch (Mu::MuReturnFromCatchException& retExc)
```
rogernelson: Should these be const & ?

src/lib/mu/MuLang/ExceptionType.cpp
```diff
@@ -151,6 +151,15 @@ namespace Mu
             NODE_ARG(0, int);
 #endif
         }
+        catch (Mu::MuReturnFromCatchException& retExc)
+        {
+            // BUG FIX: A return statement occurred in a catch block.
+            // We caught it as a C++ exception (to avoid undefined behavior from
+            // longjmp through C++ exception handling). Now convert it back to
+            // a longjmp that will be caught by the enclosing function's setjmp.
+            NODE_THREAD.jump(JumpReturnCode::Return, 1, retExc.returnValue());
```
rogernelson: Aren't you still doing a longjmp from an exception here? Maybe instead you to assign returnValue to an optional<Value> (or something like that) and do the longjmp at the end of the function if its set?

src/lib/mu/MuLang/ExceptionType.cpp
```diff
@@ -231,8 +240,31 @@ namespace Mu
                     if (rval = t->match(e->type()))
                     {
                         NODE_ARG(0, Pointer);
-                        NODE_ARG(1, void);
-                        NODE_THREAD.setException(0);
+
```
rogernelson, line 257: It is a shame we need to do all this just to be able to return from a catch inside mu. I can't necessarily think of a better way to do it though,.

src/lib/mu/Mu/Thread.cpp
pbergeron-adsk: fixed

src/lib/mu/MuLang/ExceptionType.cpp
pbergeron-adsk: fixed

src/lib/mu/Mu/Mu/Exception.h
pbergeron-adsk: fixed

src/lib/mu/MuLang/ExceptionType.cpp
pbergeron-adsk: it is what it is

src/lib/mu/Mu/Mu/Exception.h
pbergeron-adsk: fixed

src/lib/mu/Mu/Mu/Exception.h
pbergeron-adsk: both fixed

src/lib/mu/MuLang/ExceptionType.cpp
pbergeron-adsk: WOOPS! You're correct. Good catch, I fixed that too.

src/lib/mu/Mu/Mu/Exception.h
```diff
@@ -117,6 +117,36 @@ namespace Mu
...
+        explicit MuReturnFromCatchException(Value&& returnValue) noexcept
+            : _returnValue(std::move(returnValue))
+        {
+        }
+
+        virtual ~MuReturnFromCatchException() noexcept {}
+
+        const Value& returnValue() const noexcept { return _returnValue; }
+
+        virtual const char* what() const noexcept override
```
rogernelson: minor: you don't need override and virtual. Might be worth remove the virtual though as some compilers might warn about it.

rogernelson (approved):
Good job, not a simple issue to pin down.

src/lib/mu/Mu/Mu/Exception.h
```diff
@@ -117,6 +117,36 @@ namespace Mu
...
+        explicit MuReturnFromCatchException(Value&& returnValue) noexcept
+            : _returnValue(std::move(returnValue))
+        {
+        }
+
+        virtual ~MuReturnFromCatchException() noexcept {}
+
+        const Value& returnValue() const noexcept { return _returnValue; }
+
+        const char* what() const noexcept
```
rogernelson: You will want to leave the override if you're intending to reimplement the what() from std::exception.

pbergeron-adsk:
> Good job, not a simple issue to pin down.

thanks

bernie-laberge (approved):
LGTM

PR #946, opened 2025-10-14
Optimize media loading when loading an .rv session file

bernie-laberge:
### Linked issues
NA

Slow .rv session file loading when multiple sources were present. The new media loading optimization (startPreloadingMedia) was not leveraged when loading a .rv session file. 

### Summarize your change.

This commit now levereges the new startPreloadingMedia when loading a .rv session file. 

Sucessfully tested on macOS

### DIFF
--- a/src/lib/app/RvApp/RvSession.cpp
+++ b/src/lib/app/RvApp/RvSession.cpp
@@ -3123,6 +3123,30 @@ namespace Rv
                 string p = protocol(pc);
                 if (isProgressionProtocol(p))
                     ++m_gtoSourceTotal;
+
+                // Optimization: Start preloading media if this is an
+                // RVFileSource with active media
+                if (p == "RVFileSource"
+                    && !Options::sharedOptions().progressiveSourceLoading)
+                {
+                    IntProperty* mediaActive =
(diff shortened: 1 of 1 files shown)


src/lib/app/RvApp/RvSession.cpp
```diff
@@ -3123,6 +3123,25 @@ namespace Rv
...
+                    IntProperty* mediaActive =
+                        pc->property<IntProperty>("media.active");
+                    if (mediaActive && !mediaActive->empty()
+                        && mediaActive->front() == 1)
+                    {
+                        StringProperty* movie =
+                            pc->property<StringProperty>("media.movie");
+                        if (movie && !movie->empty())
+                        {
+                            startPreloadingMedia(movie->front());
```
markreidvfx: `media.movie` can be more then one item. Can we preload all of them. You can have for example a images sequence in the first index and audio in second. I also think this is used for stereo if I'm not mistaken.

bernie-laberge: Good point ! I will make the change
Thank you @markreidvfx !

PR #957, opened 2025-10-23
SG-40180: Fix rendering anomalies of concurrent paint strokes

bernie-laberge:
### Linked issues
NA

When multiple parciticpants were annotating at the same time in a Live Review session, sometimes some rendering anomalies could be observed where some paint strokes' points were missing until a new paint stroke was introduced.

### Summarize your change.

**Problem:** 
In the paint commands rendering method (ImageRenderer::renderPaint()), an FBO Paint cache is used, excluding the very last paint command because the original code assumed that this paint stroke for example (a PolyLine)  could still be being drawn. However, this assumption was not taking into account that multiple participants could be drawing strokes at the same time, thus leading to occasional paint commands rendering anomalies where a stale cached FBO would be used where it shouldn't. The source of the problem is 

**Solution:**
Now forcing to recompute the renderIDWithPartialPaint id when rendering paint command to make sure that the correct render id (a unique identifier associated with the node to be rendered) is used to fetch the correct FBO cache.

**Also in this commit:**

- When deleting a paint stroke, it still remained in the RV Session because the .splat property had not been deleted (but all the other were:
Example: 
```
"pen:2:1:labergb_23473"
    {
        int splat = 0
    }

```
This was fixed in this commit (annotate_mode.mu).

- Each paint command was unnecessaily duplicated. This must be the result of a bad merge in the oirginal Create branch 
This was fixed in this commit as well (PaintIPNode.cpp)

Successfully tested on macOS.

### DIFF
--- a/cmake/dependencies/glew.cmake
+++ b/cmake/dependencies/glew.cmake
@@ -52,6 +52,19 @@ EXTERNALPROJECT_ADD(
   URL_MD5 ${_download_hash}
   DOWNLOAD_NAME ${_target}_${_version}.zip
   DOWNLOAD_DIR ${RV_DEPS_DOWNLOAD_DIR}
+  # Patch to fix the build issue with OpenGL-Registry
+  # Pinning the OpenGL-Registry version to a specific commit
+  # https://github.com/nigels-com/glew/issues/449
+  # Also clone the required glfixes repository
+  PATCH_COMMAND 
+    cd auto && 
+    git clone https://github.com/KhronosGroup/OpenGL-Registry.git || true &&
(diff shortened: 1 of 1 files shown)


pbergeron-adsk (approved):
In the end, it turned out to be such a small thing, yet, so mysterious to find. Bravo Bernard LGTM

bernie-laberge:
This PR has been sent for Testing
Testing successfully completed.

markreidvfx:
Just a note, This PR message contains a link to non-public git repo

bernie-laberge:
Good catch. I removed it. Thank you @markreidvfx !

PR #960, opened 2025-10-24
SG-39445: Enable participants to broadcast annotations in Collaborative mode for OTIO-based Live Review sessions

eloisebrosseau:
### Summarize your change.

- Add pydantic to validate payload formats in the Live Review plugin
- Add `setCurrentAnnotateModeNode` in rv.commands to manually set the Paint node to use when manipulating stokes in annotate_mode
- Set the mode property on all strokes even if the mode is `RenderOverMode` to prevent the code from throwing an exception when trying to get the value. This is necessary to know if the user is erasing or painting before sending an annotations payload.
- Replace the hardcoded values used to set the width of the stroke with a conversion method from OTIO coordinate to the World Coordinate System in RV.

All the changes mentioned above are needed to properly enable participants to broadcast annotations in Collaborative mode for OTIO-based Live Review sessions.

Drawing strokes with the pen and the eraser tools while being a participant of an OTIO-based Live Review session with Collaborative mode enable. The tests were done on macOS 15.6.1

### DIFF
--- a/src/plugins/rv-packages/otio_reader/annotation_hook.py
+++ b/src/plugins/rv-packages/otio_reader/annotation_hook.py
@@ -29,10 +29,27 @@ def _get_transform_properties():
             global_translate_vec = otio.schema.V2d(0.0, 0.0)
             global_scale_vec = otio.schema.V2d(1.0, 1.0)
 
-            global_translate = commands.getFloatProperty(f"{transform_node}.otio.global_translate")
-            global_scale = commands.getFloatProperty(f"{transform_node}.otio.global_scale")
-            global_translate_vec = otio.schema.V2d(global_translate[0], global_translate[1])
-            global_scale_vec = otio.schema.V2d(global_scale[0], global_scale[1])
+            if commands.propertyExists(f"{transform_node}.otio.global_translate") and commands.propertyExists(
+                f"{transform_node}.otio.global_scale"
+            ):
+                global_translate = commands.getFloatProperty(f"{transform_node}.otio.global_translate")
+                global_scale = commands.getFloatProperty(f"{transform_node}.otio.global_scale")
+                global_translate_vec = otio.schema.V2d(global_translate[0], global_translate[1])
+                global_scale_vec = otio.schema.V2d(global_scale[0], global_scale[1])
+            else:
+                logging.warning(
+                    "OTIO global scale and translate properties not found, using aspect ratio from the source"
+                )
+                try:
+                    media_info = commands.sourceMediaInfo(first_source_node)
+                    height = media_info["height"]
+                    aspect_ratio = media_info["width"] / height if height != 0 else 1920 / 1080
+                except Exception:
+                    logging.warning("Unable to determine aspect ratio, using default value of 16:9")
+                    aspect_ratio = 1920 / 1080
+
+                scale = aspect_ratio / 16
+                global_scale_vec = otio.schema.V2d(scale, scale)
 
             translate = commands.getFloatProperty(f"{transform_node}.transform.translate")
             scale = commands.getFloatProperty(f"{transform_node}.transform.scale")
@@ -67,7 +84,6 @@ def _transform_otio_to_world_coordinate(point):
 
 def hook_function(in_timeline: otio.schemadef.Annotation.Annotation, argument_map: dict | None = None) -> None:
     """A hook for the annotation schema"""
-
     try:
         if argument_map["effect_metadata"]:
             effect_metadata = argument_map["effect_metadata"]


src/build/make_python.py
```diff
@@ -223,7 +223,7 @@ def patch_python_distribution(python_home: str) -> None:
 
     pip_args = python_interpreter_args + ["-m", "pip"]
 
-    for package in ["pip", "certifi", "six", "wheel", "packaging", "requests"]:
+    for package in ["pip", "certifi", "six", "wheel", "packaging", "requests", "pydantic"]:
```
bernie-laberge, line 226: nice !

src/lib/app/mu_rvui/commands.mud
```diff
@@ -691,6 +691,12 @@ nodesOfType """
 Returns all nodes of the given type. In addition to exact node type
 names, you can also use #RVSource which will match RVFileSource or
 RVImageSource.
+
+"""
+setCurrentAnnotateModeNode """
+Set the Paint node to use for the _currentNode in the annotate mode.
+If the node is not a valid Paint node or if it is set to an empty string
+(i.e. ""), the _currentNode value will not be overridden.
```
bernie-laberge: To make it more explicit how to undo this setCurrentAnnotateModeNode(), should we add ?
Therefore, you can reset its action by specifying an empty string to setCurrentAnnotateModeNode().
(I know that the text is legally correct. But when I first read it I was wondering how to reset it and double checked in the code).
Totally optional comment. Feel free to ignore.

src/lib/ip/IPMu/CommandsModule.cpp
```diff
@@ -3554,6 +3554,22 @@ namespace IPMu
         NODE_RETURN(int(s->getSessionType()));
     }
 
+    NODE_IMPLEMENTATION(setCurrentAnnotateModeNode, void)
```
bernie-laberge: I am wondering, should we really add an RV command for this ?
Why couldn't Live Review simply issue the following Annotation package specific event ?
`commands.sendInternalEvent("set-current-annotate-mode-node", nodeName)`

src/build/make_python.py
cedrik-fuoco-adsk, line 226: Is it used or is it for the future? I don't see it used in the code change.

src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -2542,6 +2578,7 @@ class: AnnotateMinorMode : MinorMode
 
     method: activate (void;)
     {
+        sendInternalEvent("annotate-mode-activated");
```
bernie-laberge: Shouldn't we set the event after it has actually been activated ? (After setTags()) ?

src/lib/app/mu_rvui/commands.mud
eloisebrosseau: I think your first sentence might be missing some words. What are you suggesting to add? I am using the empty string to make sure the current node is reset at the end of a Live Review session in my Commercial RV. I could add a sentence explicitly stating that calling `setCurrentAnnotateModeNode()` with an empty string will reset its behaviour to make it clearer

src/build/make_python.py
eloisebrosseau: Like I say in my PR description, it is only needed for the Live Review plugin not for Open RV for now: `Add pydantic to validate payload formats in the Live Review plugin`

src/lib/app/mu_rvui/commands.mud
bernie-laberge: Yes that's what I meant: 
`calling setCurrentAnnotateModeNode() with an empty string will reset its behaviour.`

src/lib/ip/IPMu/CommandsModule.cpp
bernie-laberge: That's a philosophical question: Since that annotate_mode.mu is a package that could potentially be replaced by some other package in the future, does it make sense to add an rv command that communicates with the annotation_mode.mu package ? while Live Review could interact directly with the annotation_mode.mu package via commands.sendInternalEvent() ?

eloisebrosseau: I thought it would be documented more clearly this way, but since I'm already validating the input inside the annotate-mode package, I agree that I should have simply sent the event with the nodeName. I'll make the change and add the documentation in the internal events page instead of commands.mud

src/plugins/rv-packages/annotate/annotate_mode.mu
eloisebrosseau: Sure, I'll change it!

PR #963, opened 2025-10-29
SG-40929: Continued categorization fixes based on QA.

pbergeron-adsk:
### Summarize your change.

It turns out there were a few big holes in my previous PR for SG-40182.

* in rvui.mu and elsewhere, many bind() calls for keyboard events _(or mouse inputs)_ that dont have equivalent menu item also needed categorizations.  There was a missing bind() overload in app_utils.mu. to do this so it was added.  This is by far the largest change.

* in app_utils.mu, calling the "compositeStateFunc()" could return garbage (yet another mu bug, which I'm not fixing here) causing the event func to be called even if the category was disabled. It was replaced with a more straightforward check. (the bug seemed to happen mostly if the handler func defined in menuItem() started with a leading "~" before its name, eg, "~togglePlay")

* in mode_manager.mu, needed to add equivalent protections to filterLiveReviewEvents() which got removed without an equivalent in the shuffle.

* in annotate_mode.mu, needed to make disappear the paint tools if we are in LR presenter mode (eg: annotate_categories is disabled)

* in scrub_offset.mu, needed to guard on scrubbing if playcontrol category is disabled.

* more filterLiveReviewEvents were removed, and replaced with the appropriate binding (or alternative to bind funcs)

---

Things to test, 

*_when in viewer mode:...._*

- fix: stepForward 1 or 100 frames is now blocked
- fix: stepBackward 1 or 100 frames is now blocked
- fix: motion scope popup menu is blocked
- fix: clicking and dragging in timeline/ motion scope is blocked
- fix: marking frames in timeline is blocked
- fix: setting point in timeline is blocked
- fix: annotate panel is closed 
- fix: cannot scrub in scrub_offset mode
- fix: rotate ibage is blocked
- fix: mousewheel to scrub is disabled
- fix: dragging in main view to scrub is blocked
- fix: doing space to play is blocked.

viewmode categories: 
ViewMode Categories were off in viewer mode before, but enabled now. (toggle this category in live_review.py)

 viewmode key shortcuts:
- fix: kbd keys are blocked: M, alt-r, alt-l, shift-l, shift-r, ctrl-m, down, p, pgup, pgdn.
- fix: kbd keys are blocked : aldkspmzvh,.<>OSorcR/

when in nudge mode...
- fix: kbd keys are blocked : left,right,up,down, control and shift variations.
- fix: kbd keys are blocked : z, ctrl-z, alt-n, 

Stephen reported a few problems reported with being able to do some things in LR that he shouldnt be able to do when in muc presenter mode. (participant = viewer)

macOS

### DIFF
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -403,6 +403,7 @@ class: AnnotateMinorMode : MinorMode
                      "%s.join" % n,
                      "%s.cap" % n,
                      "%s.points" % n,
+                     "%s.splat" % n,
                      "%s.startFrame" % n,
                      "%s.duration" % n
                     ];
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -6542,6 +6542,7 @@ namespace TwkMovie
         // Video
         formats["avi"] = make_pair("Audio Video Interleave", vidcap);
         formats["flv"] = make_pair("Flash Video", vidcap);
+        formats["gif"] = make_pair("Graphics Interchange Format", vidcap);
         formats["m3u8"] = make_pair("M3U8 Stream Metadata", vidcap);
         formats["m4v"] = make_pair("iTunes Video Format (from MPEG-4)", vidcap);
         formats["mkv"] = make_pair("Matroska Video", vidcap);
--- a/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
+++ b/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
@@ -705,50 +705,6 @@ namespace IPCore
         if (showPaint)
         {
             std::lock_guard<std::mutex> commandsGuard{m_commandsMutex};
-            if (m_frameMap.count(frame) != 0)
-            {
-                Components& comps = m_frameMap[frame];
-                size_t s = comps.size();
-
-                for (size_t q = 0; q < s; q++)
-                {
-                    Component* c = comps[q];
-                    int numProps = c->properties().size();
-
-                    if (m_penStrokes.count(c) >= 1)
-                    {
-                        if (numProps == 0)
-                        {
-                            m_penStrokes.erase(c);
-                        }
-                        else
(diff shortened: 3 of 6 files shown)


src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -1951,7 +1951,7 @@ class: AnnotateMinorMode : MinorMode
 
     method: onPresenterChanged(void; Event event)
     {
-        if (_active)
+        if (_active && !commands.isEventCategoryEnabled("annotate_category"))
```
eloisebrosseau, line 1955: Could you do the same in https://github.com/AcademySoftwareFoundation/OpenRV/blob/48b42f014fbe01051696213d09d93af7a249c456/src/plugins/rv-packages/session_manager/session_manager.mu.in#L1132 so the session manager panel is closed the same way the annotation panel is based on the categories?

pbergeron-adsk: When annotations are not permitted, the annotation panel is closed.  This line used to be:

if (_active && commands.filterLiveReviewEvents())
       toggle() (aka: turn off the panel)

but it got lost in the shuffle

cedrik-fuoco-adsk: We should do the same in `session_manager.mu.in` since we have a `onPresenterChanged` there also. (or I can do it in my story about auto-closing panels)

src/lib/app/mu_rvui/mode_manager.mu
```diff
@@ -213,13 +213,12 @@ class: ModeManagerMode : MinorMode
...
-                     then UncheckedMenuState
-                     else if entry.mode._active
-                            then CheckedMenuState
-                            else UncheckedMenuState;
+            // TODO: NOT SURE HOW TO REPLACE FILTERLIVEREVIEWEVENTS HERE
+            if entry.mode eq nil
+                 then UncheckedMenuState
+                 else if entry.mode._active
+                        then CheckedMenuState
+                        else UncheckedMenuState;
```
cedrik-fuoco-adsk: Is this still a TODO?

pbergeron-adsk: Fixed with team

PR #970, opened 2025-10-31
SG-40906 - Re-enable the Toolbar filtering based on the new event filtering

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.

Replaced the deprecated `setFilterLiveReviewEvents()` calls with the existing category-based filtering system. Updated toolbars (RvBottomViewToolBar, RvTopViewToolBar) to respond to category state changes and centralized event category name constants to avoid string duplication across files. 

The toolbar button now follows the category-based filtering.

We want to remove the `setFilterLiveReviewEvents()` approach baked into OpenRV code.

By switching to the category filtering system that already exists, the C++ code stays generic and doesn't need to know about plugin-specific concepts. It just enables/disables categories. 

MacOS

Example of disabled button:
image


src/lib/app/RvCommon/RvBottomViewToolBar.cpp
```diff
@@ -489,13 +490,13 @@ namespace Rv
...
-            // Commented out temporarily and will be reworked soon.
-            //else if (name == "internal-sync-presenter-changed"
-            //         || name == "sync-session-ended")
-            //{
-            //    bool isDisabled = m_session->filterLiveReviewEvents();
-            //    setLiveReviewFilteredActions(isDisabled);
-            //}
+            else if (name == "internal-sync-presenter-changed" 
+                     || name == "sync-session-mode-changed"
+                     || name == "sync-session-ended")
```
bernie-laberge: With the aim of making Open RV completely Live Review plugin agnostic, I think we should create a generic event that would trigger this updateActionAvailability().
Like `update-toolbar-action-availability`
This way, the Live Review plugin could set it in the Live Review plugin if needed (when there is a presenter change, a session mode change, or when the session has ended) by calling `commands.sendInternalEvent("update-toolbar-action-availability")`
This way, if we ever change the business logic of the Live Review plugin in the future with respect to that, we won't have to modifiy Open RV.

src/lib/app/RvCommon/RvTopViewToolBar.cpp
```diff
@@ -563,13 +564,13 @@ namespace Rv
...
-            //else if (name == "internal-sync-presenter-changed"
-            //         || name == "sync-session-ended")
-            //{
-            //    bool isDisabled = m_session->filterLiveReviewEvents();
-            //    setLiveReviewFilteredActions(isDisabled);
-            //}
+            else if (name == "internal-sync-presenter-changed" 
+                     || name == "sync-session-mode-changed"
+                     || name == "sync-session-ended")
+            {
```
bernie-laberge, line 568: Same comment as above:
https://github.com/AcademySoftwareFoundation/OpenRV/pull/970/files#r2482344966

src/lib/ip/IPMu/CommandsModule.cpp
```diff
@@ -4143,12 +4143,34 @@ namespace IPMu
...
     NODE_IMPLEMENTATION(setFilterLiveReviewEvents, void)
     {
         Session* s = Session::currentSession();
         bool shouldFilterEvents = NODE_ARG(0, bool);
 
-        s->setFilterLiveReviewEvents(shouldFilterEvents);
+        if (s && shouldFilterEvents)
+        {
+            // Disable all event categories
+            s->disableEventCategory(EventCategories::CATEGORY_ANNOTATE);
```
bernie-laberge: You need to reenable them all if shouldFilterEvents==false

cedrik-fuoco-adsk: My bad, I forgot to add it. I'll fix it

src/lib/ip/IPMu/CommandsModule.cpp
```diff
@@ -4143,12 +4143,34 @@ namespace IPMu
...
+            s->disableEventCategory(EventCategories::CATEGORY_SOURCE);
+            s->disableEventCategory(EventCategories::CATEGORY_SYSTEM);
+            s->disableEventCategory(EventCategories::CATEGORY_SCREENINGROOM);
+            s->disableEventCategory(EventCategories::CATEGORY_UNCLASSIFIED);
+            s->disableEventCategory(EventCategories::CATEGORY_VIEWMODE);
+            s->disableEventCategory(EventCategories::CATEGORY_WIPES);
+        }
     }
 
     NODE_IMPLEMENTATION(filterLiveReviewEvents, bool)
```
bernie-laberge, line 4171: You can delete the filterLiveReviewEvents() command: since it is NOT used in the RV Live Review plugin (only the setFilterLiveReviewEvents() is used by the Live Review plugin)
And remove the associated session's s->filterLiveReviewEvents().

src/lib/ip/IPCore/Session.cpp
```diff
@@ -720,11 +720,6 @@ namespace IPCore
         }
     }
 
-    void Session::setFilterLiveReviewEvents(bool shouldFilterEvents)
-    {
-        m_filterLiveReviewEvents = shouldFilterEvents;
-    }
-
     bool Session::filterLiveReviewEvents() { return m_filterLiveReviewEvents; }
```
bernie-laberge, line 723: Remove Session::filterLiveReviewEvents() as well.

src/lib/ip/IPMu/CommandsModule.cpp
cedrik-fuoco-adsk: I need @pbergeron-adsk PR: https://github.com/AcademySoftwareFoundation/OpenRV/pull/963 since we are using that method in Mu

src/lib/ip/IPCore/Session.cpp
cedrik-fuoco-adsk: I need @pbergeron-adsk PR: https://github.com/AcademySoftwareFoundation/OpenRV/pull/963 since we are using that method in Mu

src/lib/ip/IPMu/CommandsModule.cpp
bernie-laberge: should we add a new session and command for convenience ?
(Otherwise you will have to duplicate lots of code)
like 
`void filterEventCategory(const std::string& category, bool filterEvent /*true to disable, false to enable*/);`

cedrik-fuoco-adsk: Since it is temporary, I think it's not much duplication. I prefer this instead of adding more method that will get removed later on.

src/lib/app/RvCommon/RvTopViewToolBar.cpp
cedrik-fuoco-adsk: Good idea! fixed!

src/lib/ip/IPMu/CommandsModule.cpp
cedrik-fuoco-adsk: I made the change. Let me know what do you think. I think it is fine since we are going to remove this. @bernie-laberge

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
cedrik-fuoco-adsk: Good idea! fixed!

src/lib/app/RvCommon/RvTopViewToolBar.cpp
bernie-laberge: Thanks !

src/lib/ip/IPMu/CommandsModule.cpp
bernie-laberge: All good. Thanks !

src/lib/ip/IPCore/IPCore/Session.h
```diff
@@ -30,6 +30,34 @@ namespace IPCore
...
     class PropertyInfo;
 
+    //
+    //  Event Category Constants
+    //
+    //  Central definition of all event category names used throughout RV.
+    //  These categories are used by the event filtering system to control
+    //  which UI actions are enabled/disabled.
+    //
+    namespace EventCategories
```
eloisebrosseau, line 41: You should modernize how you declare the categories const variables. It should be more like this:
```suggestion
    namespace EventCategories
    {
        inline constexpr std::string_view annotate = "annotate_category";
    }
```
Instead of const char*, you should use string_view. Using all caps is for macros in C++. I would also remove `category` from the name since it's already in an EventCategories namespace, but since `export` is a keyword, this one might create an issue. If you decide to keep `category` in the name, it should at least match the really name so `annotateCategory`, not `category_annotate`. Use inline so that it doesn't violates ODR.

src/lib/ip/IPCore/IPCore/Session.h
```diff
@@ -30,6 +30,34 @@ namespace IPCore
...
+    {
```
eloisebrosseau, line 42: Instead of accessing each category one by one when you need them all, I think you should add a function to do so here. It should be less error prone that way and if a category is added, it should be straight forward to add it directly in the list below instead of trying to find where we manually set the categories individually when we actually want all of them. That way, you could loop over the array in CommandsModule.cpp instead of enabling and disabling each category one by one.

```suggestion
    {
        inline constexpr auto all_categories() {
            return std::array{annotate, clear, ...};
        }
    }
```

src/lib/app/RvCommon/RvCommon/RvBottomViewToolBar.h
```diff
@@ -104,7 +112,7 @@ namespace Rv
         QIcon m_volumeHighIcon;
         QIcon m_volumeHighMutedIcon;
 
-        std::unordered_map<QAction*, QString> m_liveReviewFilteredActions;
+        std::vector<ActionCategoryMapping> m_actionCategoryMappings;
```
eloisebrosseau: Minor but if the list is always the same length, I think this should be an std::array, not an std::vector.

src/lib/ip/IPMu/CommandsModule.cpp
```diff
@@ -4143,12 +4143,54 @@ namespace IPMu
         NODE_RETURN(stype->allocate(r));
     }
 
+    // This is for backwards compatibility with the old live review plugin and latest OpenRV.
```
eloisebrosseau, line 4146: I see this is temporary but I think it would be beneficial to disable and enable all categories without listing them one by one like I suggested in Session.h

src/lib/ip/IPCore/IPCore/Session.h
cedrik-fuoco-adsk: Modernized! Done

src/lib/ip/IPCore/IPCore/Session.h
cedrik-fuoco-adsk: Modernized! Done

src/lib/ip/IPMu/CommandsModule.cpp
cedrik-fuoco-adsk: Fixed

src/lib/app/RvCommon/RvCommon/RvBottomViewToolBar.h
cedrik-fuoco-adsk: Agree, they should not change often. Done!

cedrik-fuoco-adsk:
All comments have been addressed!

I am not sure why there are cancelled CI jobs.

src/lib/ip/IPCore/IPCore/Session.h
```diff
@@ -30,6 +31,59 @@ namespace IPCore
...
+        inline constexpr std::string_view presentationCategory = "presentation_category";
+        inline constexpr std::string_view sessionmanagerCategory = "sessionmanager_category";
+        inline constexpr std::string_view sourceCategory = "source_category";
+        inline constexpr std::string_view systemCategory = "system_category";
+        inline constexpr std::string_view screeningroomCategory = "screeningroom_category";
+        inline constexpr std::string_view unclassifiedCategory = "unclassified_category";
+        inline constexpr std::string_view viewmodeCategory = "viewmode_category";
+        inline constexpr std::string_view wipesCategory = "wipes_category";
+
+        // constexpr is implicitly inline for clang, but not necessarily inline for other compilers.
```
eloisebrosseau: Oops sorry about that! Constexpr is implicitly inline for functions for pretty much all the compilers so I think you can remove `inline` and only keep `constexpr` here

src/lib/ip/IPCore/Session.cpp
```diff
@@ -757,11 +752,11 @@ namespace IPCore
                       m_disabledEventCategories.end(), category)
             == m_disabledEventCategories.end())
         {
-            m_disabledEventCategories.push_back(category);
+            m_disabledEventCategories.push_back(std::string(category));
```
eloisebrosseau, line 755: Is there a reason why m_disabledEventCategories could hold std::string_views instead of std::strings? You wouldn't have to do an explicit conversion here if m_disabledEventCategories is only holding a view to the strings and never modifying them

cedrik-fuoco-adsk: I think doing a copy make sense here (to have the ownership of the string) since the string can come from Mu/Python and we are passing `category->c_str()` to `enableEventCategory`/`disableEventCategory`.

src/lib/ip/IPCore/IPCore/Session.h
cedrik-fuoco-adsk: fixed

src/lib/ip/IPCore/Session.cpp
eloisebrosseau: I'm pretty sure you just need to change m_disabledEventCategories to hold std::string_view instead of std::string: `std::vector<std::string_view> m_disabledCategories;`

cedrik-fuoco-adsk:
Redundant because of https://github.com/AcademySoftwareFoundation/OpenRV/pull/971

PR #972, opened 2025-11-04
Fix building OTIO wheel on windows

rogernelson:
### Summarize your change.
This fixes building the opentimelineio wheel on Windows builds. In PR https://github.com/AcademySoftwareFoundation/OpenRV/pull/910 we pinned OTIO to the commit just before OTIO upgraded their pybind dependency, since this broke the build of RV.  To unblock us from upgrading the OTIO version, this PR introduces a fix for that issue. 

The problem with the pybind upgrade is that pybind changed the way it was finding python at it was no longer finding out version of Python.  The fix is to use the CMake envars for the include path, lib path and executable path so the pybind can find our python location.

I updated our requirements.txt file to use the latest main branch from OTIO, as it was doing before.

So we can use the latest OTIO version

Tested building on Windows 11.


cedrik-fuoco-adsk (approved):
LGTM!

bernie-laberge (approved):
LGTM
Thank you so much @rogernelson !

PR #974, opened 2025-11-06
SG-41029: Fix crash on exception

pbergeron-adsk:
### Summarize your change.

As reported by a community member, if the preloader receives an invalid filename to load, it will cause an exception and try to delete a uninitialized pointer, causing a crash.

Initialize member to default values (eg: nullptr) in class definition of the movie reader

macOS

### DIFF
--- a/docs/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-five.md
+++ b/docs/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-five.md
@@ -192,6 +192,8 @@ Some events will originate from RV itself. These include things like new-source 
 | after-graph-view-change | nodename |  | The current view node changed. |
 | new-node | nodename |  | A new view node was created. |
 | graph-new-node | nodename | nodename protocol version groupname | A new node of any kind was created. |
+| set-current-annotate-mode-node | nodename |  | Set the Paint node to use for the _currentNode in the annotate mode package. Calling this event with an empty string will reset its behaviour. |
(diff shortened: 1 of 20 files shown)


bernie-laberge (approved):
LGTM

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -54,6 +54,7 @@ namespace TwkMovie
     GenericIO::Preloader::Reader::Reader(std::string_view filename,
                                          const Movie::ReadRequest& request)
     {
+        m_movieReader = nullptr;
```
rogernelson: I think these would be better off being initialized in the header at declaration (like `MovieReader m_movieReader{nullptr};`)  to avoid the same problem if someone adds another ctor, but this will do the trick...

bernie-laberge: As per @rogernelson's comment, you should initialize it in the 
OpenRV/src/lib/image/TwkMovie/TwkMovie/MovieIO.h header:
`m_movieReader{nullptr}`

eloisebrosseau: I was going to comment the same thing. `m_movieReader` should be initialized in the header file.

pbergeron-adsk: I fixed it in the class definition, so resolving.

pbergeron-adsk: fixed

PR #977, opened 2025-11-07
Fix help menu

eloisebrosseau:
### Summarize your change.

Remove `.items` in the menuList of the "Help" submenu. Note that the same line without the `.items` is in the Commercial RV version of the help menu mode file.

When launching Open RV, the following error was being displayed in the console: `openrv_help_menu_mode.mu, line 157, char 42: Unresolved member reference to "items" in type "MenuItem[]"`

Launching Open RV on macOS 15.6.1 with the fix removed the error in the console.


pbergeron-adsk:
go

PR #979, opened 2025-11-07
TEST TEST: QPainter test for internal use only 

pbergeron-adsk:
This is a test PR for the team to evaluate the possibility of rendering overlays on top of the GLView using QPainter as opposed to OpenGL.

image

### Linked issues

SG-xxxxx

### Summarize your change.

Two changes: 
* Support QPainter/QPainterPath, and missing QPen from having bindings
* Create a new rendering path, After paintGL() has completed, to draw overlays again but with QPainter
* Added a test crosshair in the HUD's (i) command to see if it works.
* it works! 

Internal reasons.

macOS. 


pbergeron-adsk:
Archiving for future referemce.

PR #982, opened 2025-11-11
Unify OTIO versions, pin to 0.18.1

rogernelson:
### Summarize your change.
Currently OTIO is 0.16 on Window debug and the latest main branch for everything else. This branch does two things:
- unify the version between all OSes
- pin the version to 0.18.1.

Windows debug builds the OTIO wheel from source since there are no pre-built debug Windows wheels available in pypi, and we can't use the release one since Windows doesn't allow mixing release and debug libraries. As a result, it's version was specified outside of the requirements.txt file we use for everything else to install from pypi.  It's version was far behind the other OSes likely because whenever we updated it, we thought doing so in a single place would take care of everything.

To remove the chance of this happening again, I removed the requirements.txt file we use for installing our python build dependencies and instead auto-generated it from a template, using the opentimelineio version as an input variable from CMake so the version is now shared between all builds.

Lastly, I pinned the OTIO version to 0.18.1 which was just released this week. It's risky to use the latest main branch since one bad commit could break OpenRV.  

Build stability and consistency.

Building on  Windows + Linux

### DIFF
--- a/cmake/dependencies/python3.cmake
+++ b/cmake/dependencies/python3.cmake
@@ -254,14 +254,17 @@ ELSE() # Not WINDOWS
 ENDIF()
 
 # Generate requirements.txt from template with the OpenTimelineIO version substituted
-CONFIGURE_FILE(
+SET(_requirements_input_file
     "${PROJECT_SOURCE_DIR}/src/build/requirements.txt.in"
+)
+SET(_requirements_output_file
     "${CMAKE_BINARY_DIR}/requirements.txt"
-    @ONLY
 )
 
-SET(_requirements_file
-    "${CMAKE_BINARY_DIR}/requirements.txt"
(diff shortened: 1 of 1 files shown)


cmake/dependencies/python3.cmake
cedrik-fuoco-adsk, line 1: Suggestion:

You could set a variable for `${PROJECT_SOURCE_DIR}/src/build/requirements.txt.in` and `${CMAKE_BINARY_DIR}/requirements.txt` since it's used multiple time.

rogernelson: `${CMAKE_BINARY_DIR}/requirements.txt` is already a variable (_requirements_file).  But I will do it for the other path.

bernie-laberge (approved):
Beautiful
Thanks!

PR #983, opened 2025-11-12
SG-40363 - Block individual annotation button based on categories

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Based on the filtering categories, individual buttons in the annotation panel can be disabled or enabled.

We want to have granular control over the button in the Annotation panel.

MacOS

### DIFF
--- a/src/lib/app/mu_rvui/rvui.mu
+++ b/src/lib/app/mu_rvui/rvui.mu
@@ -6233,7 +6233,7 @@ global bool debugGC = false;
         subMenu("Control", MenuItem[] {
             menuItem("Play", "key-down-- ", "playcontrol_category", ~togglePlayFunc, playState),
             menuItem("Stop", "", "playcontrol_category", ~stopFunc, rangeState),
-            menuItem("PingPong", "key-down--P", "playcontrol_category", ~togglePingPong, pingPongState),
+            menuItem("PingPong", "key-down--P", "playmode_pingpong_category", ~togglePingPong, pingPongState),
             menuItem("Play Once", "", "playcontrol_category", ~togglePlayOnce, playOnceState),
             menuItem("Step Forward", "key-down--right", "playcontrol_category", ~stepForward1, rangeState),
             menuItem("Step Backward", "key-down--left", "playcontrol_category", ~stepBackward1, rangeState),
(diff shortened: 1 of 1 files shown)


cedrik-fuoco-adsk:
The PR is ready to review. The implementation is done. @eloisebrosseau @bernie-laberge

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
```diff
@@ -344,11 +349,11 @@ namespace Rv
             {m_ghostAction, IPCore::EventCategories::annotateCategory, m_ghostAction->toolTip()},
             {m_backStepAction, IPCore::EventCategories::playcontrolCategory, m_backStepAction->toolTip()},
             {m_forwardStepAction, IPCore::EventCategories::playcontrolCategory, m_forwardStepAction->toolTip()},
-            {m_backPlayAction, IPCore::EventCategories::playcontrolCategory, m_backPlayAction->toolTip()},
+            {m_backPlayAction, IPCore::EventCategories::backplayCategory, m_backPlayAction->toolTip()},
```
eloisebrosseau: Shouldn't this be playback instead of backplay?

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
```diff
@@ -631,12 +686,18 @@ namespace Rv
...
         {
             QAction* a = m_playModeMenu->actions()[i];
-
-            if (!a->isEnabled())
-                continue;
-
             a->setCheckable(true);
-            a->setChecked(a->data().toMap()["enum"].toInt() == m_session->playMode());
+
+            if (a->isEnabled())
```
eloisebrosseau: Can you rename `a` to something a little bit more descriptive like `playModeMenuAction`

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
cedrik-fuoco-adsk: It's for the backward play button

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
cedrik-fuoco-adsk: Yes, I can rename it

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
eloisebrosseau: Since the normal play button is called forwardPlayAction, could we rename backPlayAction to backwardPlayAction and the category to backplayCategory? I feel like this name is really confusing and not consistent.

pbergeron-adsk: quick question, why are you changing it from "playcontrols_category" ?

cedrik-fuoco-adsk: We need granular controls over some buttons in the toolbar for Web to RV. @pbergeron-adsk 

@eloisebrosseau I renamed the category and variables.

PR #984, opened 2025-11-12
SG-40852: Improve paint speed degrading as you draw many annotations

pbergeron-adsk:
### Linked issues

SG-40852

### Summarize your change.

Pretty significant fix in the management of the annotation fbo cache for the current frame, which got broken some years ago.

In a nutshell, the rendered fbo cache should always have all of the paint commands (eg: annotation) rendered in it, except for the last paint command, in case we're currently editing it (eg: drawing a stroke).

Somehow, this cache management was really borked. It would fail to find a cache, then draw all annotations in the cache, keep the cache. At the next redraw, it would find the cache, render the start of a new stroke, but as you edited it, it wouldn't find the cache anymore, recreated the cache, found apply the last stroke, which would somehow invalidate the cache, and at the next update it would go through this cycle again. In other words, it would redraw everything, every time.

I've made a fix such that when a new stroke is added,  it will (unfortunately) still draw everything, but as the stroke updates, it will keep only rendering a single paint command on top of the cache, until you start a new stroke. 

Repainting all strokes when you start a new one, if you have many, like 250+ srokes, might take like 150ms, which is a definitely a lot, but it fixes the problem where you'd pay the  150ms at every single mouse move (ugh). 

Anyway, things should be be much better now.

customer complaints, and could reproduce this easily with rv_tests scripts.

macOS.

During the review, you should make sure that nothing got broken in the process. Try all tools: reaser, pen, dodge, burn, etc.

You should also verify that simulateneous paint strokes in live review collab mode still works.

In particular, @bernie-laberge, please make sure that your fix for recomputing the render ID still works.

During testing, please do the following:
- Start OLD RV  (eg:last week). Open media. Press F10 to show annotate tools
- Start NEW RV (this VT build). Open media. Press F10 also

Both RV, do:
-> Menu -> RV Tests -> Automatic Annotation Drawing -> start drawing circles.
(load rv tests package if not done so already)

Both RVs should start to draw circles.

Let it got for a few minutes. After a while, maybe 5-6-10 minutes, the OLD RV should be extremely slow. The NEW RV will have slowed down a bit, but not nearly as bad. (note: this will create hundreds and hundreds of strokes at that frame,


src/lib/ip/IPCore/ImageRenderer.cpp
```diff
@@ -4894,9 +4893,23 @@ namespace IPCore
...
+            // Process each command's hash individually and combine them
+            boost::hash<string> string_hash;
+            size_t hashValue = 0;
+            for (size_t i = 0; i < curCmdNum - 1; ++i)
+            {
+                ostringstream hash;
+                root->commands[i]->hash(hash);
+                size_t cmdHash = string_hash(hash.str());
+                hashValue ^= cmdHash + 0x9e3779b9 + (hashValue << 6) + (hashValue >> 2);
+            }
```
bernie-laberge: I wonder: Why is this fixing the issue ?
Shouldn't imageRenderID below already contain this hashing of all the paint commands ?
https://github.com/AcademySoftwareFoundation/OpenRV/blob/1e1f0d3c8f4d86e8d6c0e3c8190aa770bbb975d3/src/lib/ip/IPCore/IPImage.cpp#L1018

src/lib/ip/IPCore/ImageRenderer.cpp
```diff
@@ -4966,6 +4994,7 @@ namespace IPCore
         paintContext.hasStencil = hasStencil;
         paintContext.stencilBox = stencil;
         paintContext.lastCommand = root->commands.back();
+        paintContext.totalCommandsInCache = startCmd;
```
bernie-laberge: I don't think totalCommandsInCache is used anywhere.

pbergeron-adsk: I was using it for debug traces, and left it there because I thought it might be useful, but I've removed it.

src/lib/ip/IPCore/ImageRenderer.cpp
pbergeron-adsk: This is actually from an attenpt at getting the cache and hash at N-1 and N-2 (for all sorts of reasons). and I didn't fully understand (or had overlooked) that your previous fix would correct this by also looking at the hash.

I've removed the unnecessary code.

Issue #821, opened 2025-07-02
[Maintenance]: VFX Reference Platform CY25. (fixed: SG-41117)


cedrik-fuoco-adsk:
**Libraries dependencies:** 
- OpenEXR 3.3.x
- OpenColorIO 2.4.x
- Boost 1.85

manuelito02:
Hi there!
Are you planning a more recent version of OIIO? I know, it's not part of the VFX Reference Platform. But since OIIO 3 is already out for a while and other DCCs are moving forward, it might be helpful to consider this as well for RV. Any idea about the next update of RV with this issue included?
Greets,
Carlo

compaga:
Hi @manuelito02, we don't have it on the radar for now but I will add this specific request to the roadmap. May I ask, what specific feature  to OIIO 3 are you looking for?

manuelito02:
Hi @compaga !

Thanks for your quick response!

It's more about compatibility issues. I did not test yet a production environment with OIIO 3 installed, but we had issues in the past using an OIIO version that was different to the one from RV and then, we couldn't open RV anymore. And other DCCs are moving to OIIO 3 (e.g. [Gaffer](https://github.com/GafferHQ/gaffer)) and if we install an older version we can't open that anymore.

It's a bit of library compatibility juggling. So far, we could work with a single production environment per project without the need to setup anything particular for one software or another and ideally, we would like to keep it that way, although I'd admit that this is not a very strong argument to switch dependencies on your side which might be far from being trivial.

pbergeron-adsk:
Fixed with PR for SG-41117

Issue #823, opened 2025-07-02
Update to OpenColorIO 2.4.X


cedrik-fuoco-adsk:
- Update `ocio.cmake` for OpenColorIO 2.4.X
- Look for breaking changes in the release notes
- Make sure that OCIO code path are still working

pbergeron-adsk:
Fixed with PR for SG-41117

Issue #822, opened 2025-07-02
Update OpenEXR to 3.3.X


cedrik-fuoco-adsk:
Update `openexr.cmake` for OpenEXR 3.3.X

pbergeron-adsk:
Fixed with PR for SG-41117

PR #985, opened 2025-11-12
SG-41117: cmake use cy files for dependency versions and add experimental support CY2025 and CY2026

mcoliver:
### Summarize your change.
Moved hardcoded versions for various dependendencies out of individual cmake files and into a centralized file based on CY year.  There were also changes made to allow for years beyond 2024 rather than having to hardcode values for each year in various build files throughout the repo.  

The current codebase has checks for 2023 or 2024 littered across many files and it makes it difficult to understand what needs to be updated for each CY year and where you do it.  The code was also difficult to follow with macros defining vars for various years.

With this setup you can create a new CYYEAR.cmake file, copy the old one, update version numbers and even diff between them.   It makes it much easier to understand where to go to update packages.  Additionally it allows for things like CY2099.cmake which can have the most current releases of packages (perhaps ahead of a CY official designation) for people who like to run the latest dependencies with performance improvements and bug fixes.

Builds 2023, 2024, and 2025 have been verified on macOS.  

- Recently there was an OTIO commit that may conflict with this PR.
- CY2026 is still experimental.  QtWebEngine no longer is packaged with 6.8.3 so we have to go get it.
- patch files still need to be generated for new versions of openexr and python (though I would love to discuss why the patch file for openexr exists.  Covering up bad data in a file with black seems like it could result in confusion for users.  The file should be invalid if part of it is invalid.)
- I also added a helper function to rvcmds to select the CY year you want to build.  
- Finding Qt logic was updated in rvcmds
- My commit history is messy.  I can clean it up if necessary prior to merge but wanted to get this out there before more merge conflicts arise.
- I did not address CI for building 2025 and 2026.  That still needs to be configured.

### Other discoveries
RV builds on macOS still want to find things in /opt/homebrew.  This can easily cause build errors for users that have things like openexr installed through homebrew.  You can see this by greping for `/opt/homebrew` in the `_build` directory.  I tried various cmake vars to tell cmake to ignore /opt/homebrew but could never seem to get any to work.  Additionally there are things in homebrew that the build relies on like xz so you don't want to ignore everything.  The problem is if something like openexr4.3.0 is installed with homebrew, cmake for 2024 will attempt to link to that rather than the 3.2 that was compiled for that CY year.  For me I had to uninstall a bunch to get the build to work properly
```
brew uninstall jpeg-xl libheif imagemagick aom openexr openexr openssl boost ffmpeg imath libtiff
```

The resulting size of the RV.app build is quite large.  I believe it can be significantly trimmed down and will do that in a different PR.


cmake/dependencies/imgui.cmake
```diff
@@ -103,10 +103,16 @@ EXTERNALPROJECT_ADD(
...
 )
 
-RV_VFX_SET_VARIABLE(_qt_location CY2023 "${RV_DEPS_QT5_LOCATION}" CY2024 "${RV_DEPS_QT6_LOCATION}")
-RV_VFX_SET_VARIABLE(_find_qt_version CY2023 "Qt5" CY2024 "Qt6")
+SET(_qt_location 
+  ${QT_HOME}
+)
+SET(_find_qt_version 
+  "Qt${RV_DEPS_QT_MAJOR}"
+)
```
cedrik-fuoco-adsk, line 111: It should be `$ENV{QT_HOME}` if you want to use the env. variable since `QT_HOME` is not passed to CMake during the configure step.

mcoliver: Copy. Thank you. Question, would someone ever need to specify qt5 and qt6 locations for a single build? Or can we deprecate those qt specific envs in favor of a single universal one?

cedrik-fuoco-adsk: It's either Qt5 or Qt6, never both at the same time

mcoliver: Great. Made a new commit to clean that up as well.

mcoliver:
Yeah totally. I was going to make a second pr for the common thing. I tried
to keep it to the vfxplatform things but I suppose I also added otio and
oiio which are not official. Anything else?  Maybe this one could go
through and I'll fast follow with a more comprehensive common one including
moving those two envs?

eloisebrosseau:
I think OpenSSL could also be in the common file. @cedrik-fuoco-adsk, any other dependencies we could move there? 
We still need to validate your PR on our side, but I agree that having a second PR for the common dependencies would make sense since there are already a lot of changes in this one.

cedrik-fuoco-adsk:
I think I would move all non-VFX platforms dependencies into a `CYCOMMON.cmake`.

Here's the list (Will have to double check to make sure):
```
aja
atomic_ops
bmd
dav1d
doctest
expat
Ffmpeg
gc
glew
Imgui
Jpegturbo
Nanobind
OpenImageIO
Openjpeg
Openjph
Openssl
Pcre2 (windows only)
Pyimgui
Pyimplot
Raw
Spdlog
Tiff
Webp
Zlib
Vcpkg
OTIO
```

There are special cases like FFmpeg where the extra logic to select between 6, 7 or 8 will have to move to `CYCOMMON.cmake` too.

mcoliver:
sounds good.  I started migrating, and then accidentally just pushed to
this branch instead of a new one.  Apologies.  Should I pull it out or
should we include it?  For ffmpeg I'd also like to make a slight
modification to the logic to allow users to use any ffmpeg version they
wish to use (experimental and unsupported of course).

cedrik-fuoco-adsk:
Personally, I think it is manageable and I like having everything in one go. What do you think @eloisebrosseau 

@mcoliver Can you really do that for FFmpeg? We have special logic to set the right names for all the FFmpeg librairies. How are you going to handle that with any FFmpeg version?

cmake/defaults/CYCOMMON.cmake
```diff
@@ -0,0 +1,135 @@
...
+SET(RV_DEPS_DOCTEST_DOWNLOAD_HASH "a7948b5ec1f69de6f84c7d7487aaf79b")   
+
+# expat
+# https://github.com/libexpat/libexpat
+SET(RV_DEPS_EXPAT_VERSION "2.6.3")
+SET(RV_DEPS_EXPAT_DOWNLOAD_HASH "985086e206a01e652ca460eb069e4780")
+
+# ffmpeg
+# https://github.com/FFmpeg/FFmpeg
+# TODO: cmake/dependencies/ffmpeg.cmake defines library names and
```
cedrik-fuoco-adsk: What's the additional work?

cmake/defaults/cxx_defaults.cmake
```diff
@@ -61,28 +61,27 @@ ENDIF()
...
-# Add preprocessor variable for use in the code and cmake variable to decide the version of an external dependencies based on the VFX platform. Note that the
-# macro in rv_vfx.cmake are dependant on those RV_VFX_CY20XX.
-IF(RV_VFX_PLATFORM STREQUAL CY2024)
-  SET(RV_VFX_CY2024
-      ON
-  )
-  ADD_COMPILE_DEFINITIONS(RV_VFX_CY2024)
+# Include 3rd party Dependencies
+INCLUDE(CYCOMMON.cmake)
+
```
cedrik-fuoco-adsk, line 74: Shouldn't `CYCOMMON.cmake` be included before the VFX platform? We'd like the specific `CYXXX.cmake` to overwrite the versions in `CYCOMMON.cmake` if needed.

mcoliver: good catch. swapped

cedrik-fuoco-adsk, line 74: Also, I think the configure step is failing in the CI because of the `.cmake`. It should be `INCLUDE(CYCOMMON)` @mcoliver

mcoliver:
I believe this should resolve #821 #822 #823

cedrik-fuoco-adsk:
Hi @mcoliver,

Do you have time to resolved the conflict? It's due to a recent commit that changed `python3.cmake` (https://github.com/AcademySoftwareFoundation/OpenRV/commit/ce646969ed8cb5fa6ac829bf500164a35afd737f)

We should have time to test this internally soon.

cedrik-fuoco-adsk:
We tested it locally, and we are ready to merge it. It would be great if you have time to fix the conflicts. @mcoliver 

If you don't have time, I can take care of it, but I'll need to create a new PR since I won't be able to write in your branch.

mcoliver:
> We tested it locally, and we are ready to merge it. It would be great if you have time to fix the conflicts. @mcoliver
> 
> If you don't have time, I can take care of it, but I'll need to create a new PR since I won't be able to write in your branch.

Conflicts resolved

cedrik-fuoco-adsk (approved):
Thank you @mcoliver for the great contribution!

PR #994, opened 2025-11-19
SG-40625 - Play Forward/Pause button stops responding when playing media files (Linux)

cedrik-fuoco-adsk:
SG-40625 - Play Forward/Pause button stops responding when playing media files (Linux)

### Linked issues
n/a

There was a stability problem where calling `stop()` or `reset()` on the `QAudioOutput` object would cause audio playback to break after multiple play/stop cycles. To work around this, suspend() was used instead, which solved the instability. However, this introduced a second problem: `suspend()` does not clear the audio device's internal buffer. This resulted in an audio bleed issue, where sound from the previously played media could be heard for a fraction of a second when starting a new one.

This change introduces a new flushing mechanism that resolves the audio bleed on Linux while avoiding the unstable `stop() `and `reset()` calls.

When playback is stopped, instead of immediately suspending the audio device, a new flushing process is initiated. The audio device is kept active momentarily while it is fed with silent audio data, completely overwriting the contents of its internal buffer. Once the buffer is flushed, the device is safely suspended.

This ensures that the audio buffer is clear before the next playback begins, eliminating audio bleed. Because this approach continues to use `suspend()` to halt the device, it maintains the stability of the audio system. 

**This logic is confined to the Linux platform and does not affect the behavior on MacOS or Windows.**

### DIFF
--- a/src/lib/audio/QTAudioRenderer/QTAudioRenderer/QTAudioRenderer.h
+++ b/src/lib/audio/QTAudioRenderer/QTAudioRenderer/QTAudioRenderer.h
@@ -58,7 +58,7 @@ namespace IPCore
 
     private:
         QTAudioThread& m_thread;
-        bool m_silence;
+        bool m_silence = false;
     };
 
     class QTAudioOutput : public QAudioSink
@@ -97,7 +97,7 @@ namespace IPCore
         QAudioFormat& m_format;
         QTAudioIODevice& m_ioDevice;
         QTAudioThread& m_thread;
-        bool m_isFlushing;
+        bool m_isFlushing = false;
         qint64 m_flushedBytes;
     };
 


src/lib/audio/QTAudioRenderer/QTAudioRenderer/QTAudioRenderer.h
```diff
@@ -48,12 +48,17 @@ namespace IPCore
...
+
+        void stopProducingSilence() { m_silence = false; }
+
     public slots:
         void resetDevice();
         void stopDevice();
 
     private:
         QTAudioThread& m_thread;
+        bool m_silence;
```
bernie-laberge: You could initialize to false here

src/lib/audio/QTAudioRenderer/QTAudioRenderer/QTAudioRenderer.h
```diff
@@ -82,6 +97,8 @@ namespace IPCore
         QAudioFormat& m_format;
         QTAudioIODevice& m_ioDevice;
         QTAudioThread& m_thread;
+        bool m_isFlushing;
```
bernie-laberge: you could initialize to false here

cedrik-fuoco-adsk: done

cedrik-fuoco-adsk: done

PR #997, opened 2025-11-21
fix: SG-40934: RVDisplayGroup's transfer function not being set to sRGB when loading media via pyeval

pbergeron-adsk:
### Linked issues
SG-40934

### Summarize your change.

Using -pyeval to load movies stopped setting the Display Transfer Function to sRGB, leaving it at "None". This occurred because RV creates an **RVOutputGroup** node in command-line mode **instead** of **RVDisplayGroup**, and the Python code only checked for RVDisplayGroup nodes. 

It was only later when the UI started (and after the data had been loaded) that actual RVDisplayGroups got created, but they weren't being configured. 

I'm not sure how this ever worked in 2021.3 (it was in the days before OpenRV), but this fix seems to do the trick. 

The fix: 
* (1) modified the code to handle both RVOutputGroup and RVDisplayGroup node types, and ;
* (2) immediately configure any new RVDisplayGroups when they're created after the UI starts. 

Now the Display Transfer Function on the RVDisplayGroup is set to sRGB in all scenarios.

See below image for a full explanation of what happened.

image

RVDisplayGroup's transfer function was not being set properly.

macOS

### DIFF
--- a/src/plugins/rv-packages/source_setup/source_setup.py
+++ b/src/plugins/rv-packages/source_setup/source_setup.py
@@ -360,8 +360,14 @@ class SourceSetupMode(rvtypes.MinorMode):
             dICCNode = groupMemberOfType(dpipe, "ICCDisplayTransform")
             if dICCNode:
                 commands.setIntProperty(dICCNode + ".node.active", [0], True)
-
-            if dgroup is not None and commands.nodeType(dgroup) == "RVDisplayGroup" and dgroup not in dnodesWithProfile:
+            # BUGFIX: Also handle RVOutputGroup in addition to RVDisplayGroup
+            # When RV is run in command-line mode (e.g., via -pyeval), it uses RVOutputGroup
+            # instead of RVDisplayGroup for the default display output
+            if (
+                dgroup is not None
+                and commands.nodeType(dgroup) in ["RVDisplayGroup", "RVOutputGroup"]
+                and dgroup not in dnodesWithProfile
+            ):
                 dICC = commands.getStringProperty(dgroup + ".device.systemProfileURL")
                 if dICCNode and len(dICC) == 1 and not dICC[0] == "":
                     dICCPath = urlparse(dICC[0].replace("%20", " ")).path
@@ -504,6 +510,9 @@ class SourceSetupMode(rvtypes.MinorMode):
         event.reject()
         if commands.nodeType(event.contents()) == "RVDisplayGroup":
             self._haveNewDisplayGroups = True
+            # BUGFIX: When a new RVDisplayGroup is created AFTER initial setup
+            # (e.g., when UI starts after -pyeval), configure it immediately
+            self.setDisplayFromProfile()
 
     def __init__(self):
         self._readingSession = False


src/plugins/rv-packages/source_setup/source_setup.py
```diff
@@ -504,6 +510,9 @@ def checkForDisplayGroup(self, event):
         event.reject()
         if commands.nodeType(event.contents()) == "RVDisplayGroup":
             self._haveNewDisplayGroups = True
+            # BUGFIX: When a new RVDisplayGroup is created AFTER initial setup
+            # (e.g., when UI starts after -pyeval), configure it immediately
+            self.setDisplayFromProfile()
```
bernie-laberge, line 515: Can't this interfere with the ocio_source_setup.py if OCIO mode is enabled ?

src/plugins/rv-packages/source_setup/source_setup.py
```diff
@@ -360,8 +360,14 @@ def setDisplayFromProfile(self):
             dICCNode = groupMemberOfType(dpipe, "ICCDisplayTransform")
             if dICCNode:
                 commands.setIntProperty(dICCNode + ".node.active", [0], True)
-
-            if dgroup is not None and commands.nodeType(dgroup) == "RVDisplayGroup" and dgroup not in dnodesWithProfile:
+            # BUGFIX: Also handle RVOutputGroup in addition to RVDisplayGroup
+            # When RV is run in command-line mode (e.g., via -pyeval), it uses RVOutputGroup
+            # instead of RVDisplayGroup for the default display output
+            if (
```
bernie-laberge, line 366: I would really like to know what changed which led to this regression. I'll do some research in the original RV git repo and report back.

bernie-laberge:
FYI: Status report of my investigation to figure out the source of this regression:
As confirmed by @savoiei , this appears to be a regression originating from RV 2022.0.0 because RV 2021.1.0 does not have this issue.
But RV 7.7.1 does have the issue !
I was wondering Why. So I compared the code from RV 2022.0.0 and RV 2021.1.0 and check what I found:
In RV 2022.0.0, progressive source loading was disabled by default!
And in RV 7.7.1, there was no progressive source loading.
I tested it and it does work in the latest RV if you enable progressive source loading:
`export RV_PROGRESSIVE_SOURCE_LOADING=1`

Which makes sense because with progressive source loading, the source-group-complete is delayed long enough so that the RVDisplayGroup gets created. Without progressive source loading, the pyeval gets executed before the creation of RVDisplayGroup so the code in source_setup.py has no chance to be executed properly thus the bug.

src/plugins/rv-packages/source_setup/source_setup.py
bernie-laberge: I think it would be safer to move the [pyeval](https://github.com/AcademySoftwareFoundation/OpenRV/blob/65c64729b2cedfa8ec53c492353e0f7c610c185b/src/lib/app/RvApp/RvSession.cpp#L2360) further down the line where we know that the RVDisplayGroup has been properly initialized.
Like here for example:
https://github.com/AcademySoftwareFoundation/OpenRV/blob/65c64729b2cedfa8ec53c492353e0f7c610c185b/src/lib/app/RvCommon/RvApplication.cpp#L879

pbergeron-adsk:
>Without progressive source loading, the pyeval gets executed before the creation of RVDisplayGroup so the code in source_setup.py has no chance to be executed properly thus the bug.

Ok, I understand this. Now, what shall we do about this PR?  Do we fix it as per my suggested code changes, or do I try to move the pyeval code to a later spot in RvDocument as you suggested above? 

Shall we just close it because we intend to deprecate progressive loading at some point? @bernie-laberge

PR #1001, opened 2025-11-26
Fix ERROR: while loading source module inspector.mu

bernie-laberge:
### Linked issues
NA

When activating the Color Inspector, its menu failed to initialize properly and the following error was reported:
```
ERROR: while loading source module "../Mu/inspector.mu"
ERROR: syntax error
```

This is a recent regression

### Summarize your change.
Removed the trailing comma at the end of the color inspector's menu definition which was the source of the error
(At the end of "Close Inspector")

Also in this commit: Fixed the eventPattern, category inversion for the 8bits/10bits/12bits/16bits menu items.

Thanks to @markreidvfx for reporting it !

Successfully tested on macOS


pbergeron-adsk (approved):
lgtm, sorry about that!

PR #1003, opened 2025-11-26
SG-40769 - Multiple RVLinkLauncher unresponsive dialogue windows can be opened when clicking away from the window

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.

Fixed multiple dialog windows appearing when opening rvlink URLs. Added logic to terminate other running RVLinkLauncher instances on startup and automatically close any existing dialog when a new rvlink URL is received. This ensures only one chooser dialog is visible at a time - new rvlink URLs close the previous dialog and display a fresh one with the new URL.

When users clicked multiple rvlink URLs from a browser, each click would create a new modal dialog that stacked on top of previous ones. Only the topmost dialog remained interactive, leaving users unable to interact with or close the underlying dialogs. This occurred because: (1) command-line launches created separate process instances that didn't coordinate with each other, and (2) browser-launched URLs were sent as Apple Events to an already-running instance that was blocked in a modal loop. The fix improves user experience by ensuring rvlink URLs are processed one at a time with a single, responsive dialog window.

Affects MacOS only


cedrik-fuoco-adsk:
Do you mind reviewing @eloisebrosseau @bernie-laberge ? I changed how I did the fix.

- **Single instance:** RVLinkLauncher now runs as a single-instance app, Multiple `rvlink://` activations are routed to the same process instead of starting separate launchers.

- **Dialog reuse:** When a new `rvlink://` URL is received while the chooser dialog is already open, the existing `NSAlert` is brought to the front and its “Opening: …” text is updated to the latest URL, instead of opening additional dialogs.

- **Latest URL respected:** The selected RV application always opens the most recent `rvlink://` URL that was clicked.

- **Terminal and browser behave the same:** Invoking RVLinkLauncher `rvlink://...` from the terminal forwards the URL to LaunchServices (NSWorkspace openURL:), so both terminal and browser use the same single-instance launcher and chooser window.

PR #1009, opened 2025-12-01
Fix: Use RV's python version to build wheels

rogernelson:
### Summarize your change.

When building wheels with python, CMake would use whichever Python version it found first, which was usually your system python. This poses a problem when its a different version that RV's python if there are any compiled libraries built (like with opentimelineio for example), since when you try to import the package it will look module built against the matching version of python. 

To fix this issue, we will pass the python executable path to pip so we built it with RV's python. We will also use `--no-cache-dir` to ensure we always built the package. Even if our python version does have a cached version available, we might be building it with different compilers or different versions of shared libraries, so let's not take that chance and just force pip to always build with RV's python.  We will also use and `--force-reinstall` so we will always build OTIO when we build Python to ensure no underlying changes to RV's python make this issue reappear.  

Build fix

Reproduced the issue on Mac OS, and verified the fix

### DIFF
--- a/cmake/dependencies/python3.cmake
+++ b/cmake/dependencies/python3.cmake
@@ -244,6 +244,14 @@ ELSE() # Not WINDOWS
   )
 ENDIF()
 
+# Set the appropriate library for CMAKE_ARGS based on platform
+# Windows needs the import library (.lib), Unix needs the shared library (.so/.dylib)
+IF(RV_TARGET_WINDOWS)
+  SET(_python3_cmake_library ${_python3_implib})
+ELSE()
+  SET(_python3_cmake_library ${_python3_lib})
+ENDIF()
+
 # Generate requirements.txt from template with the OpenTimelineIO version substituted
 SET(_requirements_input_file
     "${PROJECT_SOURCE_DIR}/src/build/requirements.txt.in"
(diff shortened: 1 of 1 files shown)


cmake/dependencies/python3.cmake
```diff
@@ -274,18 +276,26 @@ IF(RV_TARGET_WINDOWS)
...
+  # cached wheels that may have been built for a different Python version.
+  # This ensures ABI compatibility with our custom-built Python.
+  # --no-cache-dir: Don't use pip's wheel cache (prevents using wheels built for wrong Python version)
+  # --force-reinstall: Reinstall packages even if already installed (ensures fresh build)
+  # CMAKE_ARGS: Tell OTIO's CMake build which Python to use (prevents it from finding system Python via pyenv/etc)
   SET(_requirements_install_command
-      "${_python3_executable}" -m pip install --upgrade -r "${_requirements_output_file}"
+      ${CMAKE_COMMAND} -E env
+      "CMAKE_ARGS=-DPYTHON_EXECUTABLE=${_python3_executable}"
+      "${_python3_executable}" -m pip install --upgrade --no-cache-dir --force-reinstall -r "${_requirements_output_file}"
```
cedrik-fuoco-adsk: Do you think we should pass PYTHON_INCLUDE_DIR and PYTHON_LIBRARY to make sure it uses the right includes and libraries? Basically, the same thing we are doing for Windows. @rogernelson

rogernelson: Good point. It fixed the problem without that for me, but yes, we should be on the safe side and do that.

rogernelson: I'll just unifying all the build commands together, there's no longer any point in having anything separate besides adding in one variable for debug builds.  I think some cleanup is in order here.

bernie-laberge (approved):
LGTM

PR #1013, opened 2025-12-02
Build all python package dependencies

rogernelson:
### Summarize your change.

After fixing ABI compatibility issues with OTIO arising from the fact that we are building our own custom python, I checked the other packages and there are indeed others that also have binary components (PyOpenGL_accelerate, pydantic, and cryptography).

Because of this, we should just always build the packages from source with the same compiler we use to build python, and we can do that using the --no-binary option from pip. Although this might be a little bit slower, it will ensure that any package with binary components won't have ABI compatibility issues and for pure python packages it has no effect anyway.

ABI safety

Built on Mac OS 26.1.

### DIFF
--- a/cmake/dependencies/python3.cmake
+++ b/cmake/dependencies/python3.cmake
@@ -31,18 +31,12 @@ SET(RV_DEPS_PYTHON_VERSION_SHORT
 )
 
 # This version is used for generating src/build/requirements.txt from requirements.txt.in template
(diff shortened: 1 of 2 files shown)


cedrik-fuoco-adsk (approved):

cedrik-fuoco-adsk:
Closing in favour of https://github.com/AcademySoftwareFoundation/OpenRV/pull/1014

PR #1017, opened 2025-12-05
SG-40514: Add env var to bypass Clear All confirmation dialog

bernie-laberge:
### Linked issues
Fixes NA

The reason for this change is that we are adding Annotation automation tests and in order to test the "Clear Annotations on All The Frames" functionality, we needed a way to bypass the confirmation dialog.

### Summarize your change.
This commit adds a new env var which when defined, will  skip the confirmation dialog when executing the function : "Clear Annotations on All The Frames".

`export RV_SKIP_CONFIRMATIONS=1 `

Note that for efficiency, it is evaluated in the annotate_mode constructor.

Successfully tested on macOS.

### DIFF
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -171,7 +171,8 @@ class: AnnotateMinorMode : MinorMode
 
     char              _cursorChar;
 
-    int              _hideDrawPane;
+    int               _hideDrawPane;
+    bool              _skipConfirmations;
 
     // filtering control based on time and dx/dy deltas during drag events
     // to prevent sending ridiculous number of mouse/tablet events if using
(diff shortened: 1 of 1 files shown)


src/plugins/rv-packages/annotate/annotate_mode.mu
```diff
@@ -171,7 +171,8 @@ class: AnnotateMinorMode : MinorMode
 
     char              _cursorChar;
 
-    int              _hideDrawPane;
+    int               _hideDrawPane;
+    bool              _skipClearAllConfirmation;
```
eloisebrosseau: Minor but it looks like you added an extra space in front of `_hideDrawPane` and `_skipClearAllConfirmation`. Was this intentional?

bernie-laberge: Yes it was intentional : so that the _hideDrawPane member aligns with the other parameters.
Screenshot 2025-12-05 at 12 32 24 PM

bernie-laberge: _hideDrawPane was misaligned before.

bernie-laberge:
> In general why dont you have a getenv for SKIP_CONFIRMATIONS_FOR_UNIT_TESTS?
> 
> Because I suppose that would also affect the "Clear Session" of the presenter if they're in a live review session, and also, I'm guessing there are other places in the code where we might want to have confirmation dialogs, and you could just use the same env var for everything.


This is an excellent idea @pbergeron-adsk !
I propose to prefix it with RV_ and make it even more generic.
How about:
`RV_SKIP_CONFIRMATIONS`

pbergeron-adsk:
> This is an excellent idea @pbergeron-adsk ! I propose to prefix it with RV_ and make it even more generic. How about: `RV_SKIP_CONFIRMATIONS`

Yes, that's fine! Go ahead and make the change.

PR #1020, opened 2025-12-11
Fix clearAllSlot

eloisebrosseau:
### Summarize your change.

Add missing `}` for `if(!_skipConfirmations)`.

While resolving the conflict between #1017 and #1016, the indentation of the code got changed and it resulted in a missing closing brace. Because of this issue, the following error is displayed in the console:
`annotate_mode.mu, line 2210, char 41: method declaration not allowed in this context.
Assertion failed: ((nparams == 0 && params == 0) || (nparams != 0 && params != 0)), function init, file Function.cpp, line 69.`


bernie-laberge (approved):
Exactly the same as mine !
Looks Great !

rogernelson (approved):

PR #1023, opened 2025-12-16
feat: SG-41515: Add support for optional major.minor.PATCH version number in packages and bundles

pbergeron-adsk:
### Linked issues

SG-41515

### Summarize your change and reason for change

Package and bundle files only had major.minor version numbers. This made it extremely cumbersome and awkward to issue patch releases with a patch number for hotfixes. 

Now the package manager handles an optional major.minor.PATCH number. It is backwards compatible with just major.minor, and will assume ".0" for patch number in this case.

Users reported that when we added a  "version" tag in the PACKAGE file with an extra dot/number, it didn't work. But that's because the rvpkg executable appears to load the PACKAGE file, and will blindly create a file with the extra dot/number. But the rvpkg **loader** validates the filename against a regular expression, and the filename with the extra dot/number failed the regex check.

So it's not that the "version" tag didn't support the extra dot/number. It's that the rvpkg **loader** didn't like the filename format... sigh.

In any case, the fix makes RV able to load old filenames with major.minor, and major.minor.patch also...

... but unfortunately, old RVs will still be unable to read **pluginname_major.minor.patch.rvpkg** files, just "pluginname_major.minor.rvpkg" files.

note:

As an alternative to breaking old RVs, maybe we can rename the rvpkg files by adding a service pack number (sp) to the version number, like this:

```
if patch == 0, then 
   filename = plugin_name-major.minor.rvpkg
else
   filename = plugin_name_sp{patch}.major.minor.rvpkg

```
but it's unclear if this would work correctly for install/uninstall of old bundles and packages that need conflict resolutions.

macOS.

### DIFF
--- a/src/lib/app/RvPackage/PackageManager.cpp
+++ b/src/lib/app/RvPackage/PackageManager.cpp
@@ -67,27 +67,6 @@ namespace Rv
         }
 
         string fullPath = outputPath + "/" + filename;
-        
-        // Check if this is a directory entry (ends with /)
-        if (fullPath.back() == '/')
-        {
-            unzCloseCurrentFile(uf);
-            return ""; // Skip directory entries
-        }
-        
-        // Create parent directories if needed
-        QFileInfo fileInfo_qt(QString::fromStdString(fullPath));
-        QDir parentDir = fileInfo_qt.dir();
-        if (!parentDir.exists())
-        {
-            if (!parentDir.mkpath("."))
-            {
-                cerr << "ERROR: Unable to create directory: " << parentDir.absolutePath().toStdString() << endl;
-                unzCloseCurrentFile(uf);
-                return "";
-            }
-        }
-        
         FILE* outFile = std::fopen(fullPath.c_str(), "wb");
         if (outFile == nullptr)
         {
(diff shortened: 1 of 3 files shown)


src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -45,7 +70,9 @@ namespace Rv
         FILE* outFile = std::fopen(fullPath.c_str(), "wb");
         if (outFile == nullptr)
         {
-            cerr << "ERROR: Unable to open output file" << endl;
+            cerr << "ERROR: Unable to open output file: " << fullPath << endl;
+            cerr << "       Directory: " << parentDir.absolutePath().toStdString() << endl;
```
eloisebrosseau, line 74: FYI: The CI is failing because `parentDir` isn't defined:
`/Users/runner/work/OpenRV/OpenRV/src/lib/app/RvPackage/PackageManager.cpp:74:45: error: use of undeclared identifier 'parentDir'
   74 |             cerr << "       Directory: " << parentDir.absolutePath().toStdString() << endl;`

bernie-laberge (approved):
LGTM

PR #1027, opened 2025-12-17
SG-41208: Add softDeleted paint node property to support undo, redo and clear actions in Live Review

eloisebrosseau:
### Summarize your change.

A `softDeleted` property was added for each paint node stroke to support undo, redo and clear actions for OTIO-based Live Review. This new property is following the same pattern as the new `uuid` property. These aren't known by the C++ code responsible for rendering annotations, they are simply used to help with keeping the Annotate and the Live Review package synchronized in terms of the state of each annotation. For each undo, redo and clear action, the Annotate Mode will send an internal event with the least of stroke uuids affected by the action. The `name` field in the OTIO Paint schema was completely removed since it was decided it was actually needed since it isn't an OTIO effect.

The `softDeleted` property was already added as a field in the OTIO Paint schema but it was also added as an RV node property to facilitate the interaction between the Live Review plugin that needs this information and the Annotate mode that is making the action.

Drawing, undoing, redoing and clearing strokes was tested on macOS 15.6.1.

Note that this PR branch is based for the moment on #1021 so the changes to review are only the ones from Dec 16th.

### DIFF
--- a/src/plugins/rv-packages/annotate/annotate_mode.mu
+++ b/src/plugins/rv-packages/annotate/annotate_mode.mu
@@ -2260,13 +2260,6 @@ class: AnnotateMinorMode : MinorMode
         return if _syncAutoStart then CheckedMenuState else UncheckedMenuState;
     }
 
-    method: undoSlot (void; bool checked)
-    {
-        undoPaint();
-        updateFrameDependentState();
-        redraw();
-    }
-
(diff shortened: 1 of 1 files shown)


cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
LGTM

PR #1030, opened 2025-12-18
Move out python packages installations from make_python and make_pyside6

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.

Changes:

- Refactored Python build system to separate concerns and improve Windows Debug build compatibility
- Removed pip install commands from make_python.py and moved dependency installation to CMake
- Created standalone test_python.py script for Python distribution validation
- Numpy version is a CMake variable instead of an ENV variable
- Fixed sitecustomize.py to handle missing dependencies during build and improved error handling

I changed pip install strategy from --no-binary :all: --only-binary pip,setuptools,wheel to --no-binary opentimelineio,numpy,PyOpenGL-accelerate,cryptography,pydantic,cffi.

**Note:** cffi was added because of Windows and the debug build.

**Previously:** 
Forced all packages (except pip/setuptools/wheel) to build from source, including build tools (Cython, meson-python, ninja), pure Python packages (certifi, six, packaging, requests), and packages with native extensions

**Now:** 
Only packages with C/C++/Rust extensions that need to link against our custom Python build are compiled from source. All other packages (build tools, pure Python libraries) use pre-built binary wheels

**Rationale:** 
Build tools like Cython don't need to match our Python's ABI. They're just tools used during compilation. Pre-built wheels work fine and install much faster. Only runtime dependencies with native code that links to Python (like numpy's C extensions or cryptography's Rust/C code) need to be built against our specific Python build

It was not clear why some package were installed in the `requirements.txt`, or in `make_python.py` or at both places.
I think it will be more robust like this and more clear.

MacOS + CI

### DIFF
--- a/cmake/defaults/CY2023.cmake
+++ b/cmake/defaults/CY2023.cmake
@@ -61,11 +61,11 @@ IF(RV_VFX_PLATFORM STREQUAL "CY2023")
       "3_1"
   )
 
-  # NumPy
-  # NumPy for CY2023 VFX platform is 1.23.x series, but PySide2 requires < 1.23
-  # Using numpy 1.22.4 (last version before 1.23) for PySide2 compatibility
-  SET(RV_DEPS_NUMPY_VERSION "1.22.4")
-    
+  # NumPy NumPy for CY2023 VFX platform is 1.23.x series, but PySide2 requires < 1.23 Using numpy 1.22.4 (last version before 1.23) for PySide2 compatibility
+  SET(RV_DEPS_NUMPY_VERSION
+      "1.22.4"
+  )
+
   # OCIO https://github.com/AcademySoftwareFoundation/OpenColorIO
   SET(RV_DEPS_OCIO_VERSION
       "2.2.1"
--- a/cmake/defaults/CY2024.cmake
+++ b/cmake/defaults/CY2024.cmake
@@ -1,62 +1,115 @@
 # CY2024 VFX Reference Platform versions of dependencies see https://vfxplatform.com/
 
 IF(RV_VFX_PLATFORM STREQUAL "CY2024")
-    # Year
-    SET(RV_VFX_CY_YEAR "2024")
-    SET(RV_VFX_CY2024 ON )
-    ADD_COMPILE_DEFINITIONS(QT65ON)
+  # Year
+  SET(RV_VFX_CY_YEAR
+      "2024"
+  )
+  SET(RV_VFX_CY2024
+      ON
+  )
+  ADD_COMPILE_DEFINITIONS(QT65ON)
 
-    # Boost
-    SET(RV_DEPS_BOOST_VERSION "1.82.0")
-    SET(RV_DEPS_BOOST_MAJOR_MINOR_VERSION "1_82")
-    SET(RV_DEPS_BOOST_DOWNLOAD_HASH "f7050f554a65f6a42ece221eaeec1660")
+  # Boost
+  SET(RV_DEPS_BOOST_VERSION
+      "1.82.0"
+  )
+  SET(RV_DEPS_BOOST_MAJOR_MINOR_VERSION
+      "1_82"
+  )
+  SET(RV_DEPS_BOOST_DOWNLOAD_HASH
+      "f7050f554a65f6a42ece221eaeec1660"
+  )
 
-    # Imath
-    # Can find the build version in OpenRV/_build/RV_DEPS_IMATH/install/lib/
-    SET(RV_DEPS_IMATH_VERSION "3.1.12")
-    SET(RV_DEPS_IMATH_DOWNLOAD_HASH "d4059140972da68a2b5a1287ebe5a653")
-    SET(RV_DEPS_IMATH_LIB_VER "29.11.0")
-    SET(RV_DEPS_IMATH_LIB_MAJOR "3_1")
+  # Imath Can find the build version in OpenRV/_build/RV_DEPS_IMATH/install/lib/
+  SET(RV_DEPS_IMATH_VERSION
+      "3.1.12"
+  )
+  SET(RV_DEPS_IMATH_DOWNLOAD_HASH
+      "d4059140972da68a2b5a1287ebe5a653"
+  )
+  SET(RV_DEPS_IMATH_LIB_VER
+      "29.11.0"
+  )
+  SET(RV_DEPS_IMATH_LIB_MAJOR
+      "3_1"
+  )
 
-    # NumPy
-    # https://numpy.org/doc/stable/release.html
-    SET(RV_DEPS_NUMPY_VERSION "1.24.4")
+  # NumPy https://numpy.org/doc/stable/release.html
+  SET(RV_DEPS_NUMPY_VERSION
+      "1.24.4"
+  )
 
-    # OCIO
-    # https://github.com/AcademySoftwareFoundation/OpenColorIO
-    SET(RV_DEPS_OCIO_VERSION "2.3.2")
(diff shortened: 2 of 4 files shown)


bernie-laberge (approved):
LGTM

cmake/dependencies/python3.cmake
```diff
@@ -268,8 +272,16 @@ ELSE()
...
 ENDIF()
 
-# Using --no-binary :all: to ensure all packages with native extensions are built from source
-# against our custom Python build, preventing ABI compatibility issues.
+# Force source builds only for packages with C/C++/Rust extensions that need to link against our custom Python.
+# All other packages (including build tools like Cython) can use pre-built wheels, which avoids
+# path length issues on Windows and improves build speed.
+# Packages built from source:
+#   - opentimelineio: C++ extensions, uses CMAKE_ARGS for proper linking
+#   - numpy: C extensions that need our Python ABI
```
rogernelson: This is a nice find, I did not realize that we were installing more packages in pyside6 when I did this.

cmake/dependencies/python3.cmake
```diff
@@ -282,7 +294,7 @@ ENDIF()
 
 LIST(APPEND _requirements_install_command
     "CMAKE_ARGS=-DPYTHON_LIBRARY=${_python3_cmake_library} -DPYTHON_INCLUDE_DIR=${_include_dir} -DPYTHON_EXECUTABLE=${_python3_executable}"
-    "${_python3_executable}" -s -E -I -m pip install --upgrade --no-cache-dir --force-reinstall --no-binary :all: -r "${_requirements_output_file}"
+    "${_python3_executable}" -s -E -I -m pip install --upgrade --no-cache-dir --force-reinstall --no-binary opentimelineio,numpy,PyOpenGL-accelerate,cryptography,pydantic,cffi -r "${_requirements_output_file}"
```
rogernelson: My main concern around doing that would be if we add a new dependency that has a binary component it can be easy to miss adding it into this list.  

What if we instead specified the packages that we do want to use prebuilts? Something like this:
```
SET(RV_PYTHON_WHEEL_SAFE
    "pip"           # Package installer (pure Python)
    "setuptools"    # Build system (pure Python)
    "PyOpenGL"      # OpenGL bindings without acceleration (pure Python)
    "certifi"       # SSL certificate bundle (just data files)
    "six"           # Python 2/3 compatibility (pure Python)
    "wheel"         # Wheel format support (pure Python)
    "packaging"     # Version parsing (pure Python)
    "requests"      # HTTP library (pure Python)
    CACHE STRING "Pure Python packages safe to install from wheels"
)

STRING(REPLACE ";" "," _wheel_safe_packages "${RV_PYTHON_WHEEL_SAFE}")

    --no-binary :all:
    --only-binary "${_wheel_safe_packages}" 
```
Then when adding a new package, the onus is on the person to allowlist it if its safe.

cmake/dependencies/python3.cmake
```diff
@@ -373,13 +385,30 @@ SET(${_python3_target}-requirements-flag
...
 ADD_CUSTOM_COMMAND(
-  COMMENT "Installing requirements from ${_requirements_output_file}"
+  COMMENT "Installing requirements from ${_requirements_output_file} pyside and other dependencies"
   OUTPUT ${${_python3_target}-requirements-flag}
   COMMAND ${_requirements_install_command}
   COMMAND cmake -E touch ${${_python3_target}-requirements-flag}
   DEPENDS ${_python3_target} ${_requirements_output_file} ${_requirements_input_file}
 )
 
+# Test the Python distribution after requirements are installed
```
rogernelson, line 427: Nice!

src/build/make_pyside6.py
```diff
@@ -159,18 +159,7 @@ def get_fallback_clang_filename_suffix(version):
...
-        raise ValueError("RV_DEPS_NUMPY_VERSION environment variable is not set.")
-    install_numpy_args = get_python_interpreter_args(PYTHON_OUTPUT_DIR, VARIANT) + [
-        "-m",
-        "pip",
-        "install",
-        f"numpy=={numpy_version}",
-    ]
-    print(f"Installing numpy with {install_numpy_args}")
-    subprocess.run(install_numpy_args).check_returncode()
+    # Note: numpy is now installed via requirements.txt in python3.cmake before PySide6 builds.
```
rogernelson, line 162: Great cleanup.

src/build/make_python.py
```diff
@@ -221,27 +243,6 @@ def patch_python_distribution(python_home: str) -> None:
     print(f"Ensuring pip with {ensure_pip_args}")
     subprocess.run(ensure_pip_args).check_returncode()
 
-    pip_args = python_interpreter_args + ["-m", "pip"]
```
rogernelson, line 224: So much ugliness, another nice cleanup.

rogernelson (approved):
Great cleanup, I really only have one somewhat minor suggestion around the binary component package building,

cmake/dependencies/python3.cmake
cedrik-fuoco-adsk: I agree, I'll try that

src/build/make_python.py
```diff
@@ -52,6 +50,7 @@
 all the certificate authorities that are distributed with Firefox.
 """
 import site
+import sys
```
eloisebrosseau: I know you didn't create this but is there any way this could be actual code instead of a big string? It's not possible to lint any of the code declared here

src/build/make_python.py
```diff
@@ -68,10 +67,19 @@
...
-    print("Set DO_NOT_SET_SSL_CERT_FILE to skip this step in RV's Python initialization.", file=sys.stderr)
+    # Only print errors for unexpected exceptions, and only if verbose mode is enabled
+    try:
+        import os as _os
+        if "PYTHONVERBOSE" in _os.environ:
+            print("Failed to set certifi.where() as SSL_CERT_FILE.", file=sys.stderr)
+            print(e, file=sys.stderr)
+            print("Set DO_NOT_SET_SSL_CERT_FILE to skip this step in RV's Python initialization.", file=sys.stderr)
+    except:
+        pass
```
eloisebrosseau: Silencing exceptions with a `try...except...pass` pattern is a bad practice

src/build/test_python.py
```diff
@@ -0,0 +1,166 @@
...
+        subprocess.run(wheel_install_arg).check_returncode()
+
+        python_validation_args = python_interpreter_args + [
+            "-c",
+            "\n".join(
+                [
+                    # Check for tkinter
+                    "try:",
+                    "    import tkinter",
+                    "except:",
```
eloisebrosseau: `try...except` is a bad practice, it should be `try...except Exception`

src/build/make_python.py
cedrik-fuoco-adsk: do you have a suggestion? @eloisebrosseau

eloisebrosseau: I'm not exactly sure what is the workflow you want here so it's hard for me to give a proper suggestion but I believe this should simply be in `except Exception as e:` without being inside another try except block. I'm not exactly sure I see what could be thrown in your print statements

src/build/make_python.py
cedrik-fuoco-adsk: refactored!

src/build/make_python.py
cedrik-fuoco-adsk: refactored!

src/build/test_python.py
cedrik-fuoco-adsk: refactored!

PR #1032, opened 2025-12-18
Pin python dependencies

rogernelson:
### Summarize your change.

This branch pins all the dependencies used by our Python build.  This is to ensure we don't have unexpected behaviour by one of our dependencies adding a new version.  This is especially important now that we are building all of them.  We don't want an upgraded dependency adding a new binary component that we are unable to build with our current toolchain. 

The dependencies are all pinned to the version we currently use.

I also removed the custom build of PyOpenGL_accelerate since the fix has been made in the official repo, we can just use a version containing the fix and remove the patch.

Build consistency

Mac OS 26.1


.github/workflows/ci.yml
```diff
@@ -353,6 +353,21 @@ jobs:
         run: |
           python3 -m pip install --user --upgrade -r requirements.txt
 
+      - name: Install Cython for building PyOpenGL_accelerate
+        run: |
+          python3 -m pip install --user Cython
```
cedrik-fuoco-adsk: Is this still needed with my latest changes? I don't think it is? @rogernelson

rogernelson: You're right, it's probably not anymore. I put it there to fix the initial build error on Linux (which it didn't do anyway).

cedrik-fuoco-adsk (approved):

PR #1038, opened 2025-12-22
Add pre-commit.ci to validate lint and format steps are all passing for CMake, C++ and Python files

eloisebrosseau:
### Summarize your change.

A lot of older and newer CMake files and some lines in a few C++ files were not formatted properly. `pre-commit.ci` was added as an app in `Settings > GitHub Apps` to enable running our pre-commit hooks with GitHub Actions. I removed the Lint and Format step form ci.yml since it is not required anymore with `pre-commit.ci`

In order to add a validation step that all files in this project are formatted properly, I need to format all the files to make sure we stop merging unformatted code going forward.

### DIFF
--- a/.pre-commit-config.yaml
+++ b/.pre-commit-config.yaml
@@ -17,3 +17,6 @@ repos:
       - id: clang-format
         types_or: [c++, c]
         exclude: '\.(m|mm)$'
+
+ci:
+  autofix_prs: false


cedrik-fuoco-adsk:
The pre-commit hook have issue with EasyCLA. Its commit are not signed. @eloisebrosseau 

I am trying to manually sign the commits from the bot, but It does not seems to work so far.

PR #1046, opened 2026-01-05
SG-41606: Create new softDeleted property if it doesn't exist yet

eloisebrosseau:
### Summarize your change.

Instead of only setting the value of the `softDeleted` property if it exist, we should always make sure the property exist and is updated by the different undo/redo/clear operations.

An issue was found in the Live Review plugin where clear actions would not clear the annotations for the other participants if the softDeleted property was not updated by the instance who made the action. The fix for this issue will be part of another PR in the Commercial RV repo, but it also make sense for the annotation tool to always make sure the `softDeleted` property has a value.


bernie-laberge (approved):
LGTM

PR #1058, opened 2026-01-12
SG-40506: Add space between option and arrow and make cmake detect changes

chenlj1:
### Summarize your change.

Increased padding-right to the menu items such that the right arrow is no longer right next to the menu item text. Changed the cmake file to be able to detect changes to the qss files and update the binaries when change is made.

Previously, the right arrow would be extremely close to the menu item text that was the longest. This fix makes the menu look less cramped. As for the cmake file, we realized that when updating the .qss files, despite rebuilding RV, the changes would not get picked up. After changing the cmake file, when making changes in the .qss files, they get picked up correctly and the new build would properly show the changes.

Tested the cmake file to see if it detects changes to .qss files and include them when rebuilding and tested the extra padding to see if the text is still cramped. Both were tested on macOS

The right arrow is no longer right next to the longest menu item

image

### DIFF
--- a/cmake/macros/rv_quote_file.cmake
+++ b/cmake/macros/rv_quote_file.cmake
@@ -104,52 +104,18 @@ FUNCTION(quote_file OUTPUT_FILENAME)
       "${CMAKE_CURRENT_BINARY_DIR}/${_stem}.cpp"
   )
 
-  GET_FILENAME_COMPONENT(input_file_name ${arg_INPUT_FILENAME} NAME)
-  SET(input_timestamp_file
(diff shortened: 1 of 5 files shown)


chenlj1:
@cedrik-fuoco-adsk Me and Patrick realized that the cmake file was not picking up the changes made to the .qss files when rebuilding despite having modifications made to the .qss files. We made some changes in the cmake file to accommodate this. Can you check if the changes to the cmake file are ok?

eloisebrosseau (approved):

PR #1060, opened 2026-01-13
fix: SG-41698: Add Alt modifier check for mouse wheel events

tjjackson:
### Summarize your change.

Added a check for Alt modifier to QTTranslator::sendMouseWheelEvent()

The Alt key combined with the mouse wheel triggers horizontal scrolling, and will be reported in angleDelta().x() instead of angleDelta().y().  I added a check for the Alt modifier for whether to use angleDelta().x() or angleDelta().y() as the delta for determining if the wheel is moving down or up.

Successfully tested on Rocky Linux 9.5

### DIFF
--- a/src/lib/app/RvCommon/QTTranslator.cpp
+++ b/src/lib/app/RvCommon/QTTranslator.cpp
@@ -664,7 +664,7 @@ namespace Rv
         n += "wheel";
         // use pixelDelta() or angleDelta() instead of delta() (deprecated)
         // Alt key means horizontal scroll on Linux and Windows, whereas it is the Shift key on macOS
-        Qt::KeyboardModifier horizontalModifier = 
+        Qt::KeyboardModifier horizontalModifier =
 #ifdef PLATFORM_DARWIN
             Qt::ShiftModifier;
 #else


src/lib/app/RvCommon/QTTranslator.cpp
```diff
@@ -663,7 +663,9 @@ namespace Rv
 
         n += "wheel";
         // use pixelDelta() or angleDelta() instead of delta() (deprecated)
-        if (event->angleDelta().y() < 0)
+        // Alt key means horizontal scroll
+        int delta = (event->modifiers() & Qt::AltModifier) ? event->angleDelta().x() : event->angleDelta().y();
+        if (delta < 0)
```
bernie-laberge, line 674: When the Alt modifier is used in conjunction with the mouse wheel, should we add "left" and "right" ?
@tjjackson : How are you using the Alt + mouse wheel on your end ? It is to achieve horizontal scrolling right ?
In any case: Thank you for your contribution ! Much appreciated !

bernie-laberge: I think I see it in our own bindings: 
    bind("pointer--alt-control--wheeldown", "playcontrol_category", stepForward100, "Step Forward 100 Frames");
    bind("pointer--alt-control--wheelup", "playcontrol_category", stepBackward100, "Step Backward 100 Frames");

OK so it has to remain up and down since the alt-control should be detected and added here.

bernie-laberge (approved):
Thank you @tjjackson !

src/lib/app/RvCommon/QTTranslator.cpp
tjjackson: Yeah, this just restores the previous behavior/logic before the move to Qt6 when the code was changed from delta() (deprecated in Qt6) to angleDelta()

bernie-laberge: FYI @tjjackson : Your fix solves this Qt6 regression on both Linux and Windows. This is great !
But it breaks it on macOS.
On macOS, it is the Shift key that produces an horizontal scroll.

This is the code I propose:
```
        // Alt key means horizontal scroll on Linux and Windows, whereas it is the Shift key on macOS
        Qt::KeyboardModifier horizontalModifier = 
#ifdef PLATFORM_DARWIN
            Qt::ShiftModifier;
#else
            Qt::AltModifier;
#endif
        int delta = (event->modifiers() & horizontalModifier) ? event->angleDelta().x() : event->angleDelta().y();
```

If you agree, it would be great if you could push the modification on your end so that the commit remains in your name. Thanks !

tjjackson: Sure, I can take care of that

tjjackson: Pushed

bernie-laberge: Thanks !

eloisebrosseau:
@tjjackson I think you forgot to format your code. You can use `pre-commit` like mentioned in the [documentation](https://aswf-openrv.readthedocs.io/en/latest/build_system/config_common_build.html#contributing-to-open-rv). To manually trigger it, you can run `pre-commit run clang-format --all-files` and commit the fix

bernie-laberge:
Sorry @tjjackson, the formatting error was my bad: the formatter hadn't ran on the code I proposed to you because I never committed it.
After installing the pre-commit:
`pre-commit install`
The `pre-commit run clang-format --all-files` should format the it correctly : it should remove the extra space after the `horizontalModifier =`.
Thanks !

tjjackson:
Yeah I copied-pasted it and didn't think to run pre-commit again, should be good now

src/lib/app/RvCommon/QTTranslator.cpp
```diff
@@ -663,7 +663,15 @@ namespace Rv
...
         // use pixelDelta() or angleDelta() instead of delta() (deprecated)
-        if (event->angleDelta().y() < 0)
+        // Alt key means horizontal scroll on Linux and Windows, whereas it is the Shift key on macOS
+        Qt::KeyboardModifier horizontalModifier =
+#ifdef PLATFORM_DARWIN
+            Qt::ShiftModifier;
+#else
+            Qt::AltModifier;
+#endif
+        int delta = (event->modifiers() & horizontalModifier) ? event->angleDelta().x() : event->angleDelta().y();
```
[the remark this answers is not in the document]
tjjackson: The purpose of this code change was to address the horizontal modifier, while otherwise maintaining existing logic.  I think changing from angleDelta to pixelDelta is outside the scope of this PR.

PR #1064, opened 2026-01-15
Fix building with multiple XCodes installed

rogernelson:
### Summarize your change.

When building the python packages on Mac, `pip` will use the latest SDK version available be default.  So if you compile RV with XCode 16 (as is required) and have a later SDK installed that will be used to build the python packages potentially making them incompatible with python.

To fix this, we'll pass the SDKROOT from the CMake environment to `pip` to ensure if uses the same one as OpenRV.

Fix building with multiple SDKs installed.

Mac OS 26.1 with XCode 26 and 16 installed.


cedrik-fuoco-adsk (commented):
Good catch! Thank you!

bernie-laberge (approved):
LGTM

PR #1067, opened 2026-01-16
SG-41742: Fix Annotation hook

eloisebrosseau:
### Summarize your change.

Set the uuid value as a list in the paint node property to avoid each char being added separately. Otherwise, it was stored as ["e", "5", "2", "7", ...] instead of ["e527a621-75b6-4ecf-8204-bef59fe4c6c5"]. Note that this issue made me realize the `brush` property was parsed the same way even if it wasn't causing any issues visually. The `softDeleted` property is not required to be stored here, but I added it anyway for consistency.

The wrong format of the uuid was preventing the PAINT_BATCH_UPDATE event in the Live Review plugin from being handled properly since it wasn't matching the one parsed by the Annotation hook when receiving the OTIO timeline when joining a session.

Undoing and redoing in a session where annotations were already present was tested on macOS.

### DIFF
--- a/src/plugins/rv-packages/otio_reader/annotation_hook.py
+++ b/src/plugins/rv-packages/otio_reader/annotation_hook.py
@@ -122,7 +122,7 @@ def hook_function(in_timeline: otio.schemadef.Annotation.Annotation, argument_ma
             pen_component,
             {
                 "color": list(map(float, layer.rgba)),
-                "brush": layer.brush,
+                "brush": [layer.brush],
                 "debug": 1,
                 "join": 3,
                 "cap": 1,
@@ -130,7 +130,8 @@ def hook_function(in_timeline: otio.schemadef.Annotation.Annotation, argument_ma
                 "mode": 0 if layer.type == "COLOR" else 1,
                 "startFrame": start_time,
                 "duration": duration,
-                "uuid": layer.id,
+                "softDeleted": layer.soft_deleted,
+                "uuid": [layer.id],
             },
         )
 


src/plugins/rv-packages/otio_reader/annotation_hook.py
```diff
@@ -122,15 +122,16 @@ def hook_function(in_timeline: otio.schemadef.Annotation.Annotation, argument_ma
             pen_component,
             {
                 "color": list(map(float, layer.rgba)),
-                "brush": layer.brush,
+                "brush": [layer.brush],
```
bernie-laberge, line 125: Sorry I am not sure I understand this fix.
I tested this without your fix:
1. Start a live Review session on the web
2. Join with RV
3. Draw an annotation on the web
4. I saved the RV session to file.

And I get this:
```
    "pen:1:371:annotation"
    {
        string uuid = "f434091c-f255-4124-a12f-94a38712a15f"
        string brush = "CIRCLE"

```
Not "C', "I", "R", etc...

eloisebrosseau: The annotation hook is triggered when joining a session where annotations are already being displayed. In that case, the RV session file shows this:
```
"pen:1:469:annotation"
{
    string brush = [ "C" "I" "R" "C" "L" "E" ]
    string uuid = [ "9" "1" "3" "3" "5" "c" "b" "3" "-" "9" "3" "d" "f" "-" "4" "c" "c" "5" "-" "9" "a" "1" "c" "-" "9" "7" "0" "1" "4" "7" "f" "9" "b" "b" "1" "0" ]
}
```

Instead of this:
```
"pen:1:469:annotation"
{
    string brush = "CIRCLE"
    string uuid = "91335cb3-93df-4cc5-9a1c-970147f9bb10"
}
```

bernie-laberge: Yes you are right ! I can reproduce the issue with existing CR annotations !

bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #1068, opened 2026-01-19
fix: SG-42144: Fix off-by-one frame count error for audio-only files

skyhhh666:
# PR Description: Fix off-by-one frame count error for audio-only files

## Summary
This PR fixes a bug where pure audio files (like `.wav`) could be misidentified as having one frame less than their actual duration. This was caused by floating-point truncation when converting the audio duration (in seconds) to a frame count.

## The Issue
When OpenRV calculates the frame count for an audio file, it multiplies the duration reported by FFmpeg by the target FPS. 
For example:
- An audio file with 28,000 samples at 48kHz is exactly 14 frames at 24fps.
- Because of microsecond precision in FFmpeg's duration reporting, the `timeDuration` might be `0.583333`.
- `0.583333 * 24 = 13.999992`.
- Assigning this to an integer `frames` variable resulted in `13` due to truncation.

## The Fix
Changed the calculation to use `std::round()` on the result before assigning it to the integer `frames` variable. This ensures that tiny precision errors don't cause the loss of the final frame.

## Affected File
- `src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp`

## Verification Results
- Manual calculation: `std::round(13.999992)` now correctly yields `14`.
- This matches the behavior of other industry tools like Maya and Nuke.

### DIFF
--- a/src/bin/apps/rvpkg/main.cpp
+++ b/src/bin/apps/rvpkg/main.cpp
@@ -59,6 +59,7 @@
 
 #include <QtCore/QtCore>
 #include <QFileInfo>
+#include <QDateTime>
 
 //
 //  Check that a version actually exists. If we're compiling opt error
@@ -224,6 +225,7 @@ int utf8Main(int argc, char* argv[])
 
     int install = 0;
     int uninstall = 0;
+    int update = 0;
     int list = 0;
     int info = 0;
     int add = 0;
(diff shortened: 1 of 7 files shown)


skyhhh666:
Hi @bernie-laberge  @eloisebrosseau @cedrik-fuoco-adsk , just a gentle ping. Could you please take a look when you have a chance? Thanks!

bernie-laberge (approved):
Thanks for the fix @skyhhh666 !

PR #1076, opened 2026-01-22
SG-41782 - Possible to use Zoom and Pan operations as Participants when being in Live Review Presentation mode

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Move the zoom and pan shortcuts to a new category

The live review plugin needs granular controls over those specific shortcuts.

MacOS


bernie-laberge (approved):
LGTM

eloisebrosseau (commented):
Shouldn't panzoom_category also be added to _Session.h_?

cedrik-fuoco-adsk:
> Shouldn't panzoom_category also be added to _Session.h_?

Good point. The categories in `Session.h` are only for buttons or other elements from the Toolbars, but I think all categories are there already. I can add it.

cedrik-fuoco-adsk:
> Shouldn't panzoom_category also be added to _Session.h_?

Fixed @eloisebrosseau !

PR #1077, opened 2026-01-22
SG-41848 - Random use-after-free crash

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.
Fixed a memory safety bug in the event category filtering system that could cause crashes when categories were disabled from Mu scripts.

Changed the storage type from std::vector<std::string_view> to std::vector<std::string>. 

This ensures the Session object owns copies of the category strings, making the data persist regardless of what happens to the original Mu string objects.

The `m_disabledEventCategories` member in IPCore::Session was stored as std::vector<std::string_view>. 

Since `std::string_view` is a non-owning reference, when categories were disabled from Mu scripts via `commands.disableEventCategory()`, the stored `string_view` pointed to memory owned by the Mu garbage collector.

**Assumptions based on the call stack of the crash:**

When the Mu GC freed the original string, the `string_view` became a dangling pointer. Subsequent calls to **isEventCategoryEnabled()** would then crash when comparing against the freed memory (manifesting as `EXC_BAD_ACCESS (SIGBUS) KERN_PROTECTION_FAILURE)`.

MacOS

### DIFF
--- a/src/lib/ip/IPCore/Session.cpp
+++ b/src/lib/ip/IPCore/Session.cpp
@@ -707,7 +707,7 @@ namespace IPCore
         // Don't add duplicates
         if (std::find(m_disabledEventCategories.begin(), m_disabledEventCategories.end(), category) == m_disabledEventCategories.end())
         {
-            m_disabledEventCategories.emplace_back(category);
+            m_disabledEventCategories.push_back(std::string(category));
         }
     }
 


bernie-laberge (approved):
Wow !  Super catch @cedrik-fuoco-adsk !
Thanks!

eloisebrosseau (approved):

pbergeron-adsk (commented):
why emplace_back vs push_back?
https://quuxplusone.github.io/blog/2021/03/03/push-back-emplace-back/

cedrik-fuoco-adsk:
> why emplace_back vs push_back? https://quuxplusone.github.io/blog/2021/03/03/push-back-emplace-back/

I used `emplace_back` to construct the string in-place, but I realize now it doesn't help since we're copying from a string_view anyway. I'll change it.

cedrik-fuoco-adsk:
@pbergeron-adsk I changed it back to push_back, but I needed to use explicit conversion because push_back is marked `explicit` in C++17.

PR #1080, opened 2026-01-24
fix: SG-43090: Fix SYNDisplay property regression

### LINKED ISSUE #1034: [Bug]: SYNDisplay outTransform.url property regression
### What happened?

Switching the color pipeline to use a `SYNDisplay` node used to require an outTransform.url property (I last successfully tested in OpenRV 2.1.0 - hash 45e5bae). When `outTransform.url` was set, the display colorspace would change accordingly. When not set, the node would correctly raise an error. However, that property appears to have no effect on the color pipeline in OpenRV 2.2.0 (e.g. latest on branch `RB-2.2.0-VFX2023`).

This code should set the display colorspace using an icc data file. But in the latest RV, the display colorspace has no correction. Note this code assumes you're viewing on the primary monitor, and also that `icc_file` is pointed to a real icc filepath:
```
icc_file = '<path-to>/colorspace.icc'
commands.setStringProperty("displayGroup0_colorPipeline.pipeline.nodes", ["SYNDisplay"])
syn_display_node = commands.nodesOfType("SYNDisplay")[0]
commands.setStringProperty(f"{syn_display_node}.outTransform.url", [icc_file])
```

Also, this code sample should yield the error `ERROR: OCIOIPNode: outTransform.url property needs to be set for syndisplay function` when working correctly, as `outTransform.url` is not set on the syndisplay node. But no error is thrown in the latest RV.
```
icc_file = '<path-to>/colorspace.icc'
commands.setStringProperty("displayGroup0_colorPipeline.pipeline.nodes", ["SYNDisplay"])
```

Linux RHEL 9.4, 9.6

HP Z8 G4 workstation, nVidia NVIDIA RTX A6000, 40 cores (Intel Xeon Gold 6242R 3.1), 191 GB RAM 

```shell

```

### Environment variables

_No response_

### Extra information

_No response_

### Code of Conduct

williamwira:
### Linked issues
Fixes #1034 

### Summarize your change.
Updated initialization step in OCIOIPNode used by SYNDisplay, so outTransform.url would be checked.

Regression caused SYNDisplay outTransform.url property to be ignored, so we could not set the display colorspace with an icc filepath.

Confirmed this update worked on Linux RHEL 9.6. Executed this python code on a single-monitor system to validate the change. The code switches the display group's color pipeline node to a `SYNDisplay` node, and then assigns an icc profile file to the node's `outTransform.url` property.

```
icc_file = '<path-to>/colorspace.icc'
commands.setStringProperty("displayGroup0_colorPipeline.pipeline.nodes", ["SYNDisplay"])
syn_display_node = commands.nodesOfType("SYNDisplay")[0]
commands.setStringProperty(f"{syn_display_node}.outTransform.url", [icc_file])
```

### DIFF
--- a/src/lib/ip/OCIONodes/OCIOIPNode.cpp
+++ b/src/lib/ip/OCIONodes/OCIOIPNode.cpp
@@ -162,7 +162,7 @@ namespace IPCore
             }
         }
 
-        updateConfig();
+        updateConfig(true);
     }
 
     OCIOIPNode::~OCIOIPNode()
@@ -174,7 +174,7 @@ namespace IPCore
         delete m_state;
     }
 
-    void OCIOIPNode::updateConfig()
+    void OCIOIPNode::updateConfig(const bool initializing)
     {
         try
         {
@@ -230,7 +230,10 @@ namespace IPCore
         m_state->shaderID = "";
 
         updateContext();
-        updateFunction();
+        if (!initializing)
+        {
+            updateFunction();
+        }
     }
 
     void OCIOIPNode::updateContext()
@@ -352,7 +355,7 @@ namespace IPCore
         boost::hash<string> string_hash;
         string inName = stringProp("ocio.inColorSpace", m_state->linear);
 
-        if (inName.empty())
+        if (inName.empty() && !useRawConfig())
             return;
 
         try
--- a/src/lib/ip/OCIONodes/OCIONodes/OCIOIPNode.h
+++ b/src/lib/ip/OCIONodes/OCIONodes/OCIOIPNode.h
@@ -52,7 +52,7 @@ namespace IPCore
         virtual void readCompleted(const std::string&, unsigned int);
         virtual void propertyChanged(const Property*);
 
-        void updateConfig();
+        void updateConfig(const bool initializing = false);
 
         bool useRawConfig() const { return m_useRawConfig; }
 


bernie-laberge:
Hello @williamwira, I apologize for the long delay in reviewing this PR ! We are trying to figure out a way to speed things up here with our robots friends. 
Thank you for the fix. It solves the regression as I was able to validate on my end.
Now I would like to propose some minor changes for this fix. 
I have put what I have in mind here:
https://github.com/AcademySoftwareFoundation/OpenRV/pull/1239/changes

I will also list them in your PR. They will achieve the same end result. It only changes the form.
Let me know what you think.

src/lib/ip/OCIONodes/OCIONodes/OCIOIPNode.h
```diff
@@ -52,7 +52,7 @@ namespace IPCore
         virtual void readCompleted(const std::string&, unsigned int);
         virtual void propertyChanged(const Property*);
 
-        void updateConfig();
+        void updateConfig(const bool initializing = false);
```
bernie-laberge, line 55: By adding the following data member to the class, we could leave this void updateConfig() method unchanged:
`bool m_initialized{false};`
This way we could also leave updateConfig() call unchanged below:

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -230,7 +230,10 @@ namespace IPCore
         m_state->shaderID = "";
 
         updateContext();
-        updateFunction();
+        if (!initializing)
+        {
+            updateFunction();
+        }
```
bernie-laberge, line 236: We could replace this with:
```
        updateContext();
        updateFunction();

        m_initialized = true;
```

And in the ::updateFunction. we would skip the error when the class hasn't completed its initialization:

```
                if (inTransformURL.empty() && (!m_inTransformData || m_inTransformData->size() == 0))
                {
                    if (!m_initialized) return;
                    TWK_THROW_EXC_STREAM("Either inTransform.url or inTransform.data property "
                                         "needs to be set for synlinearize function");
                }
...
                if (outTransformURL.empty())
                {
                    if (!m_initialized) return;
                    TWK_THROW_EXC_STREAM("outTransform.url property needs to "
                                         "be set for syndisplay function");
                }
```

src/lib/ip/OCIONodes/OCIOIPNode.cpp
```diff
@@ -352,7 +355,7 @@ namespace IPCore
         boost::hash<string> string_hash;
         string inName = stringProp("ocio.inColorSpace", m_state->linear);
 
-        if (inName.empty())
+        if (inName.empty() && !useRawConfig())
             return;
```
bernie-laberge, line 359: Here, to make the code easier to understand, perhaps we could add the needsInColorSpace vaviable which would make it equivalent to your original code:

```
        // synlinearize/syndisplay build their pipeline from file/URL
        // transforms, so an empty input color space is valid for those modes.
        const bool needsInColorSpace =
            ociofunction != "synlinearize" && ociofunction != "syndisplay";
        if (inName.empty() && needsInColorSpace)
            return;
```

PR #1085, opened 2026-01-26
SG-41892: Prevent play on click in lr collab mode

pbergeron-adsk:
### Summarize your change.

When in Live Review collab mode, where everyone has at least some degree of control over what goes on in the view, we prevent non-presenter participants starting "play" if someone clicks in the main view. Clicking in the view is too easy to do and makes everyone lose context.

macOS

Test for a specific command category to override the play-on-click preference.

n/a

### DIFF
--- a/src/plugins/rv-packages/otio_reader/annotation_hook.py
+++ b/src/plugins/rv-packages/otio_reader/annotation_hook.py
@@ -37,7 +37,7 @@ def _get_transform_properties():
                 global_translate_vec = otio.schema.V2d(global_translate[0], global_translate[1])
                 global_scale_vec = otio.schema.V2d(global_scale[0], global_scale[1])
             else:
-                logging.warning(
+                logging.debug(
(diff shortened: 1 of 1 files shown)


src/lib/app/mu_rvui/rvui.mu
```diff
@@ -7200,7 +7200,10 @@ global bool debugGC = false;
 \: clickTimerTimeout (void; )
 {
     State state = data();
-    if (! state.clickToPlayDisabled) togglePlayIfNoScrub();
+    if (!state.clickToPlayDisabled && commands.isEventCategoryEnabled("playcontrol_click_category"))
```
bernie-laberge, line 7203: I think @cedrik-fuoco-adsk was keeping a list in Session.h so I think we should add it here also:
https://github.com/AcademySoftwareFoundation/OpenRV/blob/8807976e0baca9e02bcf592c3edcf02ae54cd0bc/src/lib/ip/IPCore/IPCore/Session.h#L65
Apart from that it looks good thanks !

PR #1087, opened 2026-01-27
feat: SG-42131 - Migrate to OpenImageIO 3.x and build it with OpenColorIO

### LINKED ISSUE #893: Update OIIO to version 3.x

mcoliver:
OIIO development has moved to v3.  Also adding ocio support with oiio.

### Linked issues

fixes #893 
### Summarize your change.
Upgrade OIIO to v3 and build with OCIO

OIIO v2 is considered obsolete: https://github.com/AcademySoftwareFoundation/OpenImageIO/releases/tag/v2.5.19.1

```Note that this is a patch release for the obsolete 2.5 branch that merely repairs its ability to build under new versions of some dependencies. You need not upgrade if you are not encountering build problems. Except for legacy constraints, everybody should currently be using OIIO 3.0 and testing the beta for 3.1 which will become the new supported branch very soon.```

built locally and tested on macos for 2024 and 2025. Need the Academy's CICD to build for other configs

It appears OIIO implementation is pretty minimal so I think we should be good but testing on other distros to validate would be appreciated.  

### DIFF
--- a/cmake/dependencies/ocio.cmake
+++ b/cmake/dependencies/ocio.cmake
@@ -145,6 +145,9 @@ LIST(APPEND _configure_options "-DZLIB_ROOT=${RV_DEPS_ZLIB_ROOT_DIR}")
 
 # OCIO apps are not needed.
 LIST(APPEND _configure_options "-DOCIO_BUILD_APPS=OFF")
+# Use MISSING so OCIO vendors its own third-party deps when not already provided by RV_DEPS (vs NONE/default), pinning versions and avoiding ABI/licensing drift
+# with system libs; typical CI pulls in Imath/OpenEXR components, yaml-cpp, pystring, expat, lcms2, zlib, and related OCIO external packages.
+LIST(APPEND _configure_options "-DOCIO_INSTALL_EXT_PACKAGES=MISSING")
 
 IF(NOT RV_TARGET_WINDOWS)
   EXTERNALPROJECT_ADD(
--- a/cmake/dependencies/oiio.cmake
+++ b/cmake/dependencies/oiio.cmake
@@ -48,7 +48,8 @@ LIST(APPEND _byproducts ${_byprojects_copy})
 # The '_configure_options' list gets reset and initialized in 'RV_CREATE_STANDARD_DEPS_VARIABLES'
 LIST(APPEND _configure_options "-DBUILD_TESTING=OFF")
 LIST(APPEND _configure_options "-DUSE_PYTHON=0") # this on would requireextra pybind11 package
-LIST(APPEND _configure_options "-DUSE_OCIO=0")
+LIST(APPEND _configure_options "-DUSE_OCIO=ON")
+LIST(APPEND _configure_options "-DOpenColorIO_ROOT=${RV_DEPS_OCIO_ROOT_DIR}")
 LIST(APPEND _configure_options "-DUSE_FREETYPE=0")
 LIST(APPEND _configure_options "-DUSE_GIF=OFF")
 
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

File: cmake/dependencies/ocio.cmake [hunk 1]
```diff
@@ -148,6 +148,7 @@ LIST(APPEND _configure_options "-DZLIB_ROOT=${RV_DEPS_ZLIB_ROOT_DIR}")
 
 # OCIO apps are not needed.
 LIST(APPEND _configure_options "-DOCIO_BUILD_APPS=OFF")
+LIST(APPEND _configure_options "-DOCIO_INSTALL_EXT_PACKAGES=MISSING")
```
cedrik-fuoco-adsk, line 150: Why did you add this? Is there a missing dependencies? I don't think it should be needed since OCIO version hasn't changed.

cedrik-fuoco-adsk:
Thank you @mcoliver! I added the `do not merge` label because we won't merge this before mid-february at the very least. We don't want to change dependencies versions before the incoming release.

cmake/dependencies/ocio.cmake
mcoliver: https://opencolorio.readthedocs.io/en/latest/quick_start/installation.html#automated-installation

`MISSING` is the default  so this line is not needed.  I was debugging some things on my local machine (issues specifically with yaml-cpp) and had added the line so I could set it to `NONE` or `ALL`.  Would recommend leaving it for debugging purposes but also understand not adding unnecessary lines.  

I defer to your judgement here on what to do.  Please let me know which direction you would like to go with.

cedrik-fuoco-adsk:
Hi @mcoliver, 

You will need to add the following to oiio.cmake to fix the issue with OpenJPEG:
`LIST(APPEND _configure_options "-DOPENJPEG_VERSION=${RV_DEPS_OPENJPEG_VERSION}")`

cmake/dependencies/ocio.cmake
cedrik-fuoco-adsk: Oh, that's right. I forgot that it was the default value. We can keep it. It will make the OCIO build more transparent.

cedrik-fuoco-adsk:
> Hi @mcoliver,
> 
> You will need to add the following to oiio.cmake to fix the issue with OpenJPEG: `LIST(APPEND _configure_options "-DOPENJPEG_VERSION=${RV_DEPS_OPENJPEG_VERSION}")`

Hi @mcoliver, do you have time to fix the issues in the CI? Depending on the platforms, OIIO can not find OCIO or OPENJPEG_VERSION.

cmake/dependencies/oiio.cmake
```diff
@@ -51,7 +51,8 @@ LIST(APPEND _byproducts ${_byprojects_copy})
 # The '_configure_options' list gets reset and initialized in 'RV_CREATE_STANDARD_DEPS_VARIABLES'
 LIST(APPEND _configure_options "-DBUILD_TESTING=OFF")
 LIST(APPEND _configure_options "-DUSE_PYTHON=0") # this on would requireextra pybind11 package
-LIST(APPEND _configure_options "-DUSE_OCIO=0")
+LIST(APPEND _configure_options "-DUSE_OCIO=ON")
```
cedrik-fuoco-adsk, line 51: Since we are using OCIO with OpenRV, is there a reason to add OCIO to OIIO? Is there a specific feature that you needed? @mcoliver

mcoliver: My understanding was you are using OCIO in RV for display, however OIIO needs OCIO if it is to read/write images in the proper color space.  Perhaps I am wrong here and not understanding how OIIO hooks into RV.  So for example if someone wanted to export images with the color transformation baked in.

cedrik-fuoco-adsk:
Hi @mcoliver,

The download hash is wrong and You will probably need to add the following to oiio.cmake to fix the issue with OpenJPEG: LIST(APPEND _configure_options "-DOPENJPEG_VERSION=${RV_DEPS_OPENJPEG_VERSION}")

cedrik-fuoco-adsk (approved):

eloisebrosseau (approved):

PR #1093, opened 2026-01-28
build: Specify full path to the grep utility in rvcmd.sh

### LINKED ISSUE #1092: [Bug]: build errors grep not working when grep is aliased
### What happened?

It's great that if the build fails it attempts to grep for things like errors.  The issue is that sometime people (like me) have grep aliased to something like ripgrep or silver searcher (ag) which sometimes use flags differently.  For example -E is different in gnu grep and ripgrep.  Referencing the tool with the full pathname to ensure we are using /usr/bin/grep would be more deterministic.

macos

macbook

```shell

```

### Environment variables

_No response_

### Extra information

_No response_

### Code of Conduct

mcoliver:
some people have grep aliased to something else like rg which uses -E differently.  Specifying the full path to the grep bin ensures we are using the right utility.

### Linked issues

Fixes #1092 
### Summarize your change.
Full pathname to grep utility
It's great that if the build fails it attempts to grep for things like errors. The issue is that sometime people (like me) have grep aliased to something like ripgrep or silver searcher (ag) which sometimes use flags differently. For example -E is different in gnu grep and ripgrep. Referencing the tool with the full pathname to ensure we are using /usr/bin/grep would be more deterministic.
macos


mcoliver (commented):
Updated to use command as per co-pilot

bernie-laberge (approved):
LGTM
Thank you for this fix @mcoliver !

cedrik-fuoco-adsk (approved):

PR #1123, opened 2026-02-09
build: Speed up builds with ccahe/scache

rogernelson:
### Summarize your change.

The branch adds ccache (or scache for Windows) to the build system if its installed on the user's machine. If they do not have it installed, it will simply not be used (they will receive a message in the shell though that it is disabled and how they can install it). I also updated the docs with first time setup instructions on how to install the cache. Note that I modified the CICD to install ccache and scache so we can take advantage of it there.

I also added check that the user will be warned when their cache is more than 90% full to either clean it or increase its size or disable the caching.

The goal of this is to speed up builds.  The cache will store object files such that when they have not changed since they were last cached, we do not need to rebuild them.

From my benchmarking on a Mac M4 Pro with OS 26.2 the build time for a clean build went from 25 minutes to 12 when nothing changes.

Build faster

Building on OS 26.2

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -366,18 +366,20 @@ jobs:
 
       # For Rocky Linux, we can't cache the dependencies of Open RV
       # because there is not enough room in the cache (10GB limit) for all platforms.
-      # However, we CAN cache compiled objects with ccache (much smaller cache size).
-
+      # ccache is used for this run only (not stored in Actions cache) to speed up the build.
+      # Let's revisit this if we can reduce the size of the cache (potentially dropping x86_64 for Mac)
+      # or by moving some dependencies from the cache to Conan or another dependency manager.
       - name: Install ccache
         run: |
           dnf install -y ccache
 
-      - name: Setup ccache
-        uses: hendrikmuhs/ccache-action@v1.2
-        with:
-          key: rocky-${{ matrix.rocky-version }}-${{ matrix.vfx-platform }}-${{ matrix.build-type }}
-          max-size: 2G
-          verbose: 2
+      # Uncomment to persist ccache in Actions cache (uses repo 10GB limit).
+      # - name: Setup ccache
+      #   uses: hendrikmuhs/ccache-action@5ebbd400eff9e74630f759d94ddd7b6c26299639
+      #   with:
+      #     key: rocky-${{ matrix.rocky-version }}-${{ matrix.vfx-platform }}-${{ matrix.build-type }}
+      #     max-size: 2G
+      #     verbose: 2
 
       - name: Configure OpenRV
         if: ${{ matrix.vfx-platform == 'CY2023' }}
@@ -617,7 +619,7 @@ jobs:
         # Restore only if there are no changes under cmake folder and Release build.
         if: env.SKIP_DEPS_CACHE == 'false' && steps.changes-in-deps.outputs.dependencies == 'false' && matrix.build-type == 'Release'
         id: cmake-dependencies
-        uses: actions/cache@v4
+        uses: actions/cache/restore@v4
         with:
           path: |
             ./_build/RV_DEPS_*
@@ -625,12 +627,17 @@ jobs:
             ./_build/_deps
           key: macos-${{ matrix.arch-type }}-${{ matrix.vfx-platform }}-dependencies
 
-      - name: Setup ccache
-        uses: hendrikmuhs/ccache-action@v1.2
-        with:
-          key: macos-${{ matrix.os }}-${{ matrix.arch-type }}-${{ matrix.vfx-platform }}-${{ matrix.build-type }}
-          max-size: 2G
-          verbose: 2
+      # Uncomment to persist ccache in Actions cache (uses repo 10GB limit).
+      # - name: Setup ccache
+      #   uses: hendrikmuhs/ccache-action@5ebbd400eff9e74630f759d94ddd7b6c26299639
+      #   with:
+      #     key: macos-${{ matrix.os }}-${{ matrix.arch-type }}-${{ matrix.vfx-platform }}-${{ matrix.build-type }}
+      #     max-size: 2G
+      #     verbose: 2
+
+      # ccache from Homebrew is used for this run only (not stored in Actions cache).
+      # Let's revisit this if we can reduce the size of the cache (potentially dropping x86_64 for Mac)
+      # or by moving some dependencies from the cache to Conan or another dependency manager.
 
       - name: Configure OpenRV
         if: ${{ matrix.vfx-platform == 'CY2023' }}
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -91,6 +91,7 @@ SET(CMAKE_MODULE_PATH
 
 INCLUDE(rv_options) # RV's build options
 INCLUDE(rv_targets) # RV's build platform definitions
+INCLUDE(rv_compiler_cache) # RV's compiler cache (ccache/sccache) support
```
cedrik-fuoco-adsk, line 94: I think you forgot to add rv_compiler_cache file. It is missing from the PR

docs/build_system/config_linux_rocky89.md
```diff
@@ -115,6 +115,29 @@ sudo dnf install -y xz-devel mesa-libGLU mesa-libGLU-devel
...
+#### 1.3. Install ccache for faster builds (recommended)
+
+ccache dramatically speeds up rebuild times by caching compiled objects (50-80% faster rebuilds).
+
+```bash
+sudo dnf install -y ccache
+```
+
+Configure ccache with a larger cache size:
+```bash
```
eloisebrosseau, line 133: I am seeing a few [markdownlint](https://marketplace.visualstudio.com/items?itemName=DavidAnson.vscode-markdownlint) warnings (e.g. `MD031/blanks-around-fences: Fenced code blocks should be surrounded by blank lines`) from the changes you made in this PR, and I think this extension is also available as a vim plugin. I highly recommend using it to make sure all new additions to our documentation are following markdown best practices

CMakeLists.txt
rogernelson: 🤦 You're right, I forgot the `git add`.  It's there now.

docs/build_system/config_linux_rocky89.md
rogernelson: I put a markdown linter in the pre-commit hooks.  I tried the one in your link but it kept hanging for me, so I used `mdformat` which seems to work fine.

eloisebrosseau: I'm not sure `mdformat` is working that fine. I see it reformatted lists with just 1s:
```
1. [Install Pyenv / Python](rocky_install_pyenv_and_python)
1. [Install CMake](rocky_install_cmake)
1. [Install Rust](rocky_install_rust)
1. [Install Qt](rocky_install_qt)
1. [Build Open RV](rocky_build_openrv)
1. [Building with Docker (Optional)](rocky_building_with_docker)
```

What do you mean by the one I suggested kept hanging on you? How did you try to use it?

rogernelson: I tried to add it as a pre-commit hook.  It would just hang trying to format the files.

rogernelson: This turned out to be a much bigger task than anticipated. The markdownlint hook works now (I was trying it when github was being flakey so it might have been that).  But to make the pre-commit check pass I had to run it on all files.  The changes seem benign from what I can tell, but there are a lot of them. I made exceptions for the rules that we were already violating that could not be auto-fixed.

rogernelson: Split it into its own PR [here](https://github.com/AcademySoftwareFoundation/OpenRV/pull/1140) and revert all the linting commits from this PR.

cedrik-fuoco-adsk:
@rogernelson I was looking at `Post Setup Sccache` on the Windows job (https://github.com/AcademySoftwareFoundation/OpenRV/actions/runs/22149988186/job/64038299874?pr=1123) and I don't think it's caching anything.
Screenshot 2026-02-18 at 15 03 33

Also, I am not sure what is the best step forward because we are going over the caches limit.
See the cache page here: https://github.com/AcademySoftwareFoundation/OpenRV/actions/caches

`macos-x86_64-CY2024-dependencies` and `macos-arm64-CY2024-dependencies` are about 5GB each.

.github/workflows/ci.yml
cedrik-fuoco-adsk, line 1: There is an issue with **hendrikmuhs/ccache-action**. I don't see a way to split this into a "restore" step and a "save" step like the GitHub caches action. Right now, I think all PRs will create cache entries which will ended going over the limit pretty fast and the cache entries will be evicted.

If we could split that into a "restore" and "save" step, we could add a condition that prevents PR from executing the "save" step, and just use the cache entries produced by the main branch (if valid).

We could do the following, but it won't work with hendrikmuhs/ccache-action because the same action does the "restore" and the "save".

```
- name: Setup ccache
  if: github.event_name != 'pull_request'
  uses: hendrikmuhs/ccache-action@v1.2
  with:
    key: rocky-${{ matrix.rocky-version }}-${{ matrix.vfx-platform }}-${{ matrix.build-type }}
    max-size: 2G
    verbose: 2
```

cedrik-fuoco-adsk: Forget what I said. There is an option hidden in the action files.

@rogernelson Could you add the save property for all hendrikmuhs/ccache-action@v1.2 instance?
`save: ${{ github.event_name != 'pull_request' }}`

It should look like this:
```
      - name: Setup ccache
        uses: hendrikmuhs/ccache-action@v1.2
        with:
          key: rocky-${{ matrix.rocky-version }}-${{ matrix.vfx-platform }}-${{ matrix.build-type }}
          save: ${{ github.event_name != 'pull_request' }}
          max-size: 2G
          verbose: 2
```

.github/workflows/ci.yml
```diff
@@ -613,6 +625,13 @@ jobs:
             ./_build/_deps
           key: macos-${{ matrix.arch-type }}-${{ matrix.vfx-platform }}-dependencies
 
+      - name: Setup ccache
+        uses: hendrikmuhs/ccache-action@v1.2
+        with:
+          key: macos-${{ matrix.os }}-${{ matrix.arch-type }}-${{ matrix.vfx-platform }}-${{ matrix.build-type }}
+          max-size: 2G
+          verbose: 2
+
```
cedrik-fuoco-adsk, line 641: @rogernelson Do you mind changing the version used to the commit instead for all instances. It is recommended for security to use hash instead of version. (In case someone hijack the repository of the action)

It would look like this:
```
      - name: Setup ccache
        uses: hendrikmuhs/ccache-action@5ebbd400eff9e74630f759d94ddd7b6c26299639
        with:
          key: macos-${{ matrix.os }}-${{ matrix.arch-type }}-${{ matrix.vfx-platform }}-${{ matrix.build-type }}
          max-size: 2G
          verbose: 2
```

rogernelson:
> @rogernelson I was looking at `Post Setup Sccache` on the Windows job (https://github.com/AcademySoftwareFoundation/OpenRV/actions/runs/22149988186/job/64038299874?pr=1123) and I don't think it's caching anything.

You're right.  Windows needs `DCMAKE_CXX_COMPILER_LAUNCHER=sccache` and `DCMAKE_C_COMPILER_LAUNCHER=sccache` before it will work.

.github/workflows/ci.yml
rogernelson: This got me wondering... should we do the same thing with the dependency cache? It seems currently PRs can save the dependency cache right? Do we want that?

cedrik-fuoco-adsk: Yes, I think we should add an if for the "save" step of the dependencies caching. PRs should use the cache from the main branch. I don't recall why I haven't added that from the start.

rogernelson: I'll take care of it, no worries.

PR #1126, opened 2026-02-11
Fix return value in rmsImageDiff main

eloisebrosseau:
### Summarize your change.

Return the status value instead of always returning 0 when doing images comparison with -cmp

0 was always return instead of the status value returned by `computeRMS`.


src/bin/imgtools/rmsImageDiff/main.cpp
```diff
@@ -406,5 +406,5 @@ int utf8Main(int argc, char* argv[])
     TwkMovie::GenericIO::shutdown(); // Shutdown TwkMovie::GenericIO plugins
     TwkFB::GenericIO::shutdown();    // Shutdown TwkFB::GenericIO plugins
 
-    return 0;
```
eloisebrosseau, line 409: @copilot Can you review this change?

PR #1139, opened 2026-02-18
feat: SG-41979: Add hotkey and env var for no sequence formation during drag and drop

chenlj1:
### Summarize your change.

Added an Alt/Option hotkey for drag-and-drop: holding Alt while dropping files prevents sequence formation so each file is added as its own source. Modifier is read in QTTranslator, the Mu layer passes a "drop-no-sequence" tag into addSources, RvSession uses that tag so LoadState and the non-delayed path do not re-group.

Added RV_NO_SEQUENCE_PATTERN that can be set using export RV_NO_SEQUENCE_PATTERN="wanted_pattern". When set to a regex, matching file names are not grouped into sequences.

The previous env substring (RV_NO_SEQUENCE_SUBSTRING) did not meet the client’s needs. They wanted a hotkey so they can have more control over sequence formation.

The client wanted something more powerful than the previous substring approach and suggested using RegEx instead.

The hotkey was tested on macOS by holding down the Option key and verifying that no sequence would form during a drag and drop of files.

https://github.com/user-attachments/assets/edb60953-b6b6-4a65-a687-e2d4357600ef

The RegEx was tested on macOS by setting the environment variable by using export RV_NO_SEQUENCE_PATTERN="_[vV][0-9]+\.jpg$"

https://github.com/user-attachments/assets/c81ecac9-82c0-453d-8931-e068e36d3ed6

### DIFF
--- a/src/lib/app/RvCommon/QTTranslator.cpp
+++ b/src/lib/app/RvCommon/QTTranslator.cpp
@@ -912,13 +912,14 @@ namespace Rv
         }
 #endif
 
+#if defined(RV_VFX_CY2023)
+            Qt::KeyboardModifiers dropEventModifiers = devent->keyboardModifiers();
+#else
+            Qt::KeyboardModifiers dropEventModifiers = devent->modifiers();
+#endif
+
             if (devent->mimeData()->hasUrls())
             {
-#if defined(RV_VFX_CY2023)
-                Qt::KeyboardModifiers dropEventModifiers = devent->keyboardModifiers();
-#else
-                Qt::KeyboardModifiers dropEventModifiers = devent->modifiers();
-#endif
                 QList<QUrl> urls = devent->mimeData()->urls();
 
                 FileNameList files;
@@ -999,10 +1000,10 @@ namespace Rv
             {
 #if defined(RV_VFX_CY2023)
                 DragDropEvent e(ename, m_node, type, DragDropEvent::Text, devent->mimeData()->text().toUtf8().constData(),
-                                modifiers(devent->keyboardModifiers()), devent->pos().x(), h - devent->pos().y() - 1, w, h);
+                                modifiers(dropEventModifiers), devent->pos().x(), h - devent->pos().y() - 1, w, h);
 #else
                 DragDropEvent e(ename, m_node, type, DragDropEvent::Text, devent->mimeData()->text().toUtf8().constData(),
-                                modifiers(devent->modifiers()), devent->position().toPoint().x(), h - devent->position().toPoint().y() - 1,
+                                modifiers(dropEventModifiers), devent->position().toPoint().x(), h - devent->position().toPoint().y() - 1,
                                 w, h);
 #endif
 


src/lib/base/TwkUtil/FrameUtils.cpp
```diff
@@ -285,6 +288,28 @@ struct LexinumericCompare
         if (minSequenceSize == -1)
             minSequenceSize = defaultMinSequenceSize;
 
+        const char* envVar = getenv("RV_NO_SEQUENCE_PATTERN");
```
bernie-laberge, line 292: Note: In RV, we are usually assuming that environment variable are only evaluated once so we can usually optimize by only evaluating the ebv variable once. But I think that in this particular case, it can be especially useful to reevalute the env var each time sequencesInFileList() is called. This way a custom RV package could set this RV_NO_SEQUENCE_PATTERN env var to something else as needed during the life of the RV process. So leave it as is but just a note that usually we only have to evaluate the env var once (like for TWK_MIN_SEQUENCE_SIZE above)

src/lib/base/TwkUtil/FrameUtils.cpp
```diff
@@ -313,6 +338,13 @@ struct LexinumericCompare
                 continue;
             }
 
+            if (noSequencePatternRe && noSequencePatternRe->matches(f))
+            {
+                frameSequences.push_back(f);
+                allfiles.pop_front();
+                continue;
```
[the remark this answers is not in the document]
bernie-laberge: @chenlj1 : This is a valid point ! We should only match the last path component (the same part used for sequence detection) not the whole directory path.

src/lib/base/TwkUtil/FrameUtils.cpp
chenlj1: Ok, I'll keep this in mind. Thanks!

src/lib/base/TwkUtil/FrameUtils.cpp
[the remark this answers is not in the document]
chenlj1: Added const string nameForPattern = filesystem::path(f).filename().string(); so that the pattern matching only happens on the file name instead of the whole path.

src/lib/base/TwkUtil/FrameUtils.cpp
```diff
@@ -285,6 +289,28 @@ struct LexinumericCompare
...
+            {
+                cachedNoSequencePattern = envVar;
+                try
+                {
+                    noSequencePatternRe = make_unique<RegEx>(envVar);
+                }
+                catch (...)
+                {
+                    noSequencePatternRe.reset();
+                }
```
eloisebrosseau, line 306: I believe you should use `catch(const RegEx::Exception& e)` here instead of a catch-all, and you might want to add a warning message with `std::cerr` to let the user know `RV_NO_SEQUENCE_PATTERN` is not a valid pattern instead of silently handling it.

src/lib/app/RvCommon/QTTranslator.cpp
```diff
@@ -914,6 +914,11 @@ namespace Rv
 
             if (devent->mimeData()->hasUrls())
             {
+#if defined(RV_VFX_CY2023)
+                Qt::KeyboardModifiers dropEventModifiers = devent->keyboardModifiers();
+#else
+                Qt::KeyboardModifiers dropEventModifiers = devent->modifiers();
+#endif
```
eloisebrosseau, line 919: Instead of defining dropEventModifiers in side the if statement from line 915, I think it would make more sense to do it before:
```
#if defined(RV_VFX_CY2023)
            Qt::KeyboardModifiers dropEventModifiers = devent->keyboardModifiers();
#else
            Qt::KeyboardModifiers dropEventModifiers = devent->modifiers();
#endif

if (devent->mimeData()->hasUrls())
{
...
}
```

src/lib/app/RvCommon/QTTranslator.cpp
```diff
@@ -989,11 +998,12 @@ namespace Rv
...
-                DragDropEvent e(ename, m_node, type, DragDropEvent::Text, devent->mimeData()->text().toUtf8().constData(), 0,
-                                devent->pos().x(), h - devent->pos().y() - 1, w, h);
+                DragDropEvent e(ename, m_node, type, DragDropEvent::Text, devent->mimeData()->text().toUtf8().constData(),
+                                modifiers(devent->keyboardModifiers()), devent->pos().x(), h - devent->pos().y() - 1, w, h);
 #else
-                DragDropEvent e(ename, m_node, type, DragDropEvent::Text, devent->mimeData()->text().toUtf8().constData(), 0,
-                                devent->position().toPoint().x(), h - devent->position().toPoint().y() - 1, w, h);
+                DragDropEvent e(ename, m_node, type, DragDropEvent::Text, devent->mimeData()->text().toUtf8().constData(),
+                                modifiers(devent->modifiers()), devent->position().toPoint().x(), h - devent->position().toPoint().y() - 1,
+                                w, h);
```
eloisebrosseau: You should reuse `dropEventModifiers` instead of `devent->keyboardModifiers()` and `devent->modifiers()`

cedrik-fuoco-adsk (approved):

PR #1141, opened 2026-02-19
fix: SG-42272: Fix subsample422_10bit tail pixel handling for non-multiple-of-6 widths

bernie-laberge:
### Linked issues
NA

SDI bug in YCbCr 10 bits for non-multiple-of-6 widths resolutions showing up as a green line on the right side of screen.
I was able to reproduce with a Blackmagic DeckLink 8K Pro outputing in 1280x720p 50Hz.

Cause: The main loop of FastMemcpy.cpp::subsample422_10bit() processed pixels in groups of 6 (v210 packing quantum) but silently dropped any remaining 1-5 pixels at the end of each row, causing incomplete output for non-standard widths.

### Summarize your change.

Added tail handling that follows the same v210 bit packing as the main loop, padding missing pixels with neutral chroma (Cb=Cr=512, Y=0) instead of zeros to avoid green/magenta fringing artifacts at the right edge.

I have also temporarily disabled the automatic installation of the marked down linter because it was blocking on my mac and was preventing me from committing anything. 

Successfully tested on macOS

### DIFF
--- a/src/lib/image/TwkFB/FastMemcpy.cpp
+++ b/src/lib/image/TwkFB/FastMemcpy.cpp
@@ -432,13 +432,16 @@ void subsample422_10bit(size_t width, size_t height, const uint32_t* FASTMEMCPYR
     //  (hence the REV). So Y is the lowest sig bits.
     //
 
+    constexpr size_t pixelsPerGroup = 6; // v210: 6 pixels pack into 4 output words
+    const size_t tailPixels = width % pixelsPerGroup;
+
     for (size_t row = 0; row < height; row++)
     {
         uint32_t* FASTMEMCPYRESTRICT p1 = outBuf + row * outBufStride / sizeof(uint32_t);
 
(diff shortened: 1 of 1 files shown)


src/lib/image/TwkFB/FastMemcpy.cpp
```diff
@@ -432,6 +432,8 @@ void subsample422_10bit(size_t width, size_t height, const uint32_t* FASTMEMCPYR
     //  (hence the REV). So Y is the lowest sig bits.
     //
 
+    const size_t tailPixels = width % 6;
+
```
cedrik-fuoco-adsk: Could you add a comment why 6?

bernie-laberge: Good call. I'll add a more explicit constexpr variable here so that it is self explanatory
On it.

PR #1152, opened 2026-02-26
fix: SG-42334: Fix random crashes involving SoundTrackIPNode

bernie-laberge:
### Linked issues
NA

### Summarize your change.

  - Fixed race conditions in SoundTrackIPNode that caused random crashes due to unsynchronized access to m_fb across threads.                                             
  - evaluateAudioTexture() now uses a short lock-around-read pattern to safely snapshot m_fb before use, and re-checks it under lock before dereferencing to construct the
   IPImage — guarding against propertyChanged() clearing m_fb on the main thread between the two access points.                                                           
  - propertyChanged() now acquires m_fblock before allocating or restructuring m_fb, preventing data races during waveform texture resize.                                
  - clearFB() now acquires m_fblock for the duration of the pixel clear and stats reset, making it fully thread-safe.

  Root Cause

  m_fb (the waveform framebuffer) was read and written from multiple threads without synchronization. evaluateAudioTexture() ran on the audio/render thread while
  propertyChanged() and clearFB() ran on the main thread. A m_fb pointer that was valid at the check could be freed mid-use, leading to use-after-free crashes.

Random crashes had been reported originating from SoundTrackIPNode.

Successfully tested on macOS.

Test Plan

- Enable the Timeline Magnifier via the RV/Tools/Timeline Magnifier menu
- Load media that contains audio
- Play media with audio waveform display enabled and rapidly scrub / resize the timeline — no crash
- Trigger property changes (e.g., resize the waveform panel) during active audio playback — no crash

### DIFF
--- a/src/lib/ip/IPCore/SoundTrackIPNode.cpp
+++ b/src/lib/ip/IPCore/SoundTrackIPNode.cpp
@@ -98,47 +98,61 @@ namespace IPCore
 
     IPImage* SoundTrackIPNode::evaluateAudioTexture(const Context& context)
     {
-        if (!inputs().empty() && m_fb)
-        {
-            if (m_sampleCurrent < m_sampleEnd && graph()->isAudioConfigured())
-            {
-                //
-                //  Try and render
-                //
+        // Lock only to safely read m_fb. We must not hold the lock during
+        // renderAudio() calls — renderAudio() acquires m_fblock internally
+        // and would deadlock.
+        lockFB();
+        const bool hasFB = !inputs().empty() && m_fb;
+        unlockFB();
 
-                AudioCache& cache = graph()->audioCache();
+        if (!hasFB)
+            return 0;
+
+        if (m_sampleCurrent < m_sampleEnd && graph()->isAudioConfigured())
+        {
+            //
+            //  Try and render
+            //
+
+            AudioCache& cache = graph()->audioCache();
+            cache.lock();
+            const size_t packetSize = (cache.packetSize() > 0) ? cache.packetSize() : TWEAK_AUDIO_DEFAULT_PACKET_SIZE;
+            const TwkAudio::Layout layout = cache.layout();
+            const double rate = cache.rate();
+            cache.unlock();
+
+            int channels = TwkAudio::channelsCount(layout);
+            size_t npackets = MAX_SAMPLES_PROCESSED_PER_EVAL / (packetSize * channels);
+
+            for (size_t n = 0; m_sampleCurrent < m_sampleEnd && n < npackets; m_sampleCurrent += packetSize, n++)
+            {
+                AudioBuffer buffer(packetSize, layout, rate, samplesToTime(m_sampleCurrent, rate));
+
                 cache.lock();
-                const size_t packetSize = (cache.packetSize() > 0) ? cache.packetSize() : 512;
-                const TwkAudio::Layout layout = cache.layout();
-                const double rate = cache.rate();
+                const bool found = cache.fillBuffer(buffer);
                 cache.unlock();
 
-                int channels = TwkAudio::channelsCount(layout);
-                size_t npackets = MAX_SAMPLES_PROCESSED_PER_EVAL / (packetSize * channels);
-
-                for (size_t n = 0; m_sampleCurrent < m_sampleEnd && n < npackets; m_sampleCurrent += packetSize, n++)
+                if (found)
                 {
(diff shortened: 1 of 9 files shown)


src/lib/ip/IPCore/SoundTrackIPNode.cpp
```diff
@@ -98,47 +98,61 @@ namespace IPCore
...
+        if (m_sampleCurrent < m_sampleEnd && graph()->isAudioConfigured())
         {
-            if (m_sampleCurrent < m_sampleEnd && graph()->isAudioConfigured())
+            //
+            //  Try and render
+            //
+
+            AudioCache& cache = graph()->audioCache();
+            cache.lock();
+            const size_t packetSize = (cache.packetSize() > 0) ? cache.packetSize() : 512;
```
cedrik-fuoco-adsk: Minor: I would add a comment to explain why the 512

src/lib/ip/IPCore/SoundTrackIPNode.cpp
```diff
@@ -157,31 +171,34 @@ namespace IPCore
...
+            lockFB();
             if (!m_fb)
             {
                 m_fb = new FrameBuffer(w, h, 4, FrameBuffer::UCHAR);
                 m_fb->staticRef();
-                m_stats.clear();
             }
 
-            if (m_fb && (m_fb->width() != w || m_fb->height() != h))
+            if (m_fb->width() != w || m_fb->height() != h)
```
cedrik-fuoco-adsk: Shouldn't you keep checking if m_fb is not null before dereferencing it?

bernie-laberge: Yes totally. On it.

src/lib/ip/IPCore/SoundTrackIPNode.cpp
bernie-laberge: Good idea. And I can see this 512 magic number used in a couple of places in the code. 
There is already a TWEAK_AUDIO_DEFAULT_SAMPLE_RATE, so I'll create a TWEAK_AUDIO_DEFAULT_PACKET_SIZE=512 and explain why it is a good default value.
On it

src/lib/ip/IPCore/SoundTrackIPNode.cpp
bernie-laberge: OK done

src/lib/ip/IPCore/SoundTrackIPNode.cpp
bernie-laberge: OK done

src/lib/ip/IPCore/SoundTrackIPNode.cpp
```diff
@@ -98,47 +98,61 @@ namespace IPCore
...
+        // Re-check m_fb under lock before dereferencing — propertyChanged() on
+        // the main thread may have cleared it since the check above.
+        lockFB();
+        if (!m_fb)
+        {
+            unlockFB();
+            return 0;
         }
+        IPImage* image = new IPImage(this, IPImage::BlendRenderType, m_fb->referenceCopy(), IPImage::OutputTexture);
+        unlockFB();
```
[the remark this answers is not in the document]
bernie-laberge: referenceCopy() is the expensive part, not IPImage's constructor. The allocation and attribute copying happen inside referenceCopy(). Whether you unlock before or after new
  IPImage(...), the critical lock duration is dominated by referenceCopy(). Shaving off the IPImage construction is marginal.

src/lib/ip/IPCore/SoundTrackIPNode.cpp
```diff
@@ -98,47 +98,61 @@ namespace IPCore
...
+
+            int channels = TwkAudio::channelsCount(layout);
+            size_t npackets = MAX_SAMPLES_PROCESSED_PER_EVAL / (packetSize * channels);
+
+            for (size_t n = 0; m_sampleCurrent < m_sampleEnd && n < npackets; m_sampleCurrent += packetSize, n++)
             {
-                //
-                //  Try and render
-                //
+                AudioBuffer buffer(packetSize, layout, rate, samplesToTime(m_sampleCurrent, rate));
```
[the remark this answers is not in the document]
bernie-laberge: Good catch. On it.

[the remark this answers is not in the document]
bernie-laberge: OK done

PR #1158, opened 2026-03-02
ci: Fix detect change job

cedrik-fuoco-adsk:
### Linked issues
none

### Summarize your change.
Add missing actions/checkout step before dorny/paths-filter in the detect-changes job in ci.yml. 

The detect-changes job fails on push, schedule, and workflow_dispatch events with the error:
`fatal: not a git repository (or any of the parent directories): .git`

` dorny/paths-filter` can work without a checkout on pull_request events (using the GitHub API), but on push, schedule, and workflow_dispatch events it requires a local git repository to compare commits. The checkout step was missing, causing the job to fail after merging to main.

CI

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -27,6 +27,7 @@ jobs:
     outputs:
       cmake_changed: ${{ steps.filter.outputs.cmake }}
     steps:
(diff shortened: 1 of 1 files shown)


cedrik-fuoco-adsk:
Forcing the merge to unblock main branch CI

PR #1160, opened 2026-03-02
fix: SG-42402: Fix safeguard causing runtime error when clearing loaded media

chenlj1:
### Summarize your change.

Narrowed the m_beingCleared scope in Session::clear() to only cover graph().reset(), added an early-return guard in Session::propagateEvent during clearing, and added an afterClearSession handler in source_setup.py to restore the display group detection flag.

Having m_beingCleared set at the very top of Session::clear() caused Mu script handlers bound to before-graph-view-change, after-graph-view-change, and graph-node-inputs-changed to crash with "nil argument to function". These events are fired by setViewNode calls that happen before graph().reset(), so the graph is still intact at that point but the returnIfGraphUnsafe guards cause nodeType to return nil whenever m_beingCleared is true, regardless. Moving the scope of m_beingCleared to only cover graph().reset() fixes this.

The early return in Session::propagateEvent ensures no events are propagated to script handlers during graph().reset(), since those handlers call nodeType which returns nil during that phase.

The afterClearSession handler resets _haveNewDisplayGroups back to True after clearing, since the graph-new-node events that normally set it are suppressed by the early return above. Without this, setDisplayFromProfile() would skip colour setup on the next media load, resulting in incorrect colours.

This fix was tested on macOS by loading a media, clearing the session using File->Clear, and reloading the same media to make sure there are no errors and the media looks the same.

### DIFF
--- a/src/lib/ip/IPCore/Session.cpp
+++ b/src/lib/ip/IPCore/Session.cpp
@@ -1696,9 +1696,6 @@ namespace IPCore
 
     void Session::clear()
     {
-        m_beingCleared = true;
-        BOOST_SCOPE_EXIT_ALL(&) { m_beingCleared = false; };
-
         if (!m_beingDeleted)
         {
             userGenericEvent("before-clear-session", "");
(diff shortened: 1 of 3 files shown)


src/lib/ip/IPCore/Session.cpp
```diff
@@ -4575,6 +4576,9 @@ namespace IPCore
         if (m_beingDeleted)
             return EventIgnored;
 
+        if (m_beingCleared)
+            return EventAccept;
+
```
bernie-laberge, line 4581: I like your new PR but I am not 100% sure about this. I am worried that skipping events will lead to unexepected behaviour down the line.
I am tempted to revert this commit which caused this regression and to rework it later:
https://github.com/AcademySoftwareFoundation/OpenRV/commit/02a537b494ba4862f92750a6cea3a69fcb4f10c8

PR #1173, opened 2026-03-03
chore: Dependency check script

mcoliver:
@cedrik-fuoco-adsk where should a script like this go?  Root feels wrong.   Maybe a scripts directory?  Let me know before I move this out of Draft.


bernie-laberge:
How about cmake/scripts ? What do you think @cedrik-fuoco-adsk  ?

PR #1175, opened 2026-03-03
chore: SG-43295: hide noisy build log warnings

mcoliver:
Local build logs are incredibly noisy and often surpass 100,000 lines.  Hiding these two specific warnings cleans up over 10% of the log lines.  Pyside6 builds alone produce over 12,000 lines of these warnings

```
/Users/moliver/dev/_openrv/OpenRV/_build/_deps/rv_deps_pyside6-src/build/.venv/build/pyside6/PySide6/QtCore/PySide6/QtCore/qtimer_wrapper.cpp:1198:18: warning: cast from 'PyObject *(*)(PyObject *)' (aka '_object *(*)(_object *)') to 'PyCFunction' (aka '_object *(*)(_object *, _object *)') converts to incompatible function type [-Wcast-function-type-mismatch]
 1198 |     {"interval", reinterpret_cast<PyCFunction>(Sbk_QTimerFunc_interval), METH_NOARGS, nullptr},
      |                  ^~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
```

Adding
```
…no-cast-function-type-mismatch (GCC/Clang)
-Wno-deprecated-declarations (GCC/Clang)
/wd4996 (MSVC equivalent for deprecated declarations)
```


bernie-laberge (approved):
LGTM
Thank you @mcoliver !

eloisebrosseau:
Hi @mcoliver! I agree there is a lot of noise when building Open RV. However, I don't think we should suppress the warnings instead of actually fixing the underlying issues in our own code. I think we should only mute warnings coming from third-parties since we don't have any control over them. Would you mind only keeping that part? Also, note that I believe `-Wno-cast-function-type-mismatch` only exists for Clang, not GNU, so that might cause other kinds of warnings.

mcoliver:
@eloisebrosseau I don't disagree.  Would love to fix them as well.  I was noticing that a lot of the warnings are due to out of date dependencies (old versions of Qt, pyside, etc..) and my understanding is that this project is hamstrung in modernizing those things by being beholden to the vfx reference platform maximum version requirements and support of various CY years.

eloisebrosseau:
> @eloisebrosseau I don't disagree. Would love to fix them as well. I was noticing that a lot of the warnings are due to out of date dependencies (old versions of Qt, pyside, etc..) and my understanding is that this project is hamstrung in modernizing those things by being beholden to the vfx reference platform maximum version requirements and support of various CY years.

@mcoliver I totally understand where you're coming from, and it's true that the VFX Reference Platform prevents us from modernizing some dependencies, but I also know that some of these warnings are coming from issues in our own code that we could fix. I know it's more work to silence the warnings per dependency, but I feel this would be more appropriate to specifically know what we are silencing instead of acting like our code doesn't have all these warnings by ignoring them.

PR #1176, opened 2026-03-03
fix: SG-43060: update IOtiff.cpp to use public libtiff APIs and standard types

mcoliver:
- Replace internal <tiffiop.h> with public <tiffio.h>
- Use standard <cstdint> types (uint32_t, uint16_t)
- Use standard malloc/free instead of libtiff internals
- Update tag enumeration and field access to use supported public APIs

Building with private methods should be reserved for very specific cases where a public function is not available.  From what I can tell we should be able to use standard malloc and free.  This is a) best practice, and b) will allow RV to be built with precompiled binaries of libtiff (for example installed with homebrew where the private headers are not included in the distribution)

This also remedies crashes I was experiencing when pulling in malformed tiffs with unsupported bit depths (eg 10,12,14).


mcoliver:
I had this as a `refactor` but considering that the PR also makes the program less prone to crashing when receiving malformed or unsupported tiffs I switched to `fix`

mcoliver:
example tif that will segfault the current main but loads fine with the fix in this PR
[color_test_final.tif](https://github.com/user-attachments/files/25859063/color_test_final.tif)

PR #1177, opened 2026-03-03
fix: SG-43400: Fixes message in the console when clearing a session

mcoliver:
Not sure why I was not seeing this before because it looks like the code has been in there a while, however with a clean build of RV I was seeing these things in the RV Console.  This fix ensures it's in a try loop so those things don't appear.

```
File "/Users/moliver/dev/_openrv/OpenRV/_build/stage/app/RV.app/Contents/PlugIns/Python/multiple_source_media_rep.py", line 506, in _on_force_update_media_info
   self._update_media_info(event)
 File "/Users/moliver/dev/_openrv/OpenRV/_build/stage/app/RV.app/Contents/PlugIns/Python/multiple_source_media_rep.py", line 444, in _update_media_info
   sources = rvc.sourcesAtFrame(rvc.frame())
                                ^^^^^^^^^^^
Exception: Exception thrown while calling commands.frame -- exception: "nil argument to function" Traceback (most recent call last):
 File "/Users/moliver/dev/_openrv/OpenRV/_build/stage/app/RV.app/Contents/PlugIns/Python/multiple_source_media_rep.py", line 506, in _on_force_update_media_info
   self._update_media_info(event)
 File "/Users/moliver/dev/_openrv/OpenRV/_build/stage/app/RV.app/Contents/PlugIns/Python/multiple_source_media_rep.py", line 444, in _update_media_info
   sources = rvc.sourcesAtFrame(rvc.frame())
                                ^^^^^^^^^^^
Exception: Exception thrown while calling commands.frame -- exception: "nil argument to function"
```

### DIFF
--- a/src/lib/app/mu_rvui/mode_manager.mu
+++ b/src/lib/app/mu_rvui/mode_manager.mu
@@ -751,17 +751,28 @@ class: ModeManagerMode : MinorMode
 
     method: viewChange (void; Event event, bool on)
     {
-        event.reject();
-
-        let vnode = viewNode(),
-            vtype = nodeType(vnode),
-            name  = if vtype.substr(0,2) == "RV" then vtype.substr(2,0) + "_edit_mode" else "",
-            entry = findModeEntry(name);
-
-        if (entry neq nil)
+        try 
         {
-            activateEntry(entry, on);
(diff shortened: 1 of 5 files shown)


mcoliver:
@bernie-laberge I think this should do the trick on clearing those messages and errors when clearing a session

PR #1178, opened 2026-03-03
chore: update glew to 2.3.1 from 2.2.0 custom hash with patch

mcoliver:
Signed-off-by: Michael Oliver <mcoliver@gmail.com>

GLEW was pulling a non release commit hash and then applying a patch.  The fix has been rolled into an official release 2.3.1 along with other fixes and improvements.  


mcoliver:
Can we please get this in?  It adds `EGL_EXT_gl_colorspace_bt2020_hlg` along with some other extensions that can be useful https://glew.sourceforge.net/log.html

PR #1180, opened 2026-03-03
fix: SG-42881: Fix mp4 video orientation upside down

### LINKED ISSUE #1179: [Bug]: mp4's from iphone appear upside down
### What happened?

Someone sent me a mp4 from their iphone.  I loaded it into RV and it appeared upside down.  Need to fix theorientations

Image

macos

macair

```shell

```

### Environment variables

_No response_

### Extra information

_No response_

### Code of Conduct

mcoliver:
### Linked issues
Fixes #1179 

### Summarize your change.
The rotation's for -90 and 90 were wrong for ffmpeg. This adjusts the orientations appropriately.  

Received a video via sms to an android from my friend who shot it on an iphone.  Pulled the video into RV and it was upside down.  This fixes the orientation.

This PR fixes a bug in `MovieFFMpegReader` that caused some rotated MP4 videos to render upside down (180 degrees off from their intended orientation).

#### Root Cause
The bug stems from a mismatch between how FFmpeg defines rotation metadata, how `rowColumnSwap` physically reorders memory, and how OpenGL texture orientations (`TOPLEFT` vs `BOTTOMRIGHT`) map that memory to the screen. 

Previously, the orientations mapped to `case 90:` and `case 270:` were swapped, effectively causing 90-degree CCW videos to render as 90-degree CW (and vice versa)—a net error of 180 degrees (upside down).

#### Mathematical Breakdown
Here is the step-by-step logic proving the correct orientation mappings:

1. **Physical Memory Swap (`rowColumnSwap`)**
   When `rotate = true`, the code uses `rowColumnSwap` to reorder the planar YUV arrays. By inspecting the pointer arithmetic (`cp = out + h - 1 - i; ... cp += h;`), this maps an input pixel at `(x, y)` to the output coordinate `(height - 1 - y, x)`. This is not just a transpose; it is mathematically a **90-degree CLOCKWISE** rotation of the image in memory.

2. **OpenGL Texture Mapping (`orientation`)**
   The `orientation` property tells the OpenGL renderer (`IPImage::computeOrientationMatrix`) how to flip the texture coordinates:
   * **`TOPLEFT`**: Flips the Y-axis so the texture renders top-down. Applied to our already 90° CW memory, it results in exactly a **90° CW** rotation on screen.
   * **`BOTTOMRIGHT`**: Flips the X-axis and draws bottom-up. When applied to the 90° CW memory, it rotates the texture by an additional 180°. So 90° CW + 180° = **90° CCW** (or 270° CW) on screen.

3. **FFmpeg's Rotation Standard**
   According to FFmpeg's `av_display_rotation_get` documentation, the function returns the rotation angle in **COUNTER-CLOCKWISE** degrees. 

#### The Fix
* `case 90:` means the video metadata is requesting a **90° CCW** rotation. Based on the math above, this requires `BOTTOMRIGHT`. 
* `case 270:` means the video is requesting a 270° CCW (or **90° CW**) rotation. This requires `TOPLEFT`.

The previous logic had these swapped, incorrectly using `TOPLEFT` for `case 90` and `BOTTOMRIGHT` for `case 270`. This PR swaps them back to mathematically align with FFmpeg's CCW orientation standards.

macOS 26

Single change to map the rotation.

Old: (ignore the color shift.  this is another bug that will be handled in a separate PR and happens when you clear a session instead of opening a new session)
Screenshot 2026-03-03 at 3 32 46 PM

With this PR:
Screenshot 2026-03-03 at 3 49 59 PM


mcoliver:
@cedrik-fuoco-adsk builds are failing but not due to my commit.  Can you nudge them along?

mcoliver:
@bernie-laberge another one you may want to fast track

cedrik-fuoco-adsk:
@mcoliver I restarted the windows job. Do you mind sharing a sample video to reproduce this? or any mp4 will do that?

bernie-laberge:
During testing, your commit seems to be affecting the color space detection of media. To be investigated.
Also would it be possible to fill the PR with the requested info when you get the chance. Thanks.

mcoliver:
Hey @bernie-laberge I'm sorry but I'm struggling to understand how that is possible with this commit.  It's a single commit that changes the rotation only (take a look at the code).  That should be independent of any colorspace detection unless there is some other bug further upstream that is somehow pulling color directly from the framebuffer in a nonstandard way.  

As noted in my PR, there is (or was) another bug on macOS  where if you clear a session instead of creating a new session there are issues with color.  My example media shows that but it is unrelated to this commit.  Additional information including sample media and ffprobe info can be found in #1179

mcoliver:
@bernie-laberge added a bit more explanation to the PR to describe the change and root cause

PR #1182, opened 2026-03-04
fix: SG-43399: 1181 fix colorspace changes when session clear and fix memory leak with orphaned sessions

### LINKED ISSUE #1181: [Bug]: When clearing a session, subsequent imports appear much darker
### What happened?

Build and Launch RV 2025.  Load a mp4.  Colorspace looks correct.  command + shift + n (or File --> Clear) and load the video or image in again and it is much darker.

**First load**
Image

**After hitting Clear**
Image

 macos

macair

```shell

```

### Environment variables

_No response_

### Extra information

_No response_

### Code of Conduct

mcoliver:
`source_setup.py` and `ocio_source_setup.py` scripts track whether a new Display Group has been created using a variable named _haveNewDisplayGroups. When you open OpenRV normally, this is configured correctly, and properties like sRGB output on the `#RVDisplayColor` node get activated. However, when you run `File -> Clear`, the C++ `Session::clear()` function destroys and immediately recreates the `Display Group`. The python scripts were failing to track this recreation because of missing event listeners.

When I loaded an MP4 file immediately after clearing the session, the C++ code successfully identified the MP4 as sRGB and enabled the `RVLinearize` step (darkening the image), but the python scripts skipped setting up the `#RVDisplayColor` node. Because `#RVDisplayColor` remained off, the linear image was never lifted back into display color space, rendering it way too dark.

I fixed this by adding an explicit listener for the `after-clear-session` event in both `source_setup.py` and `ocio_source_setup.py`. Now, immediately following a session clear, the internal Python flags are properly reset ensuring that the display color setup happens properly when the next file is loaded. 

Also included a fix for a hardware memory leak.  Every time you open an MP4 and then `File -> Clear`, OpenRV leaks a VideoToolbox hardware decoding session in macOS. Once you hit the macOS limit for concurrent hardware decoder sessions, FFmpeg will silently fail to allocate a new hardware decoder and fall back to the much slower software decoder (libx264).  This fix frees up hardware decoding sessions so they can be available again.

Fixes #1181 

### DIFF
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -313,6 +313,9 @@ namespace TwkMovie
                 av_frame_free(&videoFrame);
             av_frame_free(&inPicture);
             av_frame_free(&outPicture);
+
+            if (hardwareContext.deviceContext)
+                av_buffer_unref(&hardwareContext.deviceContext);
         };
 
         VideoTrack& initFrom(const VideoTrack* t)
@@ -5649,7 +5652,6 @@ namespace TwkMovie
         {
             VideoTrack* track = m_videoTracks[i];
             avcodec_free_context(&track->avCodecContext);
-            av_buffer_unref(&track->hardwareContext.deviceContext);
             delete track;
         }
         if (m_audioSamples)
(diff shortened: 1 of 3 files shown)


bernie-laberge:
Two great catches ! Thank you @mcoliver !
We'll make sure to fast track that one.

PR #1186, opened 2026-03-04
feat: 998: SG-42419: Add support for Apple ProRes decode SDK on Linux and Windows

### LINKED ISSUE #998: Add support for Apple ProRes decode SDK on Linux and Windows
OpenRV currently can decode ProRes video using the VideoToolBox path on Apple Silicon devices. But there’s no support for doing high-quality ProRes decode on Linux and Windows. A ProRes SDK is available from Apple (e-mail ProRes@apple.com). This feature would be to add a configuration option to make use of the Apple-provided ProRes SDK once it has been obtained from Apple.

dpmacri:
### Linked issues
Fixes #998 

### Summarize your change.

This change enables support for building in ProRes decode using the ProRes SDK provided by Apple (get freely by e-mailing prores@apple.com). The SDK is used only on Linux and Windows and is enabled by adding:
 `-DRV_DEPS_APPLE_PRORES_SDK_ZIP_PATH='<full path to SDK .zip file>' `and 
-`DRV_FFMPEG_NON_FREE_DECODERS_TO_ENABLE="prores"` to the `rvcfg `command.

The FFMPEG bit is needed to enable FFMPEG to allow the file type.  The decoding is then done by calling out to the SDK from within MovieFFMpeg.cpp file via small wrappers in the newly added AppleProRes.cpp/.h files.

To provide official support for ProRes decode on Linux and Windows.

Tested ProRes 422 and 4444 playback on Linux (Ubuntu 2024.04.1 LTS) and Windows 11.  Also verified that macOS, which doesn't use this path, still builds and functions correctly.


src/lib/image/MovieFFMpeg/AppleProRes.cpp
```diff
@@ -0,0 +1,154 @@
...
+#include <AppleProRes.h>
+
+// Returns the number of threads to use in the Apple ProRes decoder.
+static int getNumProresThreads()
+{
+    // Only evaluate this value once.
+    static int proresDecoderThreads = -1;
+    if (proresDecoderThreads == -1)
+    {
+        const char* evPrefValue = getenv("ORV_PREF_GLOBAL_PRORES_DECODER_THREADS");
```
cedrik-fuoco-adsk: Shouldn't it be RV_PREF_GLOBAL_PRORES_DECODER_THREADS? @dpmacri

dpmacri: I wasn't sure what to call it.  The "O" was for "Open", but happy to change it.

cedrik-fuoco-adsk: Could you change it to `RV_PREF_GLOBAL_PRORES_DECODER_THREADS`? All the env. variables are prefixed with `RV_`. Thank you!

dpmacri: Changed it!

src/lib/image/MovieFFMpeg/AppleProRes.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+
+    // Now decode the frame
+    int bytesDecoded = PRDecodeFrame(prdecoder, videoPacket->data, videoPacket->size, &prpixbuf, kPRFullSize, 0);
+    if (bytesDecoded != videoPacket->size)
+    {
+        av_log(avctx, AV_LOG_ERROR, "Expected to decode %d bytes but decoded %d\n", videoPacket->size, bytesDecoded);
+        assert(0);
+        return AVERROR(EINVAL);
+    }
```
cedrik-fuoco-adsk, line 125: Hi @dpmacri, I tried your branch with a media (ProRes 4444) and I got a lot of errors when loading the media:
```
...
ERROR: Expected to decode 5543936 bytes but decoded 5543305
ERROR: Expected to decode 5539840 bytes but decoded 5538668
ERROR: Expected to decode 5541888 bytes but decoded 5540534
ERROR: Expected to decode 5537792 bytes but decoded 5537523
ERROR: Expected to decode 5545984 bytes but decoded 5544803
ERROR: Expected to decode 5543936 bytes but decoded 5542725
ERROR: Expected to decode 5556224 bytes but decoded 5555950
...
```

My solution was the following changes:

```suggestion
    // Now decode the frame
    // PRDecodeFrame returns the number of bytes consumed from the compressed
    // buffer, which may be less than videoPacket->size because the container
    // (MOV, MXF, MKV, etc.) can report a larger sample size than the actual
    // ProRes frame.  A negative return value indicates a decode error.
    int bytesDecoded = PRDecodeFrame(prdecoder, videoPacket->data, videoPacket->size, &prpixbuf, kPRFullSize, 0);
    if (bytesDecoded <= 0)
    {
        av_log(avctx, AV_LOG_ERROR, "PRDecodeFrame failed: returned %d (packet size %d)\n", bytesDecoded, videoPacket->size);
        return AVERROR(EINVAL);
    }
```

dpmacri: That's good change.  My test files obviously didn't hit that but your solution is the better approach either way.

cedrik-fuoco-adsk (approved):
I'll re-test but otherwise, looks good to me now! Thank you @dpmacri!

Only thing left is to run the pre-commit.

src/lib/image/MovieFFMpeg/AppleProRes.cpp
```diff
@@ -0,0 +1,157 @@
...
+        {
+            return AVERROR(ENOMEM);
+        }
+    }
+
+    // Now decode the frame
+    // PRDecodeFrame returns the number of bytes consumed from the compressed
+    // buffer, which may be less than videoPacket->size because the container
+    // (MOV, MXF, MKV, etc.) can report a larger sample size than the actual
+    // ProRes frame.  A negative return value indicates a decode error.
```
cedrik-fuoco-adsk: I have an odd issue in our Jenkins CI. It seems like the indentation is using the wrong character.

Lines 117-119 uses non-breaking space characters (U+00A0) as indentation instead of regular ASCII spaces (U+0020).


```
/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:117:1: error: extended character   is not valid in an identifier

  117 |     // buffer, which may be less than videoPacket->size because the container

      | ^

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:117:1: error: extended character   is not valid in an identifier

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:117:1: error: extended character   is not valid in an identifier

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:118:1: error: extended character   is not valid in an identifier

  118 |     // (MOV, MXF, MKV, etc.) can report a larger sample size than the actual

      | ^

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:118:1: error: extended character   is not valid in an identifier

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:118:1: error: extended character   is not valid in an identifier

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:119:1: error: extended character   is not valid in an identifier

  119 |     // ProRes frame.  A negative return value indicates a decode error.

      | ^

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:119:1: error: extended character   is not valid in an identifier

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:119:1: error: extended character   is not valid in an identifier

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp: In member function 'int AppleProResContext::decode_frame(AVCodecContext*, AVFrame*, AVPacket*)':

/mnt/data/jenkins/linux_rhel9_rv_0/OpenRV/src/lib/image/MovieFFMpeg/AppleProRes.cpp:117:1: error: '\U000000a0\U000000a0\U000000a0' was not declared in this scope

  117 |     // buffer, which may be less than videoPacket->size because the container
```


I have a patch file if you wanted to look at it and apply it. I assume it's due the copy from the suggestion I made earlier. Not sure what happend.

[breaking_spaces.patch](https://github.com/user-attachments/files/26147353/breaking_spaces.patch)

dpmacri: Okay, I fixed it up and pushed that change.

bernie-laberge (approved):
Thank you @dpmacri for this great contribution !

bernie-laberge:
FYI: This PR has been sent for QA Testing on all platforms.

src/lib/image/MovieFFMpeg/AppleProRes.h
```diff
@@ -0,0 +1,65 @@
...
+        if (this == &other)
+        {
+            return *this;
+        }
+        pixfmt = other.pixfmt;
+        avPixelFormat = other.avPixelFormat;
+        clearPRPixBuf();
+        prdecoder = nullptr;
+        return *this;
+    }
```
cedrik-fuoco-adsk, line 60: We encountered a crash on Windows after loading a prores media and clearing the session (File > clear). 

I have a fix in a branch. Do you mind porting the fix in your branch? I can also do a follow-up PR after this PR is merged if you prefer. @dpmacri 

The fix:
https://github.com/cedrik-fuoco-adsk/OpenRV/commit/db6b08fac3d905bb14f24e66c8d94677f9b2d341

dpmacri: Thanks @cedrik-fuoco-adsk !  I'll port and push the changes.

PR #1192, opened 2026-03-06
build: SG-42175: Update AJA SDK version to 17.6.0

### LINKED ISSUE #1110: Dependency Update: aja 17.1 -> 17.6
Update dependency for bug fixes, security improvements, and new features

eloisebrosseau:
### Linked issues

Fixes #1110 

### Summarize your change.

Update the SDK version to the newest available (i.e. 17.6.0). Note that we need `17.6.0.hotfix1` otherwise the build fail with a missing header file from AJA. `DAJANTV2_DISABLE_PLUGIN_LOAD` was also added to the configure options since we don't actually need `mbedtls` with our usage of the AJA SDK. Same thing for `DAJANTV2_DISABLE_DRIVER` that was unnecessary building the AJA drivers as part of the RV build on Linux. Moreover, since the AJA library name on Windows now matches the naming convention on macOS and Linux, we don't need to specially set it to `ajantv2_vs143_MT` anymore.

There is a new version of the AJA SDK. No changes were needed outside of the build files as this was mainly a maintenance release.

Enabling presentation mode with the Io 4K Plus and the Kona 5 card was tested on macOS.


bernie-laberge (approved):
Nice!

cedrik-fuoco-adsk (approved):

PR #1197, opened 2026-03-11
fix: Improve scrubbing annotated frames

rogernelson:
### Summarize your change.

There was a bug that was causing cached Paint FBOs to be released between frames.  In the original code, the paint FBOs were explicitly ignored when cleaning garbage collecting from the cache. However, they were still marked as available, allowing them to be reused. This fix simply avoids that call for Paint FBOs so they remain cached when changing frames (ie when scrubbing or playing back in a loop). That is the core of the fix, but I also implemented some safety mechanisms that were not done in the original implementation to be on the safe side.  Everything but the one `continue` statement to avoid marking the FBO as available is related to that.

The first safety check is to set an age limit in the Paint FBOs. If we haven't seen them in 300 renders (chosen rather arbitrarily by me, but can be adjusted), it's not likely we'll see them again, so we can safely evict them.

I also added a second safety check against unbound cache growth.  If there are many large annotated frames all in a row, we can end up generating a huge cache. Although this scenario is unlikely as under normal circumstances the age limit will free the cached FBOs, we don't want to cause problems for the GPU by caching too much, so let's set a limit of 32 Paint FBOs in the cache. If we reach this limit, we'll sort by age and evict the oldest entries, effectively creating an LRU cache for the Paint FBOs.

Other than that, I just did some cleanup:  
- The FBO deletion code was cut-pasted everywhere so I just made a helper function to do it
- Found all the FBO deletion code and made it use the new method
- Fixed an assert I encountered in src/lib/geometry/TwkPaint/Path.cpp when testing it: rDirectionCoords was getting dir pushed to it twice. This didn't cause any issues, but does trigger an assert in debug.

Improve playback performance with annotated frames

Mac OS 26.3.1 Mac Pro M4

I created a test scenario with 500 annotations per frame and 15 points per stroke on 5 sequential frames.
Rendering speed before changes: 6.8 fps
Rendering speed after changes: 59.8 fps

Nearly a 10x speedup. 

Screenshot 2026-03-11 at 15 02 36
Screenshot 2026-03-11 at 15 40 01

### DIFF
--- a/src/lib/ip/IPCore/ImageFBO.cpp
+++ b/src/lib/ip/IPCore/ImageFBO.cpp
@@ -10,25 +10,9 @@
 
 namespace
 {
-    // Number of render cycles an unused regular FBO is kept before being freed.
-    // A small grace window (roughly 200ms at 24fps) prevents thrashing when a
-    // frame is temporarily skipped during cache warm-up or off-screen evaluation.
-    constexpr size_t FBO_AGE_LIMIT = 5;
-
-    // Number of render cycles an idle paint cache FBO is kept before being freed.
-    // Paint cache FBOs are pinned (never returned to the general pool) so they
-    // survive scrubbing. This longer threshold only fires when the annotation has
-    // been genuinely abandoned — cleared, source removed, or the user has been on
-    // a completely different part of the timeline — preventing unbounded memory
-    // accumulation from orphaned cache entries.
-    constexpr size_t PAINT_FBO_AGE_LIMIT = 300;
-
-    // Maximum number of paint cache FBOs retained simultaneously. Each pinned
-    // paint FBO holds a full-resolution composited image (e.g. ~33MB at 4K 8-bit),
-    // so without a cap a long annotated timeline could exhaust GPU memory. When
-    // the limit is exceeded the least-recently-used entries (lowest fullSerialNum)
-    // are evicted first, keeping the most recently visited frames hot.
-    constexpr size_t MAX_PAINT_FBO_COUNT = 32;
+    // max length of FBO ids output to debug logs
+    //
+    constexpr size_t FBO_ID_DEBUG_LOG_LIMIT = 50;
 } // namespace
 
 namespace IPCore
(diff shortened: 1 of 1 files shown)


src/lib/ip/IPCore/ImageFBO.cpp
```diff
@@ -234,15 +266,53 @@ namespace IPCore
...
+                    destroyImageFBO(i);
+                    m_imageFBOs[q] = m_imageFBOs.back();
+                    m_imageFBOs.pop_back();
+                    q--;
+                }
+                else
+                {
+                    paintFBOCount++;
+                }
+                continue;
```
rogernelson, line 314: This is the key to the fix here (so the `i->available = true` line below is skipped). This on its own fixes the bug.  The rest is just cautionary to avoid growing the cache indefinitely.

src/lib/ip/IPCore/ImageFBO.cpp
```diff
@@ -234,15 +266,53 @@ namespace IPCore
...
+                // findExistingPaintFBO() can ever reclaim them.
+                //
+                // Evict only after a long idle period — this handles the case where
+                // annotations were cleared or the source was removed, preventing
+                // unbounded memory accumulation.
+                if (age > PAINT_FBO_AGE_LIMIT)
+                {
+                    if (m_imageFBOLog)
+                    {
+                        string s = i->identifier;
```
eloisebrosseau: I know there is a lot of one-character variables in the code, but I believe we should try to use more meaningful names for the new code we add.

src/lib/ip/IPCore/ImageFBO.cpp
```diff
@@ -234,15 +266,53 @@ namespace IPCore
...
+                        if (s.size() > 50)
```
eloisebrosseau: What is 50 here? Shouldn't this be declared as a constant in this function? Also, I see that `PAINT_FBO_AGE_LIMIT` and `FBO_AGE_LIMIT` are only used at one place in the code. Shouldn't they be declared as local variables instead?

rogernelson: The 50 (and the s variable name) just truncates the string identifier to 50 characters. It's something that I just cut and pasted that appears everywhere, but I can put another constant for it.

PR #1198, opened 2026-03-13
feat: SG-27382: Add 12G SDI single-link support for the AJA Kona 5 card

### LINKED ISSUE #860: AJA 12G output over single link

eloisebrosseau:
### Linked issues

Fixes #860 

### Summarize your change.

Note that this PR is based on #1192 that is updating the SDK version used to 17.6.0. The only relevant files to review are KonaVideoDevice.cpp and KonaVideoDevice.h.

This PR adds three new possible data formats if the device can support 12G SDI Single-link on its third port (i.e. SDI Out 3):
- 10 Bit 12G Single-link YCrCb 4:2:2 (SDI Out 3)
- 10 Bit 12G Single-link RGB (SDI Out 3)
- 12 Bit 12G Single-link RGB (SDI Out 3)

Note that the first format enables 6G SDI single-link for low frame rates, and 12G for higher frame rates in YUV.

Since the retail firmware of the Kona 5 doesn't have a built-in TSI muxers, the only supported video formats are the quad ones. Therefore, we need to ensure that 12G statements are executed before quad statements. Moreover, the 4K/8K formats options are simply ignored for the 12G single-link case as TSI is the only possible option if we want to wire the frames to one SDI output.

To avoid refactoring too much code that would affect other AJA cards, the idea here is to only support the 12G SDI single-link workflow based on the Kona 5 retail firmware that only handle it on port 3. To do so, two new functions were added: `route12GSingleLinkYUV()` and `route12GSingleLinkRGB` where all the routing is directly handled to go from the frame buffers, to the TSI multiplexers, to the Color spaces converters (if the output format is YUV), to the Dual Link Out (if the output format is RGB), to the third port. To match the rest of the code, we are always using FB 1 and FB 2. The only difference is with the output port.

This is a feature that was requested by multiple people in the last few years for the AJA Kona 5.

Tests were performed on macOS 26.3 using a Kona 5 with the retail firmware and a BMD Smartview 4K monitor.
Several YUV and RGB files were tested with the variety of 4K video formats (both UHD and DCI) supported by the card at different resolutions with the 10-bit YUV, 10-bit RGB, and 12-bit RGB data formats that were added specifically to perform 12G SDI Single-link on the third port. Note that the monitor used doesn't support 12G SDI RGB, so only the routing was validated using AJA ntv2watcher tool.

### DIFF
--- a/src/plugins/output/AJADevices/AJADevices/KonaVideoDevice.h
+++ b/src/plugins/output/AJADevices/AJADevices/KonaVideoDevice.h
@@ -69,9 +69,11 @@ namespace AJADevices
         ThreeG = 1 << 3,
         LegalRangeY = 1 << 4,
         FullRangeY = 1 << 5,
+        TwelveG = 1 << 6,
 
         RGB_3G = RGB444 | P2P | ThreeG,
         RGB_DualLink = RGB444 | P2P | DualLink,
+        YUV_6G = P2P | TwelveG,
     };
 
     struct KonaDataFormat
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):
Superb
Thank you @eloisebrosseau !

cedrik-fuoco-adsk (approved):

PR #1215, opened 2026-03-30
fix: SG-35913: Fix importing files with long path names on Windows

eloisebrosseau:
### Summarize your change.

Replace `fileExistsWithLongPathSupport ` with a manifest file enabling `longPathAware`.

RV was failing to load media files with long path names even with the fix from `fileExistsWithLongPathSupport()`. In this function, a prefix was added in front of the file path known by RV to make it a long path. However, the path contained forward slashes, but the extended path prefix is expecting backslashes. This was causing RV to not recognize long path names. A simple fix would have been to replace the forward slashes with backslashes. However, based on the [Windows documentation](https://learn.microsoft.com/en-us/windows/win32/fileio/maximum-file-path-limitation?tabs=registry), using a manifest file to make the app long path aware in conjonction with enabling the long path registry on the machine looks like the preferred approach. That way, multiple functions from the Windows API will no longer be restricted by the MAX_PATH length.

Note that there is a Qt reported bug specifically for drag and drop of single files with a long path names. This fixes only drag and dropping of folders, import importing files from `File > Open`, and launching RV through the terminal with a file with a long path name.

### Describe what you have tested and on which operating system

Testing drag and dropping of folders, importing files from `File > Open`, and launching RV through the terminal with a file with a long path name.

**BEFORE**
https://github.com/user-attachments/assets/80246ff9-b3a8-45b3-b7e3-e8e78f6e8e87

**AFTER**
https://github.com/user-attachments/assets/4cda2fec-f44a-4d3f-8bf5-aa28d9a7dc37


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #1221, opened 2026-04-09
fix: GLSL version directive never being applied

cedrik-fuoco-adsk:
### Linked issues
n/a

### Summarize your change.

Remove `#version 150` from inside `#if __VERSION__ >= 150` blocks in all TwkGLF shaders, and inject it from C++ via `glShaderSource` so it is the true first token in the concatenated shader source.

- `BasicGLProgram.cpp`: added `basicGLVersionHeader()` helper that
  returns `"#version 150\n"` for GL3+ or `""` for GL2, and passes it as
  the first element of a 2-string array to `glShaderSource` for both
  vertex and fragment shaders.
- All 28 `.glsl` files under `TwkGLF/glsl/`: removed the now-redundant
  `#version 150` line from inside the `#if __VERSION__ >= 150` block.
  The guards themselves are kept for GL2/GL3 feature selection.

`#version` must be the first token in GLSL source per spec. 
The previous code placed `#version 150` inside a `#if __VERSION__ >= 150` guard

Windows with AMD GPU.

N/A


bernie-laberge (approved):
LGTM

eloisebrosseau (approved):

PR #1223, opened 2026-04-10
build: Small build improvement for speed and fix CI for CY2023

cedrik-fuoco-adsk:
### Linked issues
none

### Summarize your change.
- Speed up the Windows CI build by introducing parallel builds (JOM) for OpenSSL and LibRaw
- Enabling **--parallel** for OCIO
- Fixing race conditions that arise from parallel compilation
- Add back OCIO 2.5.x for CY2023

- The windows build is very slow and any minor improvement helps overtime.
-  OpenSSL PDB contention: 
    Multiple cl.exe processes writing to the same app.pdb under JOM causes sporadic Error 2 failures. 
    Fixed by setting **_CL_=/Z7** to embed debug info

- Python test race condition: 
   **test_python.py** was moving the Python install directory to a temp location for relocatability testing. 
      Any concurrent build (e.g OCIO) needing Python.h would fail with **C1083: Cannot open include file**. 
      Fixed by copying instead of moving.

Windows and CI (all platforms)

  • `.github/actions/build-windows/action.yml`: Install JOM via Chocolatey and add it to PATH
  • `cmake/dependencies/raw.cmake`: Use JOM (if available) instead of nmake for LibRaw builds 
  • `cmake/dependencies/ocio.cmake`: Re-enable `--parallel ` for OCIO Windows builds
  • `cmake/dependencies/oiio.cmake`: Skip OCIO dependency for CY2023 (same behavior as before)
  • `cmake/defaults/CY2023.cmake`: Add OIIO 2.5.19.1 version/hash for CY2023
  • `src/build/make_openssl.py`: Use JOM with /FS + /Z7 for parallel OpenSSL builds; add no-tests to configure
  • `src/build/test_python.py`: Use copytree instead of move so the install dir stays in place during testing (fixes the `Python.h` not found race condition)
  • `src/lib/base/stl_ext/CMakeLists.txt`: Enable unity build for stl_ext


bernie-laberge (approved):
LGTM

eloisebrosseau (approved):

PR #1235, opened 2026-04-23
feat: SG-42241: RV: Generate thumbnail before filmstrips

chenlj1:
### Summarize your change.

Defer filmstrip generation until after the thumbnail is available so the thumbnail always loads first.

We want to replace the fallback icons as soon as possible so that the user can already see a preview of the loaded media.

Tested on MacOS by loading in media and making sure that the thumbnail preview shows up first.


bernie-laberge (approved):
LGTM

eloisebrosseau (approved):

cedrik-fuoco-adsk (approved):

bernie-laberge:
FYI @savoiei : I tested this small optimization on my end so we will merge it if you don't mind.

PR #1238, opened 2026-04-24
fix: SG-42945: [RV] Thumbnails from clips that are unavailable in RV session will display a black frame

chenlj1:
### Summarize your change.

Added a guard to prevent thumbnail and filmstrip generation of a missing source.

Previously, if you save a session and then delete some of the sources and load the saved session, the missing sources will have a black thumbnail instead of the default fallback icon.

Tested on MacOS by saving a session and deleting some of the sources. Reloaded the saved session and verified that missing sources show the fallback icon.


eloisebrosseau (approved):

bernie-laberge (approved):

bernie-laberge:
Note: I discussed it with Ling and I am no longer able to reproduce the original issue. It behaves as expected without this fix (the paperclip icon gets shown). So I will close this PR.

PR #1249, opened 2026-04-30
fix: SG-42890: Fix BADACCESS X error when running RVIO commands on Rocky Linux 9

eloisebrosseau:
### Summarize your change.

- [X]  Make the writer thread call MovieRV::uninit instead of the main thread
- [X] Replace dummyDev raw pointer with an FBOVideoDevice unique_ptr

https://github.com/AcademySoftwareFoundation/OpenRV/pull/310 added `MovieRV::uninit` in the main function of RVIO to fix a GL error on exit on Rocky Linux 9. However, in a more recent version of the Rocky Linux 9, we are now getting a BADACCESS X error when running RVIO commands from the command line. An alert panel is also being displayed when submitting an image sequence through Screening Room after each step. This was caused by the main thread trying to access the OpenGL rendering context used by the writer thread when calling `MovieRV::uninit` at the end of the main function.

It seems that the initial GL error was actually caused by the FBOVideoDevice raw pointer created as a dummy video device at the beginning of the main RVIO function by the main thread to avoid a segmentation fault. This raw pointer was never deleted, and this was causing a GL error when the `GLSyncObjectARBSync` destructor was called at the end of the program. Replacing the raw pointer with a unique_ptr fixed the leaking pointer that was the root cause of the GL error.

`MovieRV::uninit` was not technically needed to fix the original GL error, but it's a good practice to keep it to make sure we uninit any FBOs that were allocated. The only difference is that it should have been called by the writer thread instead of the main thread since it is the one that actually created and used the FBOVideoDevice and the OpenGL rendering context during the execution of RVIO.

Submitting an image sequence through Screening Room was tested on Rocky Linux 9.6, Windows 11, and macOS 26.3.
The following commands that were called during the submit process of Screening Room were also manually called from the command line to make sure no error remains:

- `rvio movie.####.dpx -o /tmp/thumb.#.jpg`
- `rvio movie.####.dpx -o /tmp/out.mov`
- `rvio movie.####.dpx -resize 240 132 -outres 1200 132 -o /tmp/filmstrip.jpg`

### DIFF
--- a/src/bin/imgtools/rvio/main.cpp
+++ b/src/bin/imgtools/rvio/main.cpp
@@ -970,14 +970,8 @@ int utf8Main(int argc, char* argv[])
     setEnvVar("LC_ALL", "C");
     TwkFB::ThreadPool::initialize();
 
-    //
-    //  XXX dummyDev is leaking here.  Best would be to pass it to the App so
-    //  that it could delete it after startup, since it's no longer needed at
-    //  that point.
-    //
-
 #ifdef RVIO_HW
-    TwkGLF::FBOVideoDevice* dummyDev = new TwkGLF::FBOVideoDevice(0, 10, 10, false);
+    auto dummyDev = std::make_unique<TwkGLF::FBOVideoDevice>(nullptr, 10, 10, false);
 #else
     TwkGLF::OSMesaVideoDevice* dummyDev = new TwkGLF::OSMesaVideoDevice(0, 10, 10, true);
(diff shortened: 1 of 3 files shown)


bernie-laberge (approved):
Nice fix and cleanup ! 
Thank you @eloisebrosseau !

cedrik-fuoco-adsk (approved):

rogernelson (approved):
Nice and clean, good job!

PR #1263, opened 2026-05-07
feat: SG-35309: Set and save OCIO environment variable if missing and config is chosen

deltag0:
### Summarize your change.
When a user loads a valid OCIO config after opening a file and sets the OCIO as an active display with the environment variable `OCIO` not set, there used to be an error log. The error log is now gone and the environment variable is set after selecting a config.
Moreover, the config path is saved to user preferences after choosing it.

Redundant error logs and console pop ups because once the user loads the OCIO config with no environment variable there's already an error log.

Tested on Mac CY2025

N/A

snipit_2026-05-07T11-15-47 251281


src/plugins/rv-packages/ocio_source_setup/ocio_source_setup.py
```diff
@@ -668,7 +670,8 @@ def addMenu(root, isSingle):
                 )
             )
 
-        final = [
+        final = [("Change Config...", self.selectConfig, None, None)]
```
bernie-laberge: I would list it at the end of the menu. Not at the beginning since it is something an RV user will rarely do. 
Please add a line separator

deltag0: fixed

src/plugins/rv-packages/ocio_source_setup/ocio_source_setup.py
```diff
@@ -716,6 +720,11 @@ def __init__(self):
         except ImportError:
             pass
 
+        if os.getenv("OCIO") is None:
```
bernie-laberge, line 735: I would add a comment for this section so that the business logic is explained.
something like:
"Restore OCIO env var from saved settings so a previously-chosen config persists across sessions; an externally-set OCIO env var takes precedence."

bernie-laberge (approved):
LGTM

src/plugins/rv-packages/ocio_source_setup/ocio_source_setup.py
deltag0: fixed

src/plugins/rv-packages/ocio_source_setup/ocio_source_setup.py
```diff
@@ -716,6 +720,13 @@ def __init__(self):
         except ImportError:
             pass
 
+        # Restore saved OCIO config from previously loaded config
+        # An externally set OCIO env var takes precendence
+        if os.getenv("OCIO") is None:
+            config = commands.readSettings("ocio_source_setup", "ocio_config", "")
+            if config != "" and os.path.isfile(config) and config.endswith("ocio"):
```
eloisebrosseau: i think it might be better to check if the config extension is `.ocio` instead of simply validating that config ends with `ocio`. Note that using `Path` from `pathlib` is more modern and the prefered way over `os.path`. You could replace `os.path.isfile()` with `Path.is_file()` and `config.endswith("ocio")` with `config.suffix == ".ocio"`

doug-walker: Is the extension check necessary?  OCIO will load other formats such as .ocioz files. I don't think we should block OCIO from trying to load them.

KevinJW: For an example, at work for historical reasons the VFX side of the business uses `ocio.conf` for the majority of the files. And it is really frustrating when tools hardcode matching patterns into file browsers etc making it so we can't actually load our files.

cedrik-fuoco-adsk: I agree with @doug-walker. I don't think we should filter the file extension.

deltag0: Removed the extension check

eloisebrosseau (approved):

markreidvfx:
Setting environmental variables at runtime while other thread are running and is not safe. If any thread just so happens to be reading one at the same time as another thread is setting a realloc can happen and getenv will read after free and potential crash. Linux at least performs not shared locking between getenv and setenv as far as I'm aware. I recently fix a bug with some of our RV plugins related exactly to this.

bernie-laberge:
> Setting environmental variables at runtime while other thread are running and is not safe. If any thread just so happens to be reading one at the same time as another thread is setting a realloc can happen and getenv will read after free and potential crash. Linux at least performs not shared locking between getenv and setenv as far as I'm aware. I recently fix a bug with some of our RV plugins related exactly to this.

According to our robot friend, @markreidvfx is right:
  Mark is right, and this matters for the PR

  The claim is technically correct, well-documented, and has bitten real applications repeatedly:

  - environ is a flat char** array maintained by libc. setenv/putenv may realloc it and replace string pointers.
  - glibc's setenv takes an internal lock, but getenv does not acquire it — see sourceware #15607 (https://sourceware.org/bugzilla/show_bug.cgi?id=15607). POSIX explicitly only guarantees getenv thread-safety
  in the absence of concurrent modifications.
  - The crash mode is exactly what Mark described: a getenv caller dereferences a char* that setenv just freed.
  - This isn't theoretical — Rust shipped a CVE around it (CVE-2024-43807 / set_var (https://blog.rust-lang.org/2024/09/19/Rust-1.82-pre-announce-set_var-unsafe.html) marking set_var unsafe in 1.82), Go has the
   same hazard, OpenSSL/curl have warnings about it. It's a known footgun, not an edge case.

I think we should explore an alternative way of setting an OCIO config without having to set the OCIO environment variable.
Thank you @markreidvfx !

deltag0:
> > Setting environmental variables at runtime while other thread are running and is not safe. If any thread just so happens to be reading one at the same time as another thread is setting a realloc can happen and getenv will read after free and potential crash. Linux at least performs not shared locking between getenv and setenv as far as I'm aware. I recently fix a bug with some of our RV plugins related exactly to this.
> 
> According to our robot friend, @markreidvfx is right: Mark is right, and this matters for the PR
> 
> The claim is technically correct, well-documented, and has bitten real applications repeatedly:
> 
> * environ is a flat char** array maintained by libc. setenv/putenv may realloc it and replace string pointers.
> * glibc's setenv takes an internal lock, but getenv does not acquire it — see sourceware #15607 (https://sourceware.org/bugzilla/show_bug.cgi?id=15607). POSIX explicitly only guarantees getenv thread-safety
>   in the absence of concurrent modifications.
> * The crash mode is exactly what Mark described: a getenv caller dereferences a char* that setenv just freed.
> * This isn't theoretical — Rust shipped a CVE around it ([CVE-2024-43807](https://github.com/advisories/GHSA-c856-w373-jm6c) / set_var (https://blog.rust-lang.org/2024/09/19/Rust-1.82-pre-announce-set_var-unsafe.html) marking set_var unsafe in 1.82), Go has the
>   same hazard, OpenSSL/curl have warnings about it. It's a known footgun, not an edge case.
> 
> I think we should explore an alternative way of setting an OCIO config without having to set the OCIO environment variable. Thank you @markreidvfx !

Removed setting the environment variable and applied broader fix.

PR #1264, opened 2026-05-08
build: Fix oiio build clashing with Homebrew on macOS

bernie-laberge:
### Linked issues
NA

The OIIO build was failing on my macOS Tahoe 26.3 because of clashes with incompatible versions of OpenEXR and Imath in my local homebrew directory.

### Summarize your change.

- Fixed the OIIO.cmake file to make sure that the OpenEXR and Imath includes are taken from the RV_DEPS_OPENEXR_ROOT_DIR/include and RV_DEPS_IMATH_ROOT_DIR/include respectively.

Also in this commit:
- Fixed the About RV's Numpy version which was missing in action because of this syntax error in src/lib/app/RvCommon/CMakeLists.txt

Successfully tested on macOS Tahoe 26.3

### DIFF
--- a/cmake/dependencies/oiio.cmake
+++ b/cmake/dependencies/oiio.cmake
@@ -84,24 +84,34 @@ IF(RV_DEPS_CMAKE_PREFIX_PATH)
   ENDIF()
   STRING(APPEND _oiio_cache_content "set(CMAKE_PREFIX_PATH \"${_oiio_clean_prefix}\" CACHE STRING \"\" FORCE)\n")
 ENDIF()
-# When deps come from a package manager (Conan), block find_package from searching the Homebrew shared prefix. OIIO's find_package(Boost) would otherwise find
-# Homebrew's Boost headers (newer version), causing ABI mismatches with Conan's Boost library. CMAKE_IGNORE_PREFIX_PATH only affects find_package, not
-# find_program (so Ninja and compilers still resolve). With Homebrew-only builds, all Boost is from the same source so no contamination occurs.
-IF(RV_CONAN_CMAKE_PREFIX_PATH
-   AND APPLE
+
+# Detect the Homebrew prefix on macOS (e.g. /opt/homebrew on arm64, /usr/local on Intel) for use by the contamination guards below.
+SET(_oiio_brew_prefix
+    ""
 )
+IF(APPLE)
   EXECUTE_PROCESS(
     COMMAND brew --prefix
     OUTPUT_VARIABLE _oiio_brew_prefix
     OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET
     RESULT_VARIABLE _oiio_brew_rc
   )
-  IF(_oiio_brew_rc EQUAL 0
-     AND _oiio_brew_prefix
-  )
-    STRING(APPEND _oiio_cache_content "set(CMAKE_IGNORE_PREFIX_PATH \"${_oiio_brew_prefix}\" CACHE STRING \"\" FORCE)\n")
+  IF(NOT _oiio_brew_rc EQUAL 0)
+    SET(_oiio_brew_prefix
+        ""
+    )
   ENDIF()
 ENDIF()
+
+# When deps come from a package manager (Conan), block find_package from searching the Homebrew shared prefix. OIIO's find_package(Boost) would otherwise find
+# Homebrew's Boost headers (newer version), causing ABI mismatches with Conan's Boost library. CMAKE_IGNORE_PREFIX_PATH only affects find_package, not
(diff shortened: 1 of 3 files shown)


cmake/dependencies/oiio.cmake
```diff
@@ -180,11 +180,31 @@ LIST(APPEND _configure_options "-DTIFF_ROOT=${RV_DEPS_TIFF_ROOT_DIR}")
...
+  STRING(APPEND _oiio_cxx_flags " -I${RV_DEPS_BOOST_ROOT_DIR}/include")
+ENDIF()
+LIST(APPEND _configure_options "-DCMAKE_CXX_FLAGS=${_oiio_cxx_flags}")
+
+# Redirect OIIO's find_package(libjpeg-turbo) to our RV_DEPS build. Without this, OIIO finds Homebrew's libjpeg-turbo CMake config first; its IMPORTED target's
+# INTERFACE_INCLUDE_DIRECTORIES = /opt/homebrew/include then gets added to OIIO's -isystem list, contributing to the OpenEXR/Imath header contamination above.
+# JPEG_LIBRARY/JPEG_INCLUDE_DIR are ignored when libjpeg-turbo is found via CONFIG mode.
+IF(RV_TARGET_DARWIN)
+  LIST(APPEND _configure_options "-Dlibjpeg-turbo_DIR=${RV_DEPS_JPEGTURBO_ROOT_DIR}/lib/cmake/libjpeg-turbo")
 ENDIF()
```
cedrik-fuoco-adsk, line 212: I'd like to change the gating for those fixes to something more robust for both find-first and build from source path.

Two things:
- The new force-includes and libjpeg-turbo_DIR redirect are gated on `IF(RV_TARGET_DARWIN)`, which fires on every macOS build, even ones with no Homebrew installed.
- The existing Boost -I flag right below at L186-188 is gated on `IF(RV_DEPS_IGNORE_PREFIXES)`, which only fires when find-first's symlink walk detected a mismatch. That means the Boost defense is currently broken for your scenario (pure source build with Homebrew on PATH): same root cause, but the gate misses it.

Both workarounds attack the same problem and should share a gate. `The right gate is Homebrew is on the system AND we're not in the Conan path that already blocks it.` The brew detection is already in the file at L93-98. We could lift it out of the Conan-only branch:


```
SET(_oiio_brew_prefix "")
IF(APPLE)
  EXECUTE_PROCESS(
    COMMAND brew --prefix
    OUTPUT_VARIABLE _oiio_brew_prefix
    OUTPUT_STRIP_TRAILING_WHITESPACE ERROR_QUIET
    RESULT_VARIABLE _oiio_brew_rc
  )
  IF(NOT _oiio_brew_rc EQUAL 0)
    SET(_oiio_brew_prefix "")
  ENDIF()
ENDIF()

IF(RV_CONAN_CMAKE_PREFIX_PATH AND _oiio_brew_prefix)
  STRING(APPEND _oiio_cache_content
    "set(CMAKE_IGNORE_PREFIX_PATH \"${_oiio_brew_prefix}\" CACHE STRING \"\" FORCE)\n")
ENDIF()
```

Then put all three workarounds (Boost -I, OpenEXR/Imath force-includes, libjpeg-turbo redirect) under one gate:
```
SET(_oiio_cxx_flags "")
IF(_oiio_brew_prefix AND NOT RV_CONAN_CMAKE_PREFIX_PATH)
  STRING(APPEND _oiio_cxx_flags " -include ${RV_DEPS_OPENEXR_ROOT_DIR}/include/OpenEXR/OpenEXRConfig.h")
  STRING(APPEND _oiio_cxx_flags " -include ${RV_DEPS_OPENEXR_ROOT_DIR}/include/OpenEXR/IexConfig.h")
  STRING(APPEND _oiio_cxx_flags " -include ${RV_DEPS_OPENEXR_ROOT_DIR}/include/OpenEXR/IlmThreadConfig.h")
  STRING(APPEND _oiio_cxx_flags " -include ${RV_DEPS_IMATH_ROOT_DIR}/include/Imath/ImathConfig.h")
  STRING(APPEND _oiio_cxx_flags " -I${RV_DEPS_BOOST_ROOT_DIR}/include")
  LIST(APPEND _configure_options
    "-Dlibjpeg-turbo_DIR=${RV_DEPS_JPEGTURBO_ROOT_DIR}/lib/cmake/libjpeg-turbo")
ENDIF()
IF(_oiio_cxx_flags)
  LIST(APPEND _configure_options "-DCMAKE_CXX_FLAGS=${_oiio_cxx_flags}")
ENDIF()
```

bernie-laberge: That's way better. Thanks !

cedrik-fuoco-adsk (approved):

PR #1265, opened 2026-05-08
fix: SG-43096: Fix Playback buttons offset displacement

deltag0:
### Summarize your change.
~~Changed toolbar UI to be designed in 3 boxes:~~
~~- left box~~
~~- center box~~
~~- right box~~
~~instead of adding the components individually to the toolbar.~~

- Display default resolution of the first image added in layout mode or the currently hovered resolution

On layout mode, we only displayed the resolution when ALL images had the same resolution, causing a button offset.

- Tested that toolbar functionality works the same
- Tested that the playback buttons don't get displaced
- Tested the fix on layout mode

- Testing with very low resolution media


zaifnatra:
this is really cool

src/lib/app/RvCommon/RvBottomViewToolBarBuild.cpp
```diff
@@ -0,0 +1,243 @@
...
+#include <IPCore/Session.h>
+#include <RvApp/Options.h>
+
+namespace Rv
+{
+    using namespace IPCore;
+
+    namespace
+    {
+        constexpr std::string_view playModeDefaultTooltip = "Select playback style";
```
eloisebrosseau: This line is duplicated in RVBottomViewToolBar.cpp

src/lib/app/RvCommon/RvBottomViewToolBarBuild.cpp
```diff
@@ -0,0 +1,243 @@
...
+
+        struct LeftActionDef
+        {
+            const char* iconPath;
+            const char* tooltip;
+            const char* tbstyle;
+            QAction** target;
+        };
+
+        const LeftActionDef leftDefs[] = {
```
eloisebrosseau: You should avoid using c-style arrays and use std::array instead

deltag0: fixed

src/lib/app/RvCommon/RvBottomViewToolBarBuild.cpp
deltag0: fixed

src/plugins/rv-packages/multiple_source_media_rep/multiple_source_media_rep.py
```diff
@@ -281,13 +278,15 @@ def _build_media_representation_widgets(self):
...
-        playback_style_action = None
-        for action in self._bottomToolBar.actions():
-            if action.toolTip() == QtCore.QObject().tr("Select playback style"):
-                playback_style_action = action
+        right_box = self._bottomToolBar.findChild(QtWidgets.QWidget, "rightBox")
+        play_mode_btn = right_box.findChild(QtWidgets.QToolButton, "playModeButton")
+        right_layout = right_box.layout()
+
+        def insert_before_play_mode(widget):
+            right_layout.insertWidget(right_layout.indexOf(play_mode_btn), widget)
```
bernie-laberge: Since that they are only used in this inline method (insert_before_play_mode), I would move play_mode_btn and right_layout in it:
```
        def insert_before_play_mode(widget):
            play_mode_btn = right_box.findChild(QtWidgets.QToolButton, "playModeButton")
            right_layout = right_box.layout()
            right_layout.insertWidget(right_layout.indexOf(play_mode_btn), widget)
```

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
```diff
@@ -9,13 +9,17 @@
 #include <RvCommon/RvApplication.h>
 #include <RvPackage/PackageManager.h>
 #include <QtCore/QVariant>
+#include <QtCore/QDebug>
```
bernie-laberge: I guess we don't want to keep that right ? (It was just for your own debugging correct?)

bernie-laberge:
I like the way the new bottom toolbar implementation behaves, and I was able to validate that it solves the original issue. I would have preferred a more limited blast radius for this change, but I also appreciate the code cleanup. Now my only concern is that this refactoring could potentially break some third-party UI customization. FYI: The RV toolbars are customizable by the users: qtutils.sessionBottomToolBar() (https://aswf-openrv.readthedocs.io/en/latest/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-four.html#pyside-example)
Let's publish an RV 2026 pre-release with it asap once it has been tested and merged, so that we get some feedback from the RV or from the Open RV community.
Thanks for the fix @deltag0 !

bernie-laberge (approved):
I like the way the new bottom toolbar implementation behaves, and I was able to validate that it solves the original issue. I would have preferred a more limited blast radius for this change, but I also appreciate the code cleanup. Now my only concern is that this refactoring could potentially break some third-party UI customization. FYI: The RV toolbars are customizable by the users: qtutils.sessionBottomToolBar() (https://aswf-openrv.readthedocs.io/en/latest/rv-manuals/rv-reference-manual/rv-reference-manual-chapter-four.html#pyside-example)
Let's publish an RV 2026 pre-release with it asap once it has been tested and merged, so that we get some feedback from the RV or from the Open RV community.
Thanks for the fix @deltag0 !

src/plugins/rv-packages/multiple_source_media_rep/multiple_source_media_rep.py
deltag0: We could, but then we'd have to re-define the `play_mode_btn` every time. Not a big cost, but still, let me know your thoughts

src/lib/app/RvCommon/RvBottomViewToolBar.cpp
deltag0: Correct, fixed this

src/plugins/rv-packages/multiple_source_media_rep/multiple_source_media_rep.py
bernie-laberge: Ah! you are right. Forget about this comment then.

bernie-laberge (approved):
LGTM
I much prefer the new implemention: it minimizes the blast radius to the multiple media representation bottom resolution in layout mode. Thank you @deltag0 !

eloisebrosseau (approved):

PR #1278, opened 2026-05-25
fix: SG-43328: Annotation drawing not always updated in Stereo mode

deltag0:
### Summarize your change.

Previously, when annotating in stereo mode with multiple eyes (buffers), the right side would look frozen. The caching logic was sound, but wasn't considering that rendering in 2 passes caused the right side to reset because the cache of the eye rendering second was holding stale data since the code was caching the entire view even when there were 2 eyes.
  
So, now, when we have multiple views we cache only the relevant part to avoid clobbering/stale data.

Not cool that annotations don't update at the same time for each eye.

Mac CY2025

- Annotations on 1 image work
- Annotations in all stereo mode work (including multiple eyes)

- Ensure no regressions for different views

### DIFF
--- a/src/lib/ip/IPCore/DisplayStereoIPNode.cpp
+++ b/src/lib/ip/IPCore/DisplayStereoIPNode.cpp
@@ -36,15 +36,19 @@ namespace IPCore
 
         // Recursively stamps the eye's paint target region on every IPImage in
         // the subtree so the renderer can find it on whichever descendant
-        // actually carries the paint commands.
+        // actually carries the paint commands. (might not be needed)
         void propagatePaintRegion(IPImage* img, float rx, float ry, float rw, float rh)
         {
-            if (!img) return;
+            if (!img)
+                return;
             img->paintTargetRegionX = rx;
             img->paintTargetRegionY = ry;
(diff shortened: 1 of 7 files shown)


src/lib/ip/IPCore/DisplayStereoIPNode.cpp
```diff
@@ -34,6 +34,24 @@ namespace IPCore
     namespace
     {
 
+        // Recursively stamps the eye's paint target region on every IPImage in
+        // the subtree so the renderer can find it on whichever descendant
+        // actually carries the paint commands. (might not be needed)
+        void propagatePaintRegion(IPImage* img, float rx, float ry, float rw, float rh)
```
eloisebrosseau, line 40: You wrote "might not be needed" in the comment above this function, but it's being called in `prepareForStereo`. Is it needed or not to fix the issue?

deltag0: Yeah, I wanted to talk about this. I wasn't sure that when we call `    void ImageRenderer::renderPaint(const IPImage* root, const GLFBO* fbo)`it's the same `IPImage` that when we prepare it for stereo.

PR #1281, opened 2026-05-28
fix: SG-43097: Thumbnail gen psutil refactor

deltag0:
### Summarize your change.

- Remove sending manual signals to start/stop process
- Use psutil to start/stop process

- More modern and standard way to start/stop processes

- Tested on CY2025 on Mac

- Build works correctly on all the OSes


imgui.ini
```diff
@@ -0,0 +1,18 @@
+[Window][WindowOverViewport_11111111]
```
bernie-laberge: This imgui.ini file should not be part of the commit.
Please remove it from this PR: `git rm imgui.ini`
And add  imgui.ini to .gitignore so it doesn't keep coming back.

src/build/make_pyside6.py
```diff
@@ -120,8 +122,9 @@ def get_fallback_clang_filename_suffix(version):
     libclang_zip = os.path.join(TEMP_DIR, "libclang.7z")
 
     # if we have a failed download, clean it up and redownload.
-    # checking for False since it can return None when archive doesn't have a CRC
-    if os.path.exists(libclang_zip) and verify_7z_archive(libclang_zip) is False:
+    # verify_7z_archive returns True on CRC pass, False on CRC fail, None when
+    # there is no CRC or the file isn't a valid 7z (e.g. an HTML 404 page).
+    if os.path.exists(libclang_zip) and verify_7z_archive(libclang_zip) is not True:
```
bernie-laberge: When you get the chance @cedrik-fuoco-adsk , if you could review those two changes ☝️  that would be great. Thanks

src/plugins/rv-packages/session_manager/local_thumbnail_gen.py
```diff
@@ -34,34 +33,16 @@
...
-            if handle:
-                try:
-                    ctypes.windll.ntdll.NtResumeProcess(handle)
-                finally:
-                    ctypes.windll.kernel32.CloseHandle(handle)
-        else:
-            proc.send_signal(signal.SIGCONT)
-    except Exception:
+        psutil.Process(proc.pid).resume()
+    except (psutil.NoSuchProcess, psutil.AccessDenied):
```
bernie-laberge: See above comment

src/plugins/rv-packages/session_manager/local_thumbnail_gen.py
```diff
@@ -34,34 +33,16 @@
...
-            if handle:
-                try:
-                    ctypes.windll.ntdll.NtSuspendProcess(handle)
-                finally:
-                    ctypes.windll.kernel32.CloseHandle(handle)
-        else:
-            proc.send_signal(signal.SIGSTOP)
-    except Exception:
+        psutil.Process(proc.pid).suspend()
+    except (psutil.NoSuchProcess, psutil.AccessDenied):
```
bernie-laberge: According to Claude Code:
🟡 Exception handling slightly narrower than before
  
  Old:
  except Exception:
      logger.warning(...)

  New:
  except (psutil.NoSuchProcess, psutil.AccessDenied):
      logger.warning(...)
      
  A few legitimate failure modes now propagate where they used to be caught:

  - psutil.ZombieProcess — on Unix, if the child exited but hasn't been reaped yet, psutil.Process(pid) can raise this. Semantically "process is gone," exactly the warning case.
  - psutil.TimeoutExpired (in some psutil paths) — won't trigger here for these methods, but worth knowing exists.

  Simplest robust catch:

  except psutil.Error:
      logger.warning(f"Failed to suspend process {proc.pid}")

  psutil.Error is the documented base of all psutil-specific exceptions. Won't suppress unrelated bugs (e.g., a TypeError from a future code change), and won't regress against the old except Exception: for
  any case psutil itself raises

src/build/make_pyside6.py
cedrik-fuoco-adsk: While the change on the conditions looks ok, is there a reason why you are falling back on Clang 19 for Clang 21 when libclang 21 exist? (https://mirrors.ocf.berkeley.edu/qt/development_releases/prebuilt/libclang/) @deltag0

deltag0: I changed this because my build was failing, I forgot to remove it, but it might be worth mentioning that without this:
```
elif major_minor_version_str == "21.0":
                return "19.1.0-based-macos-universal.7z"
```
my build wasn't working.

src/plugins/rv-packages/session_manager/local_thumbnail_gen.py
deltag0: Addressed

src/plugins/rv-packages/session_manager/local_thumbnail_gen.py
deltag0: Addressed

imgui.ini
deltag0: fixed

bernie-laberge (approved):
LGTM
Good job @deltag0 !

PR #1291, opened 2026-06-04
fix: SG-43605: OTIO export failing when OCIO is active, and support reversed-order stack blending

markreidvfx:
### Summarize your change.

Fix OTIO export failing when OCIO is active, and support reversed-order stack blending.

When OCIO is enabled, the `RVLinearize` node is not present and the CDL parameters are instead located on an `RVColor` node. The export hook calls `associatedNode("RVLinearize")`, which returns an empty string, and the subsequent call to `commands.getIntProperty("{}.CDL.active".format(linearize))` then raises an exception.

This fix loops over both `RVLinearize` and `RVColor` nodes, checking `propertyExists` before accessing CDL properties.

Also reverses the input ordering when exporting an RVStack with `supportReversedOrderBlending` enabled, so the OTIO layer order matches RV.

OTIO export throws this exception when OCIO is active:

```
File ".../sourcePostExportHook.py", line 21, in hook_function
    if commands.getIntProperty("{}.CDL.active".format(linearize))[0] != 0:
Exception: "invalid property name .CDL.active", program exception
```

Tested on Rocky Linux 9.

1. Load media with OCIO active
2. Activate a CDL on RVColor:
```python
from rv import commands, extra_commands
source = commands.nodesOfType("RVSource")[0]
color = extra_commands.associatedNode("RVColor", source)
commands.setIntProperty(f"{color}.CDL.active", [1], True)
commands.setFloatProperty(f"{color}.CDL.slope", [1.1, 0.9, 1.0], True)
```
3. File > Export > OTIO — fails with `"invalid property name .CDL.active"`

### DIFF
--- a/src/plugins/rv-packages/otio_reader/otio_writer.py
+++ b/src/plugins/rv-packages/otio_reader/otio_writer.py
@@ -226,9 +226,12 @@ def _create_stack(node_name, *args, **kwargs):
     :return: `otio.schema.Stack`
     """
     stack_node = group_member_of_type(node_name, "RVStack")
-    fps = commands.getFloatProperty(stack_node + ".output.fps")[0]
+    fps = commands.getFloatProperty(f"{stack_node}.output.fps")[0]
+    reverse_order = commands.getIntProperty(f"{stack_node}.mode.supportReversedOrderBlending")
 
     input_node_names, _ = commands.nodeConnections(node_name)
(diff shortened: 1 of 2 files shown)


eloisebrosseau (approved):

bernie-laberge (approved):
LGTM

PR #1292, opened 2026-06-04
feat: SG-43604: Add thread safe consoleWrite command for Python console output

markreidvfx:
### Linked issues

Follow-up to #907

### Summarize your change.

In python, calling print() from a background is still not thread safe and can still lead to rv crashing.

#907 added a thread-safety guard in `PyMuSymbolType::call()` to detect if python is calling a Mu command from a non-main thread. To try and fix the crashing when printing/logging from a python thread, the guard calls `self->function->fullyQualifiedName()` to check if we are calling `extra_commands._print`. If we are, it re-routes the call to a safe print.  Unfortunately it turns out, `fullyQualifiedName()` allocates memory with the Mu garbage collector. This is not thread-safe and can case the very heap corruption it was trying to prevent :(

This PR introduces a dedicated, thread-safe Python `consoleWrite` command instead of relying on Mu’s `print` command. Then it changes any attempt to run Mu from Python on a non-main thread into an unconditional `RuntimeError`. 

To prevent memory corruption if python attempts to run Mu code from a background thread and still allow printing to the RVConsole window. 

Tested on Rocky Linux 9. 

- `PyMuSymbolType.cpp`: 
   - Add unconditional `RuntimeError` for Mu calls from non-main threads. **This is the critical change**.
   - replace `boost::thread` with `std::thread`
- `PyInterface.cpp`: Initialize main thread ID at startup instead of lazily on first call.
- `PyUICommands.cpp`: Add `consoleWrite` command for thread-safe Python console output.
- `rv_commands_setup.py`: Replace `extra_commands._print` with `rv.commands.consoleWrite`.
- `RvConsoleWindow.cpp`: 
   - Add null checks for output streams and handle empty lines.
   - Fix data race in processTextBuffer

N/A

### DIFF
--- a/src/lib/app/PyTwkApp/PyInterface.cpp
+++ b/src/lib/app/PyTwkApp/PyInterface.cpp
@@ -453,6 +453,7 @@ namespace TwkApp
             PyModule_AddObject(pModule, "Event", reinterpret_cast<PyObject*>(pyEventType()));
         }
 
+        initPyMuSymbolType();
         if (PyType_Ready(pyMuSymbolType()) >= 0)
         {
             Py_XINCREF(pyMuSymbolType());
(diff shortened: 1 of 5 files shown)


eloisebrosseau (approved):

cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
LGTM
Thank you @markreidvfx !

PR #1295, opened 2026-06-04
fix: SG-43565: Source preview continues being generated after RV session is cleared

deltag0:
### Summarize your change.
- Added event listener in thumbnail generation script to terminate in flight subprocesses
- Cleared cache on source deletion

- Wasted resources used for something deleted

- Tested on CY2025 on Mac

stop_proc

### DIFF
--- a/src/plugins/rv-packages/session_manager/local_thumbnail_gen.py
+++ b/src/plugins/rv-packages/session_manager/local_thumbnail_gen.py
@@ -634,7 +634,10 @@ class LocalThumbnailGen(rvtypes.MinorMode):
 
         node = event.contents()
 
-        source_node = self._source_node_of_group(node)
+        if commands.nodeType(node) in ("RVFileSource", "RVImageSource"):
+            source_node = node
+        else:
+            source_node = self._source_node_of_group(node)
         if not source_node:
             return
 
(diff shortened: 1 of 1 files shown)


src/plugins/rv-packages/session_manager/local_thumbnail_gen.py
```diff
@@ -586,6 +606,55 @@ def _on_session_deletion(self, event: Any) -> None:
...
+        if not media_path:
+            return
+
+        cache_key = self._cache_key(media_path)
+
+        sources = self._cache_key_to_sources.get(cache_key)
+        if sources is not None:
+            sources.discard(source_node)
+            if sources:
+                return
```
bernie-laberge, line 685: The problem is that when _get_cached_path() is called for a source and that the media path is already in the cache, then we return right away: without registering the source. Note that it is possible that two source nodes share the same media paths. So with your new code, if we have two source nodes pointing to the same media path, one will be in the _cache_key_to_sources’s set and the other not. So if the first source is deleted and not the other, then you will still flush the cache.
Perhaps we should keep things as simple as possible: I propose to only address the Clear Session event. We don’t need the fine granularity of being able to kill the process for a specific media path. Killing all the running processes when there is a Clear Session is good enough in my opinion and will keep things simple. Always best to keep things simple.

deltag0: We talked about this, now _cache_key_to_sources tracks all source nodes

bernie-laberge (approved):
LGTM

PR #1301, opened 2026-06-09
feat: Add annotation shapes, text, input smoothing, and stamp rendering

### LINKED ISSUE #1267: Shared Annotations- Shapes and Text
Collaborate more effectively with shared on-screen annotations in OpenRV. Add shapes, arrows, lines, and text annotations directly on frames, then share them live with teammates during live review sessions.

rogernelson:
### Linked issues
Fixes #1267

### Summarize your change.

This branch replaces the embedded TwkPaint sources with the newly open-sourced standalone [OpenRV-annotation ](https://github.com/AcademySoftwareFoundation/OpenRV-annotation) submodule, which provides TwkPaint::Path, TwkPaint::Smoother, and TwkPaint::StampPath as a self-contained geometry library. We build TwkPaint as before, but now from the submodule sources, and we also add in the new smoother and stamp placer.

compilePenComponent() in PaintIPNode now feeds raw pointer events through TwkPaint::SmoothInterpolate2D incrementally (persistent physics state per stroke). Smoothed output replaces raw points before the geometry build runs, reducing input jitter for stamp-brush types. I left the existing brush types as is for now to preserve parity with older builds, but we could eventually smooth this input too.

New draw primitives in PaintIPNode / PaintCommand:

- LocalRect, LocalEllipse, LocalArrow, LocalLine: rendered via a bounding-box quad and per-type SDF GLSL shaders loaded from the twkpaint-src submodule (assets/shaders/): rectangle.frag  ellipse.frag  arrow.frag  line.frag All share shape_quad.vert and shape_common.glsl for the AA helper. compat_gl21.glsl is prepended at program-build time so the same shader sources work across GLSL versions. The submodule is designed to work this way to keep it renderer-independent.

- LocalText: We swap out the old FTGL library for a more modern Qt font libraries that allow us to add features like changing fonts, bold/italic/underline, etc. QPainter renders into a QImage, uploaded as a GL_RGBA texture and drawn as a textured quad, and a texture cache keyed on (text, fontFamily, fontSize, fontWeight, fontStyle, textDecoration, colour) avoids per-frame re-upload.

OTIO schema definitions added to otio_reader for each shape type: Rectangle.1, Ellipse.1, Arrow.1, Line.1, Text.1. Coordinates use normalised image space consistent with Paint.1. So shapes can be supported in OTIO interchange.

UI changes in the annotate package:
- Shape draw modes (rect, ellipse, arrow, line).
- Shift-lock constrains aspect ratio for rectangle and ellipse.
- Enter key commits in-progress text.
- SVG toolbar icons added for each shape type (Apache-2.0).

The stamp geometry and rendering backend are included to support future brush types. The stamp draw modes are not exposed in the UI in this PR.

TwkPaint::StampPath accumulates StampInstance placements (position, rotation, scale) along a stroke path at arc-length intervals. compilePenComponent() routes strokes whose brushName maps to a stamp brush through StampPath; the resulting stampInstances are stored on LocalPolyLine.

PaintCommand::buildStampQuads() CPU-batches stamp instances into a single GL_QUADS draw call per stroke. Uses VBO layout matching PolyLine::build() (split [all UVs | all positions], tightly packed, no index buffer). Thi is necessary because VAO is not supported in Legacy OpenGL, a comment notes this can be replaced by VAO if we ever move to Core Profile.

BrushTextureManager (IPCore) reads a catalogue.json at first render, uploads brush-tip PNGs as GL_LUMINANCE textures via libpng, and caches BrushInfo (textureId, blendMode, softShader) per brush name. Path resolution: $RV_BRUSH_DIR env var, falling back to Bundle::mainBundle()->top()/assets/brushes. Brush tip assets are staged from the twkpaint-src submodule at build time via CMake FILE(COPY).

TextureReplaceFrag.glsl: samples a GL_LUMINANCE brush-tip texture for alpha; luminance → alpha, uniformColor.rgb drives stroke colour. fwidth(tip) smoothstep gives sub-pixel AA on stamp edges.

BlendMarker mode uses GL_MAX blend to cap per-stamp opacity accumulation within a stroke, with a scratch FBO for compositing to preserve smooth antialiased edges at stroke boundaries.

StampBlendMode enum and associated fields on PolyLine let PaintCommand::execute() select GL blend func and shader from stroke data directly; no brush-name string comparisons remain in the render path.

- TwkGLF/CMakeLists.txt: stamp shader dir updated to twkpaint-src/assets/shaders.
- IPCore/CMakeLists.txt: SDF shape shader dir and brush asset staging path updated to twkpaint-src/assets/shaders and twkpaint-src/assets/brushes respectively.

To upgrade font rendering and new text features, to support stamp-based brush, and to add SDF shape rendering.

Linux Rocky 9, Mac OS 26.5, Windows 11.

I apologize for the size of this. I have been working on it for three months, while awaiting for legal approval to open source some new components on which this depends. Now found at:
https://github.com/AcademySoftwareFoundation/OpenRV-annotation

Screenshot 2026-06-09 at 14 31 03


bernie-laberge (approved):
Very impressive work @rogernelson !
Thank you!

PR #1303, opened 2026-06-10
fix: SG-43710: saveSession(asACopy=true) overwrites the working session filename

markreidvfx:
### Linked issues

N/A

### Summarize your change.

Fix `saveSession(asACopy=true)` so it writes a copy of the session without changing the current session's filename.

Two bugs caused `saveSession(asACopy=true)` to rename the working session to the copy path. First, `RvSession::write()` did not forward `writeAsCopy` as the `saveAs` option, so `Session::write()` defaulted to calling `setFileName()`. Second, `RvSession::write()` unconditionally called `setFileName()` after the write, so even with the correct option the rename would still happen. The fix passes `saveAs=writeAsCopy` in the `WriteRequest` and removes the unconditional `setFileName()`.

Tested on Rocky Linux 9.

- `RvSession.cpp`: Pass `saveAs=writeAsCopy` in the write request; remove unconditional `setFileName()` after write.

N/A

### DIFF
--- a/src/lib/app/RvApp/RvSession.cpp
+++ b/src/lib/app/RvApp/RvSession.cpp
@@ -2126,6 +2126,7 @@ namespace Rv
             newRequest.setOption("sessionProtocol", string("RVSession"));
             newRequest.setOption("sessionProtocolVersion", 4);
             newRequest.setOption("sessionName", string("rv"));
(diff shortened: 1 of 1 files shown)


eloisebrosseau (approved):

bernie-laberge (approved):
LGTM
Thank you @markreidvfx !

PR #1312, opened 2026-06-19
feat: SG-42630: Add crash-dump reporting capability to RV

### LINKED ISSUE #1268: Add crash-dump generation capabilities
Add support to for crash report on **multiple** OSes

Add capability to copy paste stack

bernie-laberge:
### Linked issues

Fixes #1268

### Summarize your change.

Adds minidump-based crash reporting to RV on **macOS, Linux, and Windows**, using **Google Crashpad**
for in-process capture (plus the out-of-process `crashpad_handler`) and **Google Breakpad** for
offline symbolication tooling (`dump_syms`, `minidump_stackwalk`, `minidump_dump`). The two libraries
are kept strictly disjoint. The intended architecture and its invariants are documented normatively in
`docs/crash-reporting.md`.

RV had no way to capture actionable crash reports from end users. This adds automatic minidump capture
with enough context — which Mu/Python script was executing, GPU info, and the attached session log —
to symbolicate and debug a crash offline, without needing a reproduction.

- **macOS (Apple Silicon)** — full local validation: handler init across `rv`/`RV`/`rvio`; a triggered
  `crash()` produces a minidump with the expected annotations (`platform`, `gpu_*`, `py_*`,
  `mu_function`); symbols are stripped from the Release install; the `symbols_archive` zip is produced;
  a real dump symbolicates end-to-end via `symbolicate_crash.sh`; the unit + smoke tests pass under
  `ctest`.
- **Windows (VS 2022, Release)** — validated: the Release install strips all PDBs; `symbols_archive`
  produces `RV-<version>-windows-amd64-symbols.zip` containing the PDBs; `CrashHandlerTest` builds and
  passes.
- **Linux (Rocky)** — the implementation carries Linux-specific support (the `crashpad_handler`
  wrapper with the `ulimit -t` guard, `.sym` generation, install rules); the unit + smoke tests run in
  CI on Linux.

**Crash capture**
- `src/lib/base/TwkUtil/CrashHandler.{h,cpp}` — singleton Crashpad handler: annotation table, log
  attachment, init-order-independent pre-init annotation buffering.
- `src/lib/app/MuTwkApp/CrashHandlerInit.{h,cpp}` — one shared init path for `rv`, the macOS `RV`
  bundle, and `rvio`; per-platform handler wrappers (`src/bin/apps/rv/crashpad_handler_*.sh.in`).

**Script context** — ⚠️ *please review the Mu-core touch points*
- `src/lib/mu/Mu/Mu/ExecutionObserver.h` + hooks in `MachineRep.cpp` / `Thread.cpp`: a
  **dependency-free** execution-observer hook in the Mu core (no TwkUtil/Qt dependency); the
  crash-specific `MuCrashObserver` lives in the app layer (`src/lib/app/MuTwkApp`). Function-level
  fidelity. The `crash()` Mu command (`IPMu/CommandsModule.cpp`) is a **test-only** trigger and carries
  no production logic.

**Symbols & packaging** — ⚠️ *note the OpenRV ↔ pipeline boundary*
- `cmake/macros/rv_generate_symbols.cmake` — build-time `.sym` (and native PDBs on Windows); the
  intermediate `.dSYM` is removed after generation so it does not pollute the staged plugin dirs.
- `cmake/macros/rv_archive_symbols.cmake` (+ `rv_collect_pdbs.cmake`) — the `symbols_archive` target
  builds a versioned, per-platform symbol zip. **OpenRV only produces the archive**; uploading it to a
  symbol store is left to the external build/release pipeline (no external infrastructure assumed).
- `cmake/install/pre_install*.cmake` — strip symbols from the customer (Release) package.
- `src/bin/apps/rv/symbolicate_crash.sh` — portable; adds a `--symbols <dir>` override.

**Tests** — ⚠️ *enabled by default in CI*
- `src/test/CrashHandlerTest` (doctest unit) and `src/test/CrashDumpSmokeTest` (end-to-end). The smoke
  test launches RV with `crash()` and asserts a dump is produced (annotations too where `minidump_dump`
  is available); reviewers may want to confirm the headless launch behaves on their CI runners.

**Docs**
- `docs/crash-reporting.md` (normative/technical spec).
- `docs/rv-manuals/rv-crash-reporting.md` — user-facing guide added to the docs site: how crash reporting works, where dumps are written per platform, and how to disable it (`RV_CRASH_DUMPS_ENABLED=0`).

N/A — this is a build/runtime backend feature with no UI. 

When launching RV, the crash dumps directory is shown:
```
rel (run rvcfg/rvmk) labergb@MTLGJKW517F71 MacOS % ./rv
INFO: real-time thread priorities set (bus=100000kHz)
./rv
Version 4.0.0 (RELEASE), built on Jun 19 2026 at 10:52:14 (HEAD=c5a74aaf).
Copyright Contributors to the Open RV Project
INFO: Crash handler initialized successfully
INFO: Crash dumps will be saved to: /Users/labergb/Library/Logs/ASWF/Crashes/
```

Symbolicating a crash report:
```
% ./symbolicate_crash.sh /Users/labergb/Library/Logs/ASWF/Crashes/pending/4b94d812-9292-4d51-a19d-093c0954560b.dmp 
Symbolicating crash dump...
```

Example of a symbolicated crash 
```
Operating system: Mac OS X
                  26.5.1 25F80
CPU: arm64
     10 CPUs

GPU: Apple Apple M1 Pro

Crash reason:  EXC_BAD_ACCESS / KERN_INVALID_ADDRESS
Crash address: 0x0
Process uptime: 28 seconds

Thread 0 (crashed)
 0  rv!IPMu::crash(Mu::Node const&, Mu::Thread&) + 0x8
     x0 = 0x0000000116704508    x1 = 0x0000000114cbce00
     x2 = 0x0000000114cbce00    x3 = 0x0000000130e5ed11
     x4 = 0x0000000000000000    x5 = 0x0000000000000000
     x6 = 0x0000000000000000    x7 = 0xfffff0003ffff800
...

========================================
Crashpad Annotations
========================================

gpu_renderer: Apple M1 Pro
gpu_vendor: Apple
python_caller: 
py_function: 
py_script_line: 
py_script_file: 
mu_script_file: 
mu_function: rvui.clearEverything

Simple Annotations:
app_name: RV
app_version: 4.0.0
cpu_count: 10
os_kernel: 25.5.0
os_version: 26.5
platform: macOS
qt_version: 6.5.3
```


docs/crash-reporting.md
```diff
@@ -0,0 +1,289 @@
+# Crash Reporting in RV
```
eloisebrosseau, line 1: I'm not to sure this is the right place for this kind of file. All the documentation is in a specific folder, so at the very least, I don't think this should be at the root of the docs directory. Nonetheless, I don't think we have that kind of AI rule file anywhere else in the repo. Is this something you want to start doing?

bernie-laberge: Having a spec is the key of having consistant changes done by AI. My original implementation did not have such a specification document and sometimes an AI would make a change that was breaking something else. 

But I agree it is not the best place for such a document. We are about to switch to OpenSpec for AI assisted development. I will convert this spec to OpenSpec once we do.

src/lib/app/MuTwkApp/MuCrashObserver.cpp
```diff
@@ -0,0 +1,87 @@
...
+    {
+        static MuCrashObserver theObserver;
+        return theObserver;
+    }
+
+    MuCrashObserver::MuCrashObserver()
+        : m_lastWrittenFn(nullptr)
+        , m_mainThread(pthread_self()) // constructed during install on the main thread
+    {
+        m_stack.reserve(64);
```
eloisebrosseau: Shouldn't this magic number be replaced with a named constant?

src/lib/app/MuTwkApp/MuTwkApp/MuCrashObserver.h
```diff
@@ -0,0 +1,66 @@
...
+    //  race-free it only mutates the (process-global) crash annotations on the
+    //  main thread; per-call work is a pointer compare plus, when the function
+    //  changes, one name lookup.
+    //
+    class MuCrashObserver : public Mu::ExecutionObserver
+    {
+    public:
+        static MuCrashObserver& instance();
+
+        void onEnter(const Mu::Function* f, const Mu::Node* callNode, Mu::Thread& t) override;
```
eloisebrosseau: I know there is a lot of files where parameters are a single letter, but since this is a new file, could we try to have names that are more descriptive (i.e. `function` instead of `f` and `thread` instead of `t`)? It would really help with the readability!

src/lib/app/PyTwkApp/PyInterface.cpp
```diff
@@ -306,6 +354,19 @@ namespace TwkApp
...
+        //  Install the Python trace function for crash-dump context, but only
+        //  when crash reporting is active. Installing a trace hook forces CPython
+        //  off its fast per-line eval path, so gating this avoids imposing that
+        //  overhead when crash dumps are disabled (RV_CRASH_DUMPS_ENABLED=0) or
+        //  the handler failed to initialize.
+        //
+        if (TwkUtil::CrashHandler::instance().isInitialized())
+        {
+            PyLockObject locker;
+            PyEval_SetTrace(pythonTraceFunction, NULL);
```
eloisebrosseau: You should replace `NULL` with `nullptr`

src/lib/app/PyTwkApp/PyInterface.cpp
```diff
@@ -248,6 +249,53 @@ namespace TwkApp
...
+                Py_DECREF(code);
+            }
+        }
+        else if (what == PyTrace_RETURN && TwkUtil::CrashHandler::instance().isInitialized())
+        {
+            // When the outermost Python frame returns (no caller frame), control
+            // is leaving Python for C++. Clear the Python context so a later crash
+            // in C++ is not misattributed to a Python line that already returned.
+            PyFrameObject* back = PyFrame_GetBack(frame);
+            if (back == NULL)
```
eloisebrosseau: You should replace `NULL` with `nullptr`

src/lib/app/PyTwkApp/PyMuSymbolType.cpp
```diff
@@ -230,7 +231,7 @@ namespace TwkApp
         PyLockObject locker;
         PyMuSymbolObject* self;
 
-        if (self = (PyMuSymbolObject*)type->tp_alloc(type, 0))
+        if ((self = (PyMuSymbolObject*)type->tp_alloc(type, 0)))
```
eloisebrosseau: I know you just added parentheses around this if statement, but shouldn't we moved the assigment outside of the statement?

src/lib/base/TwkUtil/CrashHandler.cpp
```diff
@@ -0,0 +1,523 @@
+//******************************************************************************
+// Copyright (c) 2026 Autodesk Inc. All rights reserved.
+//
+// SPDX-License-Identifier: Apache-2.0
+//
+//******************************************************************************
+
+#include <TwkUtil/CrashHandler.h>
+#include "CrashHandler_internal.h"
```
eloisebrosseau, line 11: I thought the convention was to always put include files inside `<>` or is it different for this one?

src/lib/base/TwkUtil/CrashHandler.cpp
```diff
@@ -0,0 +1,523 @@
...
+    std::string CrashHandler::getCrashDumpDirectory() const { return m_impl->getCrashDumpDirectory(); }
+
+    //
+    // Platform-specific implementation using Crashpad
+    //
+
+    // Platform-specific configuration
+#if defined(PLATFORM_DARWIN)
+#define CRASH_HANDLER_IMPL_CLASS CrashHandlerDarwinImpl
+#define PLATFORM_STRING "macOS"
```
eloisebrosseau: Does this needs to be a macro? Otherwise, I believe a constexpr constant might be more appropriate

src/lib/base/TwkUtil/CrashHandler.cpp
```diff
@@ -0,0 +1,523 @@
...
+        // Map of annotation names to their Crashpad annotation objects
+        struct AnnotationMapping
+        {
+            const char* name;
+            crashpad::Annotation* annotation;
+            char* buffer;
+            size_t buffer_size;
+        };
+
+        AnnotationMapping g_annotationMappings[] = {
```
eloisebrosseau: This c-style array should be an std::array instead

src/lib/base/TwkUtil/CrashHandler.cpp
```diff
@@ -0,0 +1,523 @@
...
+    //
+    namespace
+    {
+        // Buffer size for annotations (max 1024 chars each)
+        constexpr size_t kAnnotationBufferSize = 1024;
+
+        // Static annotation buffers - must have static storage duration
+        // Mu script annotations (function-level: the Mu crash observer records
+        // the function and its file, not a source line - see MuCrashObserver)
+        char g_muScriptFileBuffer[kAnnotationBufferSize];
```
eloisebrosseau: I believe all the below c-style arrays could be std::arrays instead

src/lib/base/TwkUtil/CrashHandler.cpp
```diff
@@ -0,0 +1,523 @@
...
+
+    std::string CrashHandler::getCrashDumpDirectory() const { return m_impl->getCrashDumpDirectory(); }
+
+    //
+    // Platform-specific implementation using Crashpad
+    //
+
+    // Platform-specific configuration
+#if defined(PLATFORM_DARWIN)
+#define CRASH_HANDLER_IMPL_CLASS CrashHandlerDarwinImpl
```
eloisebrosseau: Out of curiosity, why do we need a macro to define the platform implementation if it's all in a unified class?

src/lib/base/TwkUtil/CrashHandler.cpp
```diff
@@ -0,0 +1,523 @@
...
+        // solely by their live crashpad::Annotation, so they must not also be
+        // written to m_annotations (that duplicated them as stale, usually-empty
+        // entries in the dump's "Simple Annotations").
+        bool matched = false;
+        for (const auto& mapping : g_annotationMappings)
+        {
+            if (key == mapping.name)
+            {
+                // Copy value to the annotation buffer
+                size_t len = std::min(value.length(), mapping.buffer_size - 1);
```
eloisebrosseau: From what I see, it looks like you could replace the c-style char arrays above with `crashpad::StringAnnotation` (and ignore my comment about making them std::arrays) instead of manually handling the annotation buffers with memcpy. Is this something you have looked into? What do you think? https://crashpad.chromium.org/doxygen/classcrashpad_1_1StringAnnotation.html

src/lib/app/MuTwkApp/MuTwkApp/MuCrashObserver.h
bernie-laberge: Good point ! On it

src/lib/app/MuTwkApp/MuCrashObserver.cpp
bernie-laberge: Absolutely. OK done.

src/lib/app/MuTwkApp/MuTwkApp/MuCrashObserver.h
bernie-laberge: OK done.

src/lib/app/PyTwkApp/PyInterface.cpp
bernie-laberge: OK done.

src/lib/app/PyTwkApp/PyInterface.cpp
bernie-laberge: OK done.

src/lib/app/PyTwkApp/PyMuSymbolType.cpp
bernie-laberge: Yes we should. Plus we return it so it is best to move the assignment outside. You are right. Thanks.
OK done.

src/lib/base/TwkUtil/CrashHandler.cpp
bernie-laberge: OK done.

src/lib/base/TwkUtil/CrashHandler.cpp
bernie-laberge: True. OK done.

src/lib/base/TwkUtil/CrashHandler.cpp
bernie-laberge: OK done

src/lib/base/TwkUtil/CrashHandler.cpp
bernie-laberge: Ah ! At first, Claude had created 3 separated pretty much identical implementation classes - one per platform - which I made him refactor to a common one. But the name remained but you are right, it is not needed. 
OK done: No longer using the CRASH_HANDLER_IMPL_CLASS macro. Using CrashHandlerPlatformImpl instead.

src/lib/base/TwkUtil/CrashHandler.cpp
bernie-laberge: Excellent point!
OK done.

src/lib/base/TwkUtil/CrashHandler.cpp
bernie-laberge: True but this is a private header that is not exposed publicly. 
I will add a comment.

eloisebrosseau (approved):

cedrik-fuoco-adsk (approved):

PR #1315, opened 2026-06-23
fix: hang when movie plugins invoke python

rogernelson:
### Summarize your change.
If a plugin in the MovieIO plugin folder invokes Python in its initialization, a deadlock will occur.  This is because the main thread is holding the GIL (global interpreter lock) when the plugins are initialized (in their own threads). If any of these plugins uses Python it will wait on the GIL held by the main thread, which is waiting on its preloader threads to terminate so deadlock will occur.

The cleanest fix for this is to pass callbacks to MovieIO to avoid introducing a Python dependency in it. We'll add one callback to release any resources (optional) and another to re-acquire them (optional). Then we'll call these before and after our blocking wait on the preloader if they are set.

We'll set this in RvApp where we already initialize a bunch of other statics in RvSession. We'll use this to release the GIL (if we are holding it) and reacquire it (if we were holding it before). We'll use a `thread_local` variable for the GIL state to be correct, even though this shouldn't really ever be called outside of the main thread. 

To avoid deadlock when loading plugins.

Mac OS 26.5.1

### DIFF
--- a/src/lib/app/RvApp/RvSession.cpp
+++ b/src/lib/app/RvApp/RvSession.cpp
@@ -6,6 +6,7 @@
 //
 //
 #include <RvApp/CommandsModule.h>
+#include <Python.h>
 #include <RvApp/Options.h>
 #include <RvApp/RvNodeDefinitions.h>
 #include <RvApp/RvSession.h>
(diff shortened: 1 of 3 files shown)


bernie-laberge (approved):
LGTM
Thank you @rogernelson !

cedrik-fuoco-adsk (approved):

bernie-laberge:
It was decided to close this PR for now because it is not currently needed. But we might revive it if needs be in the future.

PR #1316, opened 2026-06-25
fix: SG-43388: RV leaks memory because of opened decoders

deltag0:
### Summarize your change.
- Re-enable old context map that manages the number of open threads to decode videos

- Without the context map, the number of open decoder threads keeps increasing as we load more videos.
- RAM usage would increase a lot even by loading video clips without playing the video at all, and just staying on the first frame.

Tested on CY 2025 on Rocky Linux 9.8 and Mac

Opened over 1000 clips with 1GB of caching, peaking at 9GB, but staying bounded.
RAM no longer increases when loading clips without playback being on.

- After freeing the decoder with `avcodec_free_context`, nullify the tracked pointer to not point to freed memory
No regressions by adding the context map

Screenshot from 2026-06-25 08-44-08
Screenshot from 2026-07-02 08-17-49

### DIFF
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -446,7 +446,7 @@ namespace TwkMovie
     //  Global pool object:
     //
 
-    ContextPool* globalContextPool = 0;
+    std::unique_ptr<ContextPool> globalContextPool{nullptr};
 
     void ContextPool::flushContext(MovieFFMpegReader* reader, int streamIndex)
     {
@@ -5959,7 +5959,7 @@ namespace TwkMovie
                 poolSize = atoi(c);
             }
             if (poolSize > 0)
-                globalContextPool = new ContextPool(poolSize);
+                globalContextPool = std::make_unique<ContextPool>(poolSize);
             else
             {
                 report("Disabling mio_ffmpeg context thread pool.", true);
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -5958,7 +5958,8 @@ namespace TwkMovie
             {
                 poolSize = atoi(c);
             }
-            if (poolSize > 0) globalContextPool = new ContextPool(poolSize);
+            if (poolSize > 0)
+                globalContextPool = new ContextPool(poolSize);
             else
             {
                 report("Disabling mio_ffmpeg context thread pool.", true);
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -539,9 +539,15 @@ namespace TwkMovie
                 avcodec_free_context(&closeContext.avContext);
 
                 if (closeContext.vTrack)
+                {
+                    closeContext.vTrack->avCodecContext = nullptr;
                     closeContext.vTrack->isOpen = false;
+                }
                 if (closeContext.aTrack)
+                {
+                    closeContext.aTrack->avCodecContext = nullptr;
                     closeContext.aTrack->isOpen = false;
+                }
             }
         }
     }
@@ -1154,11 +1160,11 @@ namespace TwkMovie
         for (unsigned int i = 0; i < m_audioTracks.size(); i++)
         {
             AudioTrack* track = m_audioTracks[i];
+            ContextPool::flushContext(this, track->number);
             if (track->isOpen)
             {
                 avcodec_free_context(&track->avCodecContext);
             }
-            ContextPool::flushContext(this, track->number);
             delete track;
         }
         m_audioTracks.resize(0);
@@ -1166,11 +1172,11 @@ namespace TwkMovie
         for (unsigned int i = 0; i < m_videoTracks.size(); i++)
         {
             VideoTrack* track = m_videoTracks[i];
+            ContextPool::flushContext(this, track->number);
             if (track->isOpen)
             {
                 avcodec_free_context(&track->avCodecContext);
             }
-            ContextPool::flushContext(this, track->number);
             delete track;
         }
         m_videoTracks.resize(0);
@@ -2741,7 +2747,8 @@ namespace TwkMovie
             audioChannels = idAudioChannels(layout, numChannels);
 
             // XXX Assume this thread will never decode audio
-            avcodec_free_context(&audioCodecContext);
+            ContextPool::flushContext(this, track->number);
+            avcodec_free_context(&track->avCodecContext);
             track->isOpen = false;
         }
 
@@ -5944,28 +5951,19 @@ namespace TwkMovie
             addType(formatsItr->first, formatsItr->second.first, formatsItr->second.second, video, audio, separams, sdparams);
         }
 
-        // Note : No longer using the global context pool (since FFmpeg 6.0)
-        // Rationale: The global context pool was based on the premise that a
-        // context could be opened and closed multiple times. However, with
-        // FFmpeg 6.0, this premise is no longer valid and was causing crashes.
-        // As per the FFmpeg 6 documentation:
-        // https://ffmpeg.org/doxygen/trunk/deprecated.html: "Opening and
-        // closing a codec context multiple times is not supported anymore – use
-        // multiple codec contexts instead."
-
-        // if (!globalContextPool)
-        // {
-        //     int poolSize = 500;
-        //     if (const char* c = getenv("TWK_MOVIEFFMPEG_CONTEXT_POOL_SIZE"))
-        //     {
-        //         int poolSize = atoi(c);
-        //     }
-        //     if (poolSize > 0) globalContextPool = new ContextPool(poolSize);
-        //     else
-        //     {
-        //         report("Disabling mio_ffmpeg context thread pool.", true);
-        //     }
-        // }
+        if (!globalContextPool)
+        {
+            int poolSize = 500;
+            if (const char* c = getenv("TWK_MOVIEFFMPEG_CONTEXT_POOL_SIZE"))
+            {
+                poolSize = atoi(c);
+            }
+            if (poolSize > 0) globalContextPool = new ContextPool(poolSize);
+            else
+            {
+                report("Disabling mio_ffmpeg context thread pool.", true);
+            }
+        }
     }
 
     MovieFFMpegIO::~MovieFFMpegIO()
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -2725,7 +2725,6 @@ namespace TwkMovie
             AVChannelLayout layout = audioCodecContext->ch_layout;
             audioChannels = idAudioChannels(layout, numChannels);
 
-            // XXX Assume this thread will never decode audio
             ContextPool::flushContext(this, track->number);
             avcodec_free_context(&track->avCodecContext);
             track->isOpen = false;
--- a/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
+++ b/src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
@@ -20,7 +20,6 @@
 #include <TwkUtil/PathConform.h>
 #include <TwkUtil/File.h>
 #include <TwkUtil/sgcHop.h>
-#include <cstddef>
 #include <iostream>
 #include <iomanip>
 #include <algorithm>
@@ -413,7 +412,7 @@ namespace TwkMovie
         class Reservation
         {
         public:
-            Reservation(MovieFFMpegReader* reader, int streamIndex, VideoTrack* vTrack=nullptr, AudioTrack* aTrack=nullptr);
+            Reservation(MovieFFMpegReader* reader, int streamIndex);
             ~Reservation();
 
         private:
@@ -474,7 +473,7 @@ namespace TwkMovie
         gcp.m_contextMap.erase(i);
     }
 
-    ContextPool::Reservation::Reservation(MovieFFMpegReader* reader, int streamIndex, VideoTrack* vTrack, AudioTrack* aTrack)
+    ContextPool::Reservation::Reservation(MovieFFMpegReader* reader, int streamIndex)
         : m_context(0)
         , m_dbline(0)
         , m_dblline(0)
@@ -497,13 +496,6 @@ namespace TwkMovie
         context.reserved = true;
         context.reader = reader;
         context.streamIndex = streamIndex;
-
-        if (vTrack) {
-            context.vTrack = vTrack;
-        }
-        if (aTrack) {
-            context.aTrack = aTrack;
-        }
 
         //
         //  If it's already in the list move it to the front now.
@@ -592,6 +584,7 @@ namespace TwkMovie
         {
             AVStream* avStream = context.reader->m_avFormatContext->streams[context.streamIndex];
 
+            context.reader->trackFromStreamIndex(context.streamIndex, context.vTrack, context.aTrack);
             if (context.vTrack)
             {
                 context.avContext = context.vTrack->avCodecContext;
@@ -2323,14 +2316,15 @@ namespace TwkMovie
                 if (heroVideoTracks[i])
                 {
                     VideoTrack* track = new VideoTrack;
-                    ContextPool::Reservation reserve(this, i, track);
+                    ContextPool::Reservation reserve(this, i);
                     track->useOpenJPH = isJ2K;
 #if defined(RV_USE_APPLE_PRORES_SDK)
                     track->useAppleProRes = isProRes;
 #endif
                     if (isJ2K)
                     {
-                        if (openAVCodec(i, &track->avCodecContext, &track->hardwareContext)) {
+                        if (openAVCodec(i, &track->avCodecContext, &track->hardwareContext))
+                        {
                             track->number = i;
                             ostringstream trackName;
                             trackName << "track " << m_videoTracks.size() + 1;
@@ -2338,7 +2332,8 @@ namespace TwkMovie
                             track->isOpen = true;
                             m_videoTracks.push_back(track);
                         }
-                        else {
+                        else
+                        {
                             delete track;
                         }
                     }
@@ -2450,7 +2445,7 @@ namespace TwkMovie
                 if (heroAudioTracks[i])
                 {
                     AudioTrack* track = new AudioTrack;
-                    ContextPool::Reservation reserve(this, i, nullptr, track);
+                    ContextPool::Reservation reserve(this, i);
                     if (openAVCodec(i, &track->avCodecContext))
                     {
                         track->number = i;
@@ -3220,7 +3215,7 @@ namespace TwkMovie
             if (start < 0)
                 continue;
             AudioTrack* track = m_audioTracks[i];
-            ContextPool::Reservation reserve(this, track->number, nullptr, track);
+            ContextPool::Reservation reserve(this, track->number);
             track->isOpen = openAVCodec(track->number, &track->avCodecContext);
             if (!track->isOpen)
             {
@@ -4096,7 +4091,7 @@ namespace TwkMovie
         for (unsigned int i = 0; i < m_videoTracks.size(); i++)
         {
             VideoTrack* track = m_videoTracks[i];
-            ContextPool::Reservation reserve(this, track->number, track);
+            ContextPool::Reservation reserve(this, track->number);
             track->isOpen = openAVCodec(track->number, &track->avCodecContext, &track->hardwareContext);
             if (!track->isOpen)
             {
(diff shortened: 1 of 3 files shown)


src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -2741,7 +2747,8 @@ namespace TwkMovie
             audioChannels = idAudioChannels(layout, numChannels);
 
             // XXX Assume this thread will never decode audio
```
bernie-laberge: We should change this comment since it no longer makes sense with your fix. Our robot friend recommends this:
// The context was opened only to read the audio stream parameters above.
// Drop it from the pool before freeing so the pool does not retain a
// dangling AVCodecContext* and its open-thread count stays accurate.
// Decoding clones reopen their own context on demand in audioFillBuffer().

bernie-laberge:
FYI @deltag0 : From our robot friend:
Blocking issue

Use-after-free in ~IPGraph() — job dispatcher destroyed before the nodes that use it

IPGraph.cpp:325-330 deletes the JobDispatcher at the top of the destructor and does not null m_jobDispatcher. The new node-deletion loop at IPGraph.cpp:345-355 then runs afterward.

FileSourceIPNode::~FileSourceIPNode() calls cancelJobs() (FileSourceIPNode.cpp:322), which does:

int workItemID = m_workItemID.exchange(0);
if (workItemID)
{
    graph()->removeWorkItem(workItemID);   // FileSourceIPNode.cpp:2405
    graph()->waitWorkItem(workItemID);     // FileSourceIPNode.cpp:2408
}

and both unconditionally dereference the freed dispatcher:

auto jobDispatcher = reinterpret_cast<JobDispatcher*>(m_jobDispatcher);
jobDispatcher->removeJob(id);              // IPGraph.cpp:3775-3776

Failure scenario: quit RV while any media-load work item is still outstanding (m_workItemID != 0) — e.g. quitting during or shortly after loading sources. ~IPGraph frees the dispatcher, then deletes the FileSourceIPNode, which calls removeJob()/waitJob() on freed memory. waitJob() blocking on a destroyed condition variable is the nastier of the two.

It's intermittent — m_workItemID is usually 0 by shutdown, which is why manual testing wouldn't catch it — but it's a genuine crash-on-exit regression, and this PR is the thing that made those destructors run at all.

Fix: move the node-deletion block before the dispatcher teardown, and defensively null the pointer:

IPGraph::~IPGraph()
{
    setCachingMode(NeverCache, 1, 2, 1, 2, 1, 1, 24.0);
    finishCachingThread();
    finishAudioThread();

    //  Delete all nodes first: source-node destructors call back into the
    //  graph (FileSourceIPNode::cancelJobs -> removeWorkItem/waitWorkItem),
    //  so the job dispatcher must still be alive here.
    for (auto& [name, node] : m_viewNodeMap)
    {
        node->willDelete();
        node->disconnectInputs();
    }

    while (!m_viewNodeMap.empty())
    {
        delete m_viewNodeMap.begin()->second;
    }

    m_nodeMap.clear();

    if (m_jobDispatcher)
    {
        auto jobDispatcher = reinterpret_cast<JobDispatcher*>(m_jobDispatcher);
        jobDispatcher->stop();
        delete jobDispatcher;
        m_jobDispatcher = nullptr;
    }

    pthread_mutex_destroy(&m_internalLock);
    ...
}

Separately, addWorkItem/removeWorkItem/waitWorkItem (IPGraph.cpp:3766-3792) should null-checnow prioritizeWorkItem is the only one that guards anything, and it guards id, not thedispatcher.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -446,7 +446,7 @@ namespace TwkMovie
     //  Global pool object:
     //
 
-    ContextPool* globalContextPool = 0;
+    std::unique_ptr<ContextPool> globalContextPool{nullptr};
```
bernie-laberge: 1. std::unique_ptr for globalContextPool is a regression, not an improvement (MovieFFMpeg.cpp:449)

The old ContextPool* globalContextPool = 0; was an intentionally-leaked singleton — no destruction, no teardown ordering to reason about. Making it a unique_ptr adds a static destructor, and:

- ~unique_ptr calls delete ptr but does not null the stored pointer. So the if (!globalContextPool) return; guards in flushContext (:453), Reservation::Reservation (:481) and ~Reservation (:557) all still pass after the pool is destroyed, and every one of them then does LockGuard lock(gcp.m_mutex) on a destroyed boost::mutex.
- Static destruction order across the MovieFFMpeg plugin vs. any still-running decode thread or later-destroyed MovieFFMpegReader is unspecified.

Note that MovieFFMpegIO::~MovieFFMpegIO() (:5970) still carries the // XXX delete context pool ? comment — the ownership question is unresolved, and unique_ptr answers it with the riskiest option. I'd revert to the raw pointer

bernie-laberge:
Other issues:

Other issues

1. std::unique_ptr for globalContextPool is a regression, not an improvement (MovieFFMpeg.cpp:449)

The old ContextPool* globalContextPool = 0; was an intentionally-leaked singleton — no destruction, no teardown ordering to reason about. Making it a unique_ptr adds a static destructor, and:

- ~unique_ptr calls delete ptr but does not null the stored pointer. So the if (!globalContextPool) return; guards in flushContext (:453), Reservation::Reservation (:481) and ~Reservation (:557) all still pass after the pool is destroyed, and every one of them then does LockGuard lock(gcp.m_mutex) on a destroyed boost::mutex.
- Static destruction order across the MovieFFMpeg plugin vs. any still-running decode thread or later-destroyed MovieFFMpegReader is unspecified.

Note that MovieFFMpegIO::~MovieFFMpegIO() (:5970) still carries the // XXX delete context pool ? comment — the ownership question is unresolved, and unique_ptr answers it with the riskiest option. I'd revert to the raw pointer, or keep the unique_ptr and explicitly globalContextPool.reset() at a controlled shutdown point so the guards actually work.

2. Re-enabled Reservation ctor can walk off an empty list (MovieFFMpeg.cpp:516-552)

Pre-existing code, but this PR is what makes it live again:

while (gcp.m_currentOpenThreads >= gcp.m_maxOpenThreads)
{
    Context& closeContext = *gcp.m_openContexts.back();   // UB if list is empty
    gcp.m_openContexts.pop_back();
    ...
    if (closeContext.reserved) { cout << "ERROR: ..."; }  // no decrement -> loop spins
    else if (closeContext.avContext) { /* decrement, free */ }
}

If enough contexts are simultaneously reserved that their combined thread_count reaches m_maxOpenThreads, the loop pops entries without ever decrementing the counter, drains m_openContexts, and then calls .back() on an empty list. With the default 500-thread budget and typical thread_count of 8-16 that needs ~30-60 concurrently-reserved contexts — reachable under heavy multi-source playback, which is precisely the workload this PR targets. The XXX Should never happen comment at :526 is the author of the original code flagging the same doubt.

Minimum fix: if (gcp.m_openContexts.empty()) break; at the top of the loop body.

3. Deleting the FFmpeg 6.0 rationale comment

The removed block (:5947 in the base) is institutional knowledge — it records why the pool was disabled and cites the FFmpeg docs. Since the PR resolves that concern rather than refuting it, replace it rather than delete it:

//  The global context pool caps the number of simultaneously-open decoder
//  threads. It was disabled during the FFmpeg 6.0 migration because a context
//  cannot be reopened after being closed
//  (https://ffmpeg.org/doxygen/trunk/deprecated.html). That is now handled:
//  when the pool evicts a context it frees it and nulls the owning track's
//  avCodecContext, so openAVCodec() allocates a fresh context on next use.

4. Reservation::~Reservation writes through a potentially-freed pointer (MovieFFMpeg.cpp:564-574)

Context& context = *m_context;
context.reserved = false;                                  // write happens first
ContextKey key(context.reader, context.streamIndex);       // reads from the same object
if (gcp.m_contextMap.find(key) == gcp.m_contextMap.end())
    return;                                                // ...check happens after

flushContext() erases the map node (:473), which destroys the Context. If that races with a live Reservation, the dtor writes to and reads from freed memory before the liveness check that was clearly meant to prevent exactly this. The check is unsalvageable as written since it derives the key from the dead object — the Reservation should cache reader/streamIndex at construction and look up by that.

I traced the two flushContext call sites in this PR and neither currently races: the Reservation at :2450 is scoped to its case block and destroyed before initializeAudio() runs at :2550, and close() shouldn't overlap a decode on the same reader. So this is latent, not active — but it's fragile enough to be worth a comment or a proper fix while the pool is being brought back.

5. Style — magic number (MovieFFMpeg.cpp:5956)

int poolSize = 500; — per the project's C++ conventions, non-obvious literals should be named:

constexpr int kDefaultContextPoolSize = 500;

Also worth documenting the units in a comment: it's a count of decoder threads, not contexts, which isn't obvious from the name.

deltag0:
I took a look at the code again to refresh myself, and I found that the `m_subtitleMap` looks like an unused field, so we could delete it.

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
deltag0: Fixed, do we need a justification comment, we do the free and flush pattern in other places with no comments

deltag0:
> FYI @deltag0 : From our robot friend: Blocking issue
> 
> Use-after-free in ~IPGraph() — job dispatcher destroyed before the nodes that use it
> 
> IPGraph.cpp:325-330 deletes the JobDispatcher at the top of the destructor and does not null m_jobDispatcher. The new node-deletion loop at IPGraph.cpp:345-355 then runs afterward.
> 
> FileSourceIPNode::~FileSourceIPNode() calls cancelJobs() (FileSourceIPNode.cpp:322), which does:
> 
> int workItemID = m_workItemID.exchange(0); if (workItemID) { graph()->removeWorkItem(workItemID); // FileSourceIPNode.cpp:2405 graph()->waitWorkItem(workItemID); // FileSourceIPNode.cpp:2408 }
> 
> and both unconditionally dereference the freed dispatcher:
> 
> auto jobDispatcher = reinterpret_cast<JobDispatcher*>(m_jobDispatcher); jobDispatcher->removeJob(id); // IPGraph.cpp:3775-3776
> 
> Failure scenario: quit RV while any media-load work item is still outstanding (m_workItemID != 0) — e.g. quitting during or shortly after loading sources. ~IPGraph frees the dispatcher, then deletes the FileSourceIPNode, which calls removeJob()/waitJob() on freed memory. waitJob() blocking on a destroyed condition variable is the nastier of the two.
> 
> It's intermittent — m_workItemID is usually 0 by shutdown, which is why manual testing wouldn't catch it — but it's a genuine crash-on-exit regression, and this PR is the thing that made those destructors run at all.
> 
> Fix: move the node-deletion block before the dispatcher teardown, and defensively null the pointer:
> 
> IPGraph::~IPGraph() { setCachingMode(NeverCache, 1, 2, 1, 2, 1, 1, 24.0); finishCachingThread(); finishAudioThread();
> 
> ```
> //  Delete all nodes first: source-node destructors call back into the
> //  graph (FileSourceIPNode::cancelJobs -> removeWorkItem/waitWorkItem),
> //  so the job dispatcher must still be alive here.
> for (auto& [name, node] : m_viewNodeMap)
> {
>     node->willDelete();
>     node->disconnectInputs();
> }
> 
> while (!m_viewNodeMap.empty())
> {
>     delete m_viewNodeMap.begin()->second;
> }
> 
> m_nodeMap.clear();
> 
> if (m_jobDispatcher)
> {
>     auto jobDispatcher = reinterpret_cast<JobDispatcher*>(m_jobDispatcher);
>     jobDispatcher->stop();
>     delete jobDispatcher;
>     m_jobDispatcher = nullptr;
> }
> 
> pthread_mutex_destroy(&m_internalLock);
> ...
> ```
> 
> }
> 
> Separately, addWorkItem/removeWorkItem/waitWorkItem (IPGraph.cpp:3766-3792) should null-checnow prioritizeWorkItem is the only one that guards anything, and it guards id, not thedispatcher.

Fixed & double checked the use of the addition :D

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
deltag0: I'm not sure if how this could happen (global variable being uninitialized and then another function using it). Unless there's mismanaged threads, so if we want to be 100% sure and take a small memory leak, it's good to revert.

deltag0:
> Other issues:
> 
> Other issues
> 
> 1. std::unique_ptr for globalContextPool is a regression, not an improvement (MovieFFMpeg.cpp:449)
> 
> The old ContextPool* globalContextPool = 0; was an intentionally-leaked singleton — no destruction, no teardown ordering to reason about. Making it a unique_ptr adds a static destructor, and:
> 
> * ~unique_ptr calls delete ptr but does not null the stored pointer. So the if (!globalContextPool) return; guards in flushContext (:453), Reservation::Reservation (:481) and ~Reservation (:557) all still pass after the pool is destroyed, and every one of them then does LockGuard lock(gcp.m_mutex) on a destroyed boost::mutex.
> * Static destruction order across the MovieFFMpeg plugin vs. any still-running decode thread or later-destroyed MovieFFMpegReader is unspecified.
> 
> Note that MovieFFMpegIO::~MovieFFMpegIO() (:5970) still carries the // XXX delete context pool ? comment — the ownership question is unresolved, and unique_ptr answers it with the riskiest option. I'd revert to the raw pointer, or keep the unique_ptr and explicitly globalContextPool.reset() at a controlled shutdown point so the guards actually work.
> 
> 2. Re-enabled Reservation ctor can walk off an empty list (MovieFFMpeg.cpp:516-552)
> 
> Pre-existing code, but this PR is what makes it live again:
> 
> while (gcp.m_currentOpenThreads >= gcp.m_maxOpenThreads) { Context& closeContext = _gcp.m_openContexts.back(); // UB if list is empty gcp.m_openContexts.pop_back(); ... if (closeContext.reserved) { cout << "ERROR: ..."; } // no decrement -> loop spins else if (closeContext.avContext) { /_ decrement, free */ } }
> 
> If enough contexts are simultaneously reserved that their combined thread_count reaches m_maxOpenThreads, the loop pops entries without ever decrementing the counter, drains m_openContexts, and then calls .back() on an empty list. With the default 500-thread budget and typical thread_count of 8-16 that needs ~30-60 concurrently-reserved contexts — reachable under heavy multi-source playback, which is precisely the workload this PR targets. The XXX Should never happen comment at :526 is the author of the original code flagging the same doubt.
> 
> Minimum fix: if (gcp.m_openContexts.empty()) break; at the top of the loop body.
> 
> 3. Deleting the FFmpeg 6.0 rationale comment
> 
> The removed block (:5947 in the base) is institutional knowledge — it records why the pool was disabled and cites the FFmpeg docs. Since the PR resolves that concern rather than refuting it, replace it rather than delete it:
> 
> // The global context pool caps the number of simultaneously-open decoder // threads. It was disabled during the FFmpeg 6.0 migration because a context // cannot be reopened after being closed // (https://ffmpeg.org/doxygen/trunk/deprecated.html). That is now handled: // when the pool evicts a context it frees it and nulls the owning track's // avCodecContext, so openAVCodec() allocates a fresh context on next use.
> 
> 4. Reservation::~Reservation writes through a potentially-freed pointer (MovieFFMpeg.cpp:564-574)
> 
> Context& context = *m_context; context.reserved = false; // write happens first ContextKey key(context.reader, context.streamIndex); // reads from the same object if (gcp.m_contextMap.find(key) == gcp.m_contextMap.end()) return; // ...check happens after
> 
> flushContext() erases the map node (:473), which destroys the Context. If that races with a live Reservation, the dtor writes to and reads from freed memory before the liveness check that was clearly meant to prevent exactly this. The check is unsalvageable as written since it derives the key from the dead object — the Reservation should cache reader/streamIndex at construction and look up by that.
> 
> I traced the two flushContext call sites in this PR and neither currently races: the Reservation at :2450 is scoped to its case block and destroyed before initializeAudio() runs at :2550, and close() shouldn't overlap a decode on the same reader. So this is latent, not active — but it's fragile enough to be worth a comment or a proper fix while the pool is being brought back.
> 
> 5. Style — magic number (MovieFFMpeg.cpp:5956)
> 
> int poolSize = 500; — per the project's C++ conventions, non-obvious literals should be named:
> 
> constexpr int kDefaultContextPoolSize = 500;
> 
> Also worth documenting the units in a comment: it's a count of decoder threads, not contexts, which isn't obvious from the name.

Addressed 1,2,3,4

I think 5 is ok

deltag0:
There's a comment saying something "should never happen" should we try to make sure it never happens, or leave it as is?

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
bernie-laberge: Let's revert to the original code to be safe:
`ContextPool* globalContextPool = 0;`

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
bernie-laberge: Sorry I meant: please replace this comment:
`// XXX Assume this thread will never decode audio`
with this one:
```
// The context was opened only to read the audio stream parameters above.
// Drop it from the pool before freeing so the pool does not retain a
// dangling AVCodecContext* and its open-thread count stays accurate.
// Decoding clones reopen their own context on demand in audioFillBuffer().
```

deltag0: fixed

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
deltag0: fixed

src/lib/ip/IPCore/IPGraph.cpp
```diff
@@ -3744,21 +3764,31 @@ IPGraph::findNodesByAbstractPath(int frame,
 
     IPGraph::WorkItemID IPGraph::addWorkItem(const VoidFunction& function, const char* tag)
     {
-        auto jobDispatcher = reinterpret_cast<JobDispatcher*>(m_jobDispatcher);
+        if (tag)
```
bernie-laberge: The tag is optional so we should not check its value here. 
It is the m_jobDispatcher that we should check here to be safe.

src/lib/ip/IPCore/IPGraph.cpp
```diff
@@ -3744,21 +3764,31 @@ IPGraph::findNodesByAbstractPath(int frame,
...
 
-        return (WorkItemID)jobDispatcher->addJob(WorkItem(function, tag));
+        return 0;
     }
 
     void IPGraph::removeWorkItem(WorkItemID id)
     {
-        auto jobDispatcher = reinterpret_cast<JobDispatcher*>(m_jobDispatcher);
-        jobDispatcher->removeJob(id);
+        if (id)
```
bernie-laberge: It is the m_jobDispatcher that we should check here to be extra safe, not the id

src/lib/ip/IPCore/IPGraph.cpp
```diff
@@ -3744,21 +3764,31 @@ IPGraph::findNodesByAbstractPath(int frame,
...
+            auto jobDispatcher = reinterpret_cast<JobDispatcher*>(m_jobDispatcher);
+            jobDispatcher->removeJob(id);
+        }
     }
 
     void IPGraph::waitWorkItem(WorkItemID id)
     {
-        auto jobDispatcher = reinterpret_cast<JobDispatcher*>(m_jobDispatcher);
-        jobDispatcher->waitJob(id);
+        if (id)
```
bernie-laberge: It is the m_jobDispatcher that we should check here to be extra safe, not the id

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -5944,28 +5891,20 @@ namespace TwkMovie
...
-        // }
+        if (!globalContextPool)
+        {
+            int poolSize = 500;
+            if (const char* c = getenv("TWK_MOVIEFFMPEG_CONTEXT_POOL_SIZE"))
+            {
+                poolSize = atoi(c);
+            }
+            if (poolSize > 0)
+                globalContextPool = std::make_unique<ContextPool>(poolSize);
```
bernie-laberge: This should not compile anymore due to your recent change.

src/lib/ip/IPCore/IPGraph.cpp
deltag0: fixed

src/lib/ip/IPCore/IPGraph.cpp
deltag0: fixed

src/lib/ip/IPCore/IPGraph.cpp
deltag0: fixed

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
deltag0: fixed

bernie-laberge:
Closing this PR to reopen it so that we can modify a couple of things.
Thank you @deltag0 !

PR #1321, opened 2026-06-30
fix: SG-44492: harden screen selection in RvApplication against invalid indices (#1211)

pjurkas:
### Linked issues

Fixes #1211

### Summarize your change.

Harden screen selection in `RvApplication` against invalid `QScreen` indices and
null pointers so OpenRV no longer crashes (SIGSEGV in `QScreen::geometry()`) at
startup when multiple displays are connected.

Issue #1211 reports a SIGSEGV in `Rv::RvApplication::newSessionFromFiles` →
`QScreen::geometry()` on multi-monitor setups (RHEL 9.5). The crash happens
because the existing code dereferences `QScreen*` values that can be `nullptr`,
and indexes into `QGuiApplication::screens()` with values (from saved
preferences or `-screen` command-line options) that may be out of range or
stale. Under certain GNOME multi-display configurations the primary screen or
the requested screen index is not what the code assumes, and we crash.

- RHEL 9.5 with multiple displays attached — OpenRV now starts cleanly on
  configurations that previously reproduced the SIGSEGV from #1211.
- Single-display setup — no behavior change observed.
- Stale/out-of-range `-screen` command-line value — falls back gracefully
  instead of crashing.

`src/lib/app/RvCommon/RvApplication.cpp`:

- Guard `isVirtualDesktop` against a null primary screen and null entries in
  `QGuiApplication::screens()`.
- Use `QGuiApplication::screenAt()` for more robust point-to-screen mapping,
  with a manual geometry fallback when it returns null.
- Validate `opts.screen` against the current screen count before applying it,
  so stale preferences or out-of-bounds command-line values do not propagate.
- Bounds-check screen indices and null-check `QScreen*` pointers before
  dereferencing `geometry()` in the relative-position adjustment path.

Reviewer focus: please double-check the fallback path when
`QGuiApplication::screenAt()` returns null and the chosen index when
`opts.screen` is invalid — those are the behavioral edges most likely to
affect existing multi-monitor workflows.

N/A — crash fix.

### DIFF
--- a/src/lib/app/RvCommon/RvApplication.cpp
+++ b/src/lib/app/RvCommon/RvApplication.cpp
@@ -766,17 +766,48 @@ namespace Rv
 
         auto isVirtualDesktop = [&screens, primaryScreen]() -> bool
         {
-            // Not a virtual desktop if there is only one screen.
-            if (screens.size() <= 1)
+            // Not a virtual desktop if there is only one screen, or if
+            // we have no primary screen to compare against.
+            if (screens.size() <= 1 || primaryScreen == nullptr)
                 return false;
 
             QRect totalGeometry;
             for (const auto& screen : screens)
             {
+                if (screen == nullptr)
+                    continue;
                 totalGeometry = totalGeometry.united(screen->geometry());
             }
 
             return totalGeometry != primaryScreen->geometry();
+        };
+
+        //
+        //  Find the screen index for a given global point. Prefer Qt's
+        //  QGuiApplication::screenAt() which handles edge cases (such as
+        //  monitors whose top edges are not aligned, leaving gaps in the
+        //  virtual desktop bounding box) more robustly than a manual
+        //  QRect::contains() check.
+        //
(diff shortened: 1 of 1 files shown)


src/lib/app/RvCommon/RvApplication.cpp
```diff
@@ -766,19 +766,50 @@ namespace Rv
...
-            if (screens.size() <= 1)
+            // Not a virtual desktop if there is only one screen, or if
+            // we have no primary screen to compare against.
+            if (screens.size() <= 1 || primaryScreen == nullptr)
                 return false;
 
             QRect totalGeometry;
             for (const auto& screen : screens)
             {
+                if (screen == nullptr)
```
cedrik-fuoco-adsk, line 777: Could you add curly braces? That style is not followed everywhere, but we are trying to add curly braces whenever we can.

cedrik-fuoco-adsk:
Thank you @pjurkas! Could you run the pre-commit hook and fix the DCO issue?

PR #1323, opened 2026-06-30
fix: handle unknown schema in otio import more robustly

rogernelson:
### Summarize your change.

This branch will handle unknown effect schemas and unknown schemas nested inside known annotations. Instead of erroring out on these schemas, we will now simply ignore them logging an error instead. This will at least allow the import to continue with the rest of the structure that it does understand.

This is intended to improve backward compatibility, where a newer version of OpenRV might support schemas that an older one does not. 

Mac OS 26.5.1.


bernie-laberge (approved):
Beautiful. Thanks !

cedrik-fuoco-adsk (approved):

eloisebrosseau (approved):

PR #1329, opened 2026-07-03
fix: SG-43908: stale frame-invariant paint cache

rogernelson:
### Summarize your change.

This fixes an issue in the paint cache where we would use stale paint caches with erased frames.

In a regular annotation with a held sequence, only the geometry matters, it's perfectly fine to reapply the same geometry over all frames in the range. However, once we have erased strokes which reveal what is underneath the image, the annotation becomes unique to that specific frame since part of the underlying image has effectively been burned into it.

This was effectively only ever visible when using a frame-invariant source in the composition however, since the FBO id is used in the cache key, and that is only ever reused across a frame-invariant source.

Thus the simplest fix is to simply include the frame number in the cache key when we have erased stokes, so that we can re-evaluate the underlying image for each frame regardless of what is in the composition.

This fix an issue in the context of live review which has a dedicated annotation track over a gap composited over top of another track containing the media.

Mac OS 26.5.1

### DIFF
--- a/src/lib/ip/IPCore/IPCore/ImageRenderer.h
+++ b/src/lib/ip/IPCore/IPCore/ImageRenderer.h
@@ -799,7 +799,7 @@ namespace IPCore
         //
 
         void renderImage(InternalRenderContext&);
-        void renderPaint(const IPImage*, const GLFBO*);
+        void renderPaint(const IPImage*, const GLFBO*, int frame);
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):
Beautiful !

eloisebrosseau (approved):

PR #1333, opened 2026-07-06
fix: SG-43881: Fix scrollbar on reordering with previews enabled

eloisebrosseau:
### Summarize your change.

Record the top most input in the visible viewport before clearing it and scrolling back to that position once the list has been updated.

When "Show Source Preview" is enabled in the Session Manager, reordering inputs up and down caused the inputs list to scroll back to the top. A reorder rebuilds the list in `updateInputs`. This means that all the rows are removed from the model and added back in the new order. With previews enabled, each source row also has a preview QWidget. Clearing the model destroys those widgets and the rebuild creates new ones, which resets the scroll position to the top. Without previews, there are no QWidgets to recreate, which explains why the issue is only visible when "Show Source Preview" is enabled.

Reordering up and down a single input and multiple inputs with and without "Show Source Preview" was tested on macOS.

### DIFF
--- a/src/plugins/rv-packages/session_manager/session_manager.mu.in
+++ b/src/plugins/rv-packages/session_manager/session_manager.mu.in
@@ -1486,6 +1486,13 @@ class: SessionManagerMode : MinorMode
 
         _inputOrderLock = true;
 
+        string topNode = nil;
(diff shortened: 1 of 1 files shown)


cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
LGTM

PR #1334, opened 2026-07-06
fix: SG-43924 Fix text size slider in the UI

rogernelson:
### Fix text size slider

The UI text tool already had a size slider wired through the generic sizeSlider/setBrushSize path, but the text node's fontSize property was hardcoded to 24.0, so moving the slider had no visible effect. The slider's min/max range was also miscalibrated.

Also recalibrate text size slider range to match legacy FTGL sizing. The 24/48/72-out-of-1080 range made max-size text noticeably smaller than the old FTGL renderer produced. The legacy renderer computed ptsize = size * 10000 (15-100 pt for the old 0.0015-0.01 slider range), a vector-glyph scale unrelated to the new QFont raster pipeline's WCS-fraction fontSize. Reinterpreting that 15-100 range as pixels-at-1080p gives 15/1080-100/1080, restoring a comparable max size.

Lastly, load the same Luxi Sans font for QFont-rendered text annotations that was used by FTGL to get the same font look.

Make the text slider useable again

Mac OS 26.5.1

### DIFF
--- a/src/lib/ip/IPCore/PaintCommand.cpp
+++ b/src/lib/ip/IPCore/PaintCommand.cpp
@@ -19,6 +19,7 @@
 #include <TwkMath/Iostream.h>
 #include <TwkMath/Frustum.h>
 #include <TwkGLText/TwkGLText.h>
+#include <TwkGLText/defaultFont.h>
 #include <stl_ext/string_algo.h>
 #include <QString>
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):
LGTM

eloisebrosseau (approved):

PR #1336, opened 2026-07-07
fix: SG-43948 more stale paint cache fixes

rogernelson:
### Summarize your change.

I previously fixed a stale cache bug with eraser strokes in https://github.com/AcademySoftwareFoundation/OpenRV/pull/1329. However, a similar problem exists more generally as well. 

Even if we do not have erase strokes and we are using hold and/or ghost we still need to make the cache frame-dependent in order to work when composited over a frame-invariant source, since if we are using the same frame invariant FBO and the same geometry, we still may need to be able to see whatever is being composited underneath that frame-invariant source (if it has alpha, is a gap, etc).

Thus, in order to fix this in all cases, I added a flag we can set to a PaintCommand when hold and/or ghost are set.  When the flag is true, we do the same thing we did for the previous fix -- append the frame number to the cache key.

Note that I did not revert the previous fix in favour of this one, since erase strokes could still technically have a duration > 1 and we would then have the same bug without hold or ghost, so I kept them complimentary.

There was also an off-by-one bug in PaintCommand that I thought was the initial cause of this, but it was not. It's still a fix worth doing, so I kept it. The real source of my problem was that when comparing cache keys we were only comparing up to length of the source string( so xyz would match xyzFrameN for example).  

The Review app in FPT constructs a topology where all the annotations are on a dedicated gap track above the media, which exposes this particular bug. This fix will make RV composite that topology correctly.

Mac OS 26.5.1

### DIFF
--- a/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
+++ b/src/lib/ip/IPBaseNodes/PaintIPNode.cpp
@@ -70,6 +70,18 @@ namespace
         return std::min(1.0f, ghostOpacity);
     }
 
+    // This should be set if the paint command depends on the frame content
+    // and as such cannot be cached independently of it -- such as holding or
+    // ghosting, where the underlying frame needs to change even though the
+    // annotation geometry remains the same.
+    void setFrameDependent(PaintIPNode::LocalCommand* command, bool value)
+    {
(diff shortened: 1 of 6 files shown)


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #1337, opened 2026-07-07
fix: SG-43961 text size in otio_reader

rogernelson:
### Summarize your change.

After https://github.com/AcademySoftwareFoundation/OpenRV/pull/1334 I need to make the same changes as were made there to the text scaling.  Text sizing is now simply a normalized value, so just input it directly from the OTIO Annotations as such.

Fix image scaling in OTIO import.

Mac OS 26.5.1


bernie-laberge (approved):
LGTM

eloisebrosseau (approved):

PR #1338, opened 2026-07-08
fix: SG-43970 loading pre-qt font sizes, line wrapping, backwards comp

rogernelson:
### Summarize your change.

pre-QT fonts were size was in points (based on a 1080p source size). If we load a session file containing the old value, convert it to the new normalized value (fraction of image height), so we can load older sessions.

If you tried to load the old default value of 24, it will try to load something 24x the size of the image, and hang for a long time, so I added a reasonable max clamp in the renderer as a safety measure to avoid that.

I also made sure that multi-line text (user hits shift-enter) scales down.

Backwards compatibility with FTGL font renderer

Mac OS 26.5.1

### DIFF
--- a/src/bin/imgtools/rvio/main.cpp
+++ b/src/bin/imgtools/rvio/main.cpp
@@ -88,6 +88,8 @@
 #include <gl/glew.h>
 #include <QtGui/QtGui>
 #include <QtWidgets/QApplication>
+#else
+#include <QtGui/QGuiApplication>
 #endif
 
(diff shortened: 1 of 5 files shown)


bernie-laberge (approved):

cedrik-fuoco-adsk (approved):

bernie-laberge (approved):

bernie-laberge (approved):
LGTM

PR #1339, opened 2026-07-08
fix: SG-43974 Make font decorators case-insensitive

rogernelson:
### Summarize your change.

This makes the font decorator fields (decoration, weight, stlye) case insensitive

In the LiveReview protocol, the web client was sending all upper-case strings.

Mac OS 26.5.1


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #1340, opened 2026-07-08
fix: SG-43880: Optimize session manager inputs reordering with previews

eloisebrosseau:
### Summarize your change.

Add a cache for the previews to avoid loading from disk each thumbnail and filmstrip in `makeSourceRowWidget`. When the image stored at a specific path is updated, the cache will drop the preview for it, and the new image will be added to the cache on the next call to `makeSourceRowWidget`.

Reordering and erasing sources from the inputs list with `Show Source Preview` enabled would take exponentially more time  the more sources there were. If the option was disabled, it would take less than 1 second no matter the number of sources loaded. The user experience while reordering or erasing sources shouldn't be affected by previews being enabled in the session manager.

Reordering and erasing sources from the inputs list with `Show Source Preview` enabled was tested on macOS.

### DIFF
--- a/src/plugins/rv-packages/session_manager/session_manager.mu.in
+++ b/src/plugins/rv-packages/session_manager/session_manager.mu.in
@@ -621,10 +621,13 @@ class: ThumbnailWidget : QLabel
 
     method: setFallback (void; QPixmap pixmap) { setPixmap(pixmap); }
 
-    method: load (void; string path)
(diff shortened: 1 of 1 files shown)


cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
LGTM

PR #1342, opened 2026-07-10
fix: SG-44076: Fix session file thumbnail generation

eloisebrosseau:
### Summarize your change.

When an RV session file is loaded, the whole session is cleared, triggering a `before-clear-session` event. However, in `local_thumbnail_gen.py`, the listener on that event is setting `self._shutting_down` to True to prevent thumbnails and filmstrips generation when clearing a session. However, nothing was listening to `after-clear-session` in order to set back `self._shutting_down` to False, preventing the generation of previews after that point for RV session files. It was working for movie clips because this would trigger a `before-progressive-loading` event, and the handler of this event would set back `self._shutting_down` to False.

Thumbnails and filmstrips of sources in the Session Manager panel were not being generated with `Show Source Preview` enabled when loading an RV session file.

Loading an RV session file with `Show Source Preview` enabled in the Session Manager was tested on macOS.

### DIFF
--- a/src/plugins/rv-packages/session_manager/local_thumbnail_gen.py
+++ b/src/plugins/rv-packages/session_manager/local_thumbnail_gen.py
@@ -109,8 +109,13 @@ class LocalThumbnailGen(rvtypes.MinorMode):
             ),
             (
                 "before-clear-session",
-                self._on_clear_session,
+                self._on_before_clear_session,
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
Good catch ! Thanks!

cedrik-fuoco-adsk (approved):

PR #1343, opened 2026-07-10
fix: SG-44089 add soft_deleted to shape schemas

rogernelson:
### Summarize your change.

This branch adds a `soft_deleted` flag to the shape schemas and uses it in the annotation_hook to set visibility.  This will allow us to receive payloads and hide annotations that have been undone.

This fixes setting the initial payload from the Review app in FPT when it contains shape annotations that have been undone.

Mac OS 26.5.1.


bernie-laberge (commented):
LGTM

PR #1350, opened 2026-07-16
fix: SG-43367: eliminate EXR+WAV playback stuttering

bernie-laberge:
DO NOT REVIEW : Waiting for customer confirmation that this fix solves their issue first

Rebased on `main` after #1375, which moved `paintGL()`/`eventProcessingTimeout()` out of `GLView` into the new `GLWindow` class -- hence the playback instrumentation below lives in `GLWindow.cpp`.

### Linked issues
NA

### Summarize your change.

Fixes stuttering during playback of high-resolution multi-part EXR sequences with WAV/MOV audio on Linux. The image was correct but playback dropped well below the target frame rate even with frames fully resident in cache. Root causes were in the GPU texture-upload path, the EXR decode scheduling, and an OpenEXR 3.3 I/O regression. Playback now holds 24 fps with ~100% of large frames on the DMA fast path and per-frame present stalls effectively gone.

GPU texture upload (primary fix) -- src/lib/ip/IPCore/ImageRenderer.{cpp,h}:
- Make the staging PBO transient. It was allocated in initializeTexture() and pinned to the cached TextureDescription for the texture's entire cache life (~10+ frames), so with a large look-ahead cache the small fixed PBO pool was permanently occupied and most large frames fell back to the slow synchronous glTexImage2D-from-client-memory path (~72 ms, ~1.4 GB/s). The PBO is now acquired inside uploadPlane() for a single upload and released back to the pool immediately after the transfer is issued, so a handful of buffers serve an unbounded stream. Result: PBO fast-path use went ~64% -> 99.9% and CPU submit ~72 ms -> ~14 ms (~6.8 GB/s) for 4546x2864 frames.
- Upload 3-channel RGB half/float frames via a native RGBA path: allocate the texture as GL_RGBA16F/RGBA32F and expand RGB->RGBA on the CPU during upload (GL_TEXTURE_RECTANGLE only). GL_RGB16F/RGB32F have no native DMA path; the driver was expanding per-pixel on upload, which is far slower.
- Fix a texture cross-reuse corruption: an expanded 3ch->RGBA texture and a genuine 4ch RGBA texture produce identical destination geometry/format, so the getTexture() Compatible-reuse path could recycle one for the other and run the wrong upload path (wrong channel stride), causing image artifacts. compatible() now also compares expandRGBToRGBA so the two are never reused across each other.

PBO pool sizing -- src/lib/graphics/TwkGLF/GLPixelBufferObjectPool.cpp:
- Raise the fixed-size PBO size constants from `x6` to `x8` bytes/px (`MAX_SIZE` `4032*4536*8`, `MIN_SIZE` `1920*1080*8`). Both express resolution x bytes-per-pixel, and the RGB->RGBA expansion above changes the uploaded format from 3-channel half (6 B/px) to 4-channel half (8 B/px). Left at 6, `MAX_SIZE` gave an effective ceiling of 13.7 Mpx at RGBA16 -- only 5.1% above the 4546x2864 frames in the repro -- and a frame over the max makes `pop()` return NULL, silently falling back to the slow non-PBO upload, i.e. reinstating the very stall the transient-PBO change removes. Measured on 4546x3400 (117.9 MiB) frames, same media and window: old ceiling 0.0% PBO / 1.97 GB/s / 414 uploads; new ceiling 99.9% PBO / 5.23 GB/s / 739 uploads. The pool pre-allocates ceil(7*1.05)=8 buffers of `MAX_SIZE`, so the pre-allocated footprint goes 837 -> 1116 MiB. `MIN_SIZE` is consistency only: it is read solely inside `initPBOPool()` branches gated on a `poolUpperLimit` that is always assigned `poolSize`, so nothing reachable consumes it today.

RGB->RGBA expanders -- src/lib/image/TwkFB/FastMemcpy.{cpp,h}:
- Add parallel (task-pool) expand_rgb_to_rgba_16bit/32bit(_MP) helpers used by the upload path.

EXR decode scheduling -- src/lib/ip/IPBaseNodes/FileSourceIPNode.cpp, src/lib/ip/IPCore/IPGraph.{cpp,h}, src/lib/ip/IPCore/IPCore/IPNode.h:
- When a source mixes slow-random-access media (e.g. a MOV used only for audio) with fast EXR image media, no longer serialize the whole source onto a single caching thread. testEvaluate() now inspects only the media component actually supplying the displayed image and reports hasFastVideoSource, and the caching scheduler only forces single-threaded slow-media handling when the frame has no fast image source to parallelize. This restores concurrent EXR decoding for the common "EXR + MOV-for-audio" layout.

OpenEXR 3.3 I/O regression -- src/lib/image/IOexr/FileStreamIStream.{cpp,h}, src/lib/image/IOexr/IOMultiPartEXR.cpp:
- OpenEXR 3.3.x is much slower for custom Imf::IStream subclasses that do not implement size()/stateless read(). RV's default memory-mapped I/O uses such a stream. Implement size(), isStatelessRead() and the stateless read() overload (guarded by IOEXR_HAS_STATELESS_ISTREAM) so 3.3+ takes the fast, concurrent read path from RV's mapped buffers.

Playback diagnostics (opt-in, env-gated) -- src/lib/base/TwkUtil/ PlaybackDiagnostics.{cpp,h} (+ CMakeLists), and instrumentation in Session.{cpp,h}, GLWindow.cpp, FBCache.{cpp,h}, IPGraph.cpp, FileSourceIPNode.cpp, ImageRenderer.cpp, ALSASafeAudioModule/ALSASafeAudioRenderer.cpp (+ CMakeLists):
- Add a thread-safe, RV_PLAYBACK_DIAG-gated CSV logger capturing background decode times/concurrency/compression, audio cache-miss/underrun events, buffering pauses and frame skips, display pacing (interval/dframe/refreshes), cache hit/miss and runway, per-plane GPU upload cost/PBO usage, and a stall anatomy split (present/swap vs in-graph work). All timing is behind the env gate and has no cost when disabled.
- Remove two leftovers of hypotheses this investigation disproved: the `RV_REDRAW_MIN_INTERVAL_MS` redraw throttle in `IPCore/Application.cpp` (a no-op at its default of 0; posting rate was not the stall, and the QOpenGLWindow viewport no longer presents through a widget composite queue), and the analyzer verdict that pointed a zero-cache-miss result at `render_v2 targetFrame/shift` pacing. Both now point at the upload path instead.
- tools/analyze_playback_diag.py: analyzer that summarizes the log and reports a bottleneck verdict (decode vs audio vs cache vs present).

Playback of high-resolution multi-part EXR sequences with WAV/MOV audio on Linux could potentially stutter

Successfully tested on Rocky Linux 9


KevinJW:
Not wanting to derail anything here, but something that used to be beneficial in the old days of 10bit DPX, was performing the GPU upload with the 'wrong' pixel type and then running a shader to fix up the pixels after it was uploaded. In the old case you upload 10bit RGB packed (32 bits) as 8 bit RGBA (same 32 bits size) and then do the unpack on the GPU. This was a similar case where the 8bit RGBA upload case was optimised.

In this case the EXR originating as 16 bit RGB but pretending to be 16bit RGBA the bandwidth used to to the upload would be 25% less at the cost of a more complex RGB -> RGBA buffer conversion on the GPU. though  I'm guessing we are not hitting any bandwidth limits on PCIe4 or later devices, even if you account for reading things back to feed to SDI presentation devices.

It was a problem on older gen 3 PCIe GFX cards at these kinds of resolutions.

PR #1356, opened 2026-07-21
Draft: SG-44542: source setup modernization

herronelou:
### Summarize your change.

In this MR I do a pass of trying to modernize ocio_source_setup for python 3.10+, attempting to maintain compatibility, logic and signatures from previous implementation.
I would have liked to change a bit more but figured I’d limit the scope for this MR to mostly stylistic changes.
I made separate commits for each change I made so we could pick and choose what would make sense to use or not.

This module used some non-conventional python patterns, making the code difficult to read and follow. Rather than an extensive rewrite, I figured cleaning up the odd patterns and adding docstrings could be a step in the right direction.

This draft is currently barely tested, and some of the commits were done with assistance from Gemini. Drafting this from Siggraph after a conversation with @rogernelson

### DIFF
--- a/src/plugins/rv-packages/ocio_source_setup/ocio_source_setup.py
+++ b/src/plugins/rv-packages/ocio_source_setup/ocio_source_setup.py
@@ -81,9 +81,9 @@ def ocio_node_from_media(
 
     result = [{"nodeType": d, "context": {}, "properties": {}} for d in default]
 
-    node_type = commands.nodeType(node)
+    nodeType = commands.nodeType(node)
 
-    if node_type == "RVDisplayPipelineGroup":
+    if nodeType == "RVDisplayPipelineGroup":
         display = config.getDefaultDisplay()
         result = [
             {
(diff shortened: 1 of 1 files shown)


src/plugins/rv-packages/ocio_source_setup/ocio_source_setup.py
```diff
@@ -103,111 +147,228 @@ def ocio_node_from_media(config, node, default, media=None, attributes={}):
...
-            groupName = "RVDisplayPipelineGroup"
-            dpipeline = groupMemberOfType(group, groupName)
-            dOCIO = groupMemberOfType(dpipeline, "OCIODisplay")
-            managed = commands.getIntProperty(dOCIO + ".ocio.active")[0] != 0
-            return commands.CheckedMenuState if managed else commands.UncheckedMenuState
-        except Exception:
-            return commands.UncheckedMenuState
 
-    return F
+def isOCIOManaged(nodeType: str) -> Callable[[], int]:
```
herronelou, line 167: I’d have removed these as I don’t think anyone should really import from this file, but decided to maintain “public API” for this PR

PR #1359, opened 2026-07-22
feat: SG-43305: Add new annotation UI to expose new features

eloisebrosseau:
### Summarize your change.

This PR brings the work of Roger to add a beta Python-based annotation package with a brand new UI. Several PRs will follow this one to polish the new tool and fix some issues. Note that the current Mu-based annotation package is set as the default in the preferences but the new one can be loaded by showing the hidden packages.  A new `excludes` field was added to the PACKAGE fields to list conflicting packages. This allow us to automatically unload a package when a conflicting one is loaded.

We want to modernize and unify annotation tools to get a better user experience during review sessions.

Loaded both annotations packages was tested on macOS.


bernie-laberge (approved):
LGTM !

src/lib/app/RvPackage/PackageManager.cpp
```diff
@@ -265,6 +265,11 @@ namespace Rv
                 }
             }
 
+            for (const int exclusion : package.exclusions)
+            {
+                allowLoading(m_packages[exclusion], false, depth + 1);
```
rogernelson, line 270: If A excludes B and B excludes A, does that mean neither will load, or whichever loads first will win and exclude the other?

rogernelson (approved):
Looks good, thanks!

src/lib/app/RvPackage/PackageManager.cpp
eloisebrosseau: I made it mutually exclusive so that you only need to add the `excludes` field to one of the packages for simplicity but having each package excludes the other wouldn't change the behaviour because both package exclusion lists will be updated to contain the other package: https://github.com/AcademySoftwareFoundation/OpenRV/pull/1359/changes#diff-e0d325be388e4038f2c05616cb63ff0e666bdaafadc586ce96bb2fa86ae1a4e6R1587.
Thus, making A excludes B and B excludes A will give you the exact same result as only making A excludes B without making B explicitly excludes A.

`allowLoading` is only used by selecting a package to be loaded in the RV Preferences tab. This loop will simply make the other package that got added in the line I shared above not loadable anymore, so that when you relaunch RV, only one is loaded.

PR #1360, opened 2026-07-22
fix: SG-44428: Fix text cursor

eloisebrosseau:
### Summarize your change.

This PR adds a few missing session events the annotation package needs to listen on to commit the text annotations and remove the cursor that is displayed on the frame while typing. The method `on_text_commit` is passed as a lambda to know if the event needs to be consumed (i.e. hotkeys) or not (i.e. session events) before committing the text annotations.

When saving a session after adding text annotations, the "|" indicating where the cursor is on the screen was still being kept as part of the annotations. This caused the symbol to be shown when reloading the session file previously saved.

Adding text annotations, saving the session file, and reloading the session file to validate that the text cursor is not displayed on the frame anymore was tested on macOS.


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #1370, opened 2026-08-04
fix: SG-42147: Fix rvpkg -remove <package_name>

eloisebrosseau:
### Summarize your change.

- [X] Add `baseNameFromPackageFileName()` helper function to avoid duplicating regex matching that was already present in `loadPackageInfo()`
- [X] Replace all `cd()` calls with `filePath()` to make sure the path is formed the expected way instead of failing silently
- [X] Remove unnecessary `pfile` variable that was unused in `uninstallPackage()`
- [X] Remove unnecessary supportdir/<package_name> directory creation just to delete it after and only call `removeRecursively()` if that directory exists
- [X] Set `m_packages[i].file` and `m_packages[i].baseName` respectively to the missing package path and name to make sure both values exist before using them in `uninstallPackage()`
- [X] Validate that the incoming package to remove actually exists in `removePackages()` to prevent logging an error about it not being removed if it was already manually removed

When calling `rvpkg -remove <package_name>` after manually deleting the .rvpkg for that package, the whole parent directory of the one you were currently in would get recursively removed. Since the file did not exist, the bare file name from rvinstall was kept instead of its absolute path causing `uninstallPackage()` to resolved against the current working directory. Every `cd()` into the sub-directories were silently failing since nothing was checking the return value of the call, so `supportdir` was left pointing at the parent of the current working directory (because of `rdir.cdUp()`) when `removeRecursively()` was called on it. Some other QDir errors were also shown because the package base name would resolved to an empty string which is not a valid directory.

Calling `rvpkg -remove <package_name>` after manually deleting the .rvpkg for that package was tested on Rocky Linux 9.

### DIFF
--- a/src/lib/app/RvPackage/PackageManager.cpp
+++ b/src/lib/app/RvPackage/PackageManager.cpp
@@ -20,6 +20,21 @@
 
 #include <QRegularExpression>
 
+namespace
+{
+    QString baseNameFromPackageFileName(const QString& fileName)
+    {
+        QRegularExpressionMatch match = QRegularExpression(R"((.*)-[0-9]+\.[0-9]+\.rvpkg)").match(fileName);
+
+        if (!match.hasMatch())
+        {
+            match = QRegularExpression(R"((.*)\.zip)").match(fileName);
+        }
+
+        return match.hasMatch() ? match.captured(1) : QString();
+    }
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
LGTM

PR #1372, opened 2026-08-07
fix: SG-44913: Fix QOpenGLWidget import for PySide6

tjjackson:
### Summarize your change.

Fixed import for QOpenGLWidget in PySide6.

In PySide6, QOpenGLWidget (used by rv.qtutils.sessionGLView) was moved from PySide6.QtWidgets to PySide6.QtOpenGLWidgets. 

Tested with a CY2025 build on Rocky Linux 9.7

### DIFF
--- a/src/lib/app/py_rvui/rv/qtutils.py
+++ b/src/lib/app/py_rvui/rv/qtutils.py
@@ -16,6 +16,7 @@ except ImportError:
         from PySide6 import QtGui, QtWidgets
         from PySide6.QtGui import *
         from PySide6.QtWidgets import *
+        from PySide6.QtOpenGLWidgets import QOpenGLWidget
         from shiboken6 import wrapInstance
         from shiboken6 import getCppPointer
     except ImportError:


bernie-laberge:
Hello @tjjackson, thanks for catching this and putting up the fix.

Since you opened this PR, a refactor of the RV viewport took place: PR #1375 which changed rv.qtutils.sessionGLView() in src/lib/app/py_rvui/rv/qtutils.py so it no longer wraps the session view as a QOpenGLWidget at all — it's now a plain QWidget (the GL surface is a native QOpenGLWindow embedded via createWindowContainer). So qtutils.py doesn't reference QOpenGLWidget anymore, and the import fix here would just be a dead import on top of current main.

Hopefully, the viewport refactor did not break any of your RV packages or integration. Please let us know if it did.
Let us know if this is OK that we close this PR or you can close if yourself if you prefer.

Thanks again for the contribution!

bernie-laberge (approved):

tjjackson:
OK, I'll close it out if it is no longer relevant, thanks

PR #1373, opened 2026-08-10
fix: SG-44546: Fix annotations always enabled

eloisebrosseau:
### Summarize your change.

The new annotation package was updated to match the behaviour of the old one where the mode manager is the one responsible of handling the activation and deactivation instead of having it being active all the time. Methods related to manage not showing the new panel at the same time as the old was was removed since only one package can be loaded at once. Some cleanup was also done to remove mentions of Live Review in annotate_meta_mode.py.

The new annotation tool was always enabled instead of being activated and deactivated by the mode manager like the old annotation tool is. This was prevented the normal events for play/pause and playback to go through, and resulted in strokes being drawn on the frames instead. 

Loading a clip with the new annotation panel closed, and clicking on the frames or the timeline to validate that playback was triggered instead of drawing strokes, was tested on macOS.


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #1376, opened 2026-08-12
build: Fix Windows Debug OTIO pip build (cmake 4.4.1+ FindPython, python311_d.lib)

cedrik-fuoco-adsk:
### Linked issues
  none
  ### Summarize your change
  Fix Windows **Debug** CI failures when pip builds OpenTimelineIO against our custom debug Python.
  Two changes in `cmake/dependencies/python3.cmake` and a new `src/build/pip-build-constraints.txt`:
  1. **Pin cmake in pip build-isolation**: OTIO's `setup.py` runs `find_package(Python Development.Module)` using the cmake wheel pip installs into its isolated build env, not the cmake driving the outer OpenRV build. CMake **4.4.1+** changed Windows `FindPython` to prefer release-only ABI by default, which breaks against our debug artifacts (`python311_d.lib`, `python_d.exe`). Pin `cmake==3.31.6` (same line as the outer build) via `PIP_CONSTRAINT` and `PIP_BUILD_CONSTRAINT`.
  2. **Prepend `LIB` on Windows Debug**: Even with FindPython fixed, OTIO's debug Python headers `#pragma-link python311_d.lib`. Pip builds under `D:/_t/...`, so MSVC cannot find our import library without prepending the Python libs directory to `LIB` during the pip install step (using `cmake -E env --modify`, same pattern as the existing PATH fix).
  Release builds are unaffected.
  ### Describe the reason for the change
  Windows Debug jobs have failed on every nightly since **2026-08-05** with no OpenRV source change. Two nightlies of the same commit straddle the boundary: Aug 4 passed, Aug 5 failed. **cmake 4.4.2** was published to PyPI on **2026-08-04**, between those runs.

### DIFF
--- a/cmake/dependencies/python3.cmake
+++ b/cmake/dependencies/python3.cmake
@@ -250,6 +250,13 @@ SET(_requirements_output_file
 
 CONFIGURE_FILE(${_requirements_input_file} ${_requirements_output_file} @ONLY)
 
+# Constraints for the build-isolation environments pip creates for packages built from source. Their build requirements are resolved from PyPI on every build
+# and are otherwise unpinned, so an upstream release can break a build that has not changed. See the file itself for what is pinned and why. PIP_CONSTRAINT must
+# be an absolute path: pip resolves it relative to the working directory of each build-environment subprocess, which is not ours.
+SET(_pip_constraint_file
+    "${PROJECT_SOURCE_DIR}/src/build/pip-build-constraints.txt"
+)
+
 # OpenTimelineIO needs to be built from source with CMAKE_ARGS to ensure it uses the correct custom-built Python libraries. This is required for both old and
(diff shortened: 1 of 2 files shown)


bernie-laberge (approved):
LGTM

eloisebrosseau (approved):
What is the purpose of this new pip-build-constraints.txt?

cedrik-fuoco-adsk:
It pins the versions pip is allowed to resolve inside the build-isolation environments (PEP 517) that pip creates when it builds a package from source.

When a package installed through requirements.txt is built from source, pip resolves that package's build requirements (cmake, in this case) in a separate isolated environment that doesn't inherit our requirements.txt pins. The constraints file is what pins those. @eloisebrosseau

PR #1380, opened 2026-08-17
feat: SG-44605: Optimize the loading and scrubbing of S3/Streaming media in RV

deltag0:
### Summarize your change.
- Updated FFMPEG 8 to 9
- Added local caching for streamed media using the FFMPEG `shared:` protocol
- Added background threads to open FFMPEG contexts for MovieReader clones since FFMPEG even if all the media is cached with FFFMPEG, it will try to establish a connection, which adds latency (even if everything was cached). If we do this process at the launch of RV, the latency is removed
- Added background threads to start caching media over the internet in the background to optimize scrubbing further
- Added a lookahead thread that reads and caches ahead of the playhead which, only 1 can be active, and if there is a lookahead thread in the worker queue, no new workers are started.
- Disabled slow random access for streamed media

Previously, when media hosted on shotgrid was opened in OpenRV, to view a new file, there had to be a request made over the network, which was adding 1-2 seconds of latency every time a user jumped to a new file. This made scrubbing virtually impossible because of the delay. To improve this problem, the idea was to cache the media's raw bytes locally, so that RV can decode the bytes locally instead of having to establish a new HTTP connection for a new file. This would work, but would require us to manually jump to the new file for the caching to start. So, background threads to download every file were spawned.
After testing, on my machine, I found that 4 threads for background fetching provided the best performance.
Soon after, it was found that even if all media was cached _before_ a user streams it on RV, scrubbing would still be painful because there were FFMPEG reader clones which had to establish their own connection, so they still added latency. To fix this, the connections were established at load time, with background threads, when the media is opened (after pre-loading).

The lookahead thread helps with scrubbing when not everything is cached, however it currently still reads ahead if all the media is cached. When testing without it, scrubbing was still laggy for the first half minute until the background threads finished caching everything, but with it enabled, scrubbing worked almost immediately.

- Tested on a Mac
- Viewed improvements that if files are cached locally, scrubbing is instant

- Verify update from FFMPEG 8 to 9

Notes:

* Currently no limit to the size of the raw media cache

* Background threads will still run even if all media is cached (although they will just read from disk)

* Verified that warming reader clones was still necessary after background threads and lookahead thread, the performance degrades without it

* manual latency test: 
```
time ffmpeg -loglevel error \
  -headers $'Cookie: _session_id=fa0d43df120581721917cd2d820484e8\r\n' \
  -i 'https://sg-create-prod.shotgunstudio.com/file_serve/version/14232/mp4?20097' \
  -frames:v 1 \
  -f null -
ffmpeg -loglevel error -headers  -i  -frames:v 1 -f null -  0.19s user 0.06s system 21% cpu 1.159 total
rel (.venv) termev@ADSKJ7HQLQCG6D rv %
Goiing through S3 directly only saves about 0.1ms
```

* Tested blocking new workers while the lookahead thread was active, but observed no noticeable performance improvement


src/lib/mu/MuQt6/qt2mu.py
```diff
@@ -1 +1 @@
-qt6_to_mu.py
\ No newline at end of file
+qt6_to_mu.py
```
eloisebrosseau: Is there a reason why you touched this file? It looks like you like what you did broke the symlink

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -79,17 +85,55 @@ namespace TwkMovie
...
+            }
+            setStatus(Reader::Status::LOADED);
         }
         catch (...)
         {
             setStatus(Reader::Status::LOADERROR);
         }
     }
 
+    void GenericIO::Preloader::Reader::queuePrefetch()
```
eloisebrosseau: This method can be made const

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -79,17 +85,55 @@ namespace TwkMovie
...
+
+        if (evUseUploadedMovieForStreaming.getValue())
+        {
+            boost::replace_all(url, "#.mp4", "");
+        }
+
+        url = "shared:" + url;
+
+        StreamerPool::Options options;
+        options.push_back(std::make_pair("cache_dir", evStreamCachePath.getValue()));
```
eloisebrosseau: You should use `emplace_back` instead of `push_back` here

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -79,17 +85,55 @@ namespace TwkMovie
...
         catch (...)
         {
             setStatus(Reader::Status::LOADERROR);
         }
     }
 
+    void GenericIO::Preloader::Reader::queuePrefetch()
+    {
+        if (!TwkUtil::pathIsURL(filename()))
+            return;
```
eloisebrosseau, line 105: Try to put your statement inside braces even if they are only one line. I see you have a mix of doing it and not doing it across the different files you changed

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -79,17 +85,55 @@ namespace TwkMovie
...
+
+        StreamerPool::Options options;
+        options.push_back(std::make_pair("cache_dir", evStreamCachePath.getValue()));
+        options.push_back(std::make_pair("seekable", "1"));
+        options.push_back(std::make_pair("reconnect", "1"));
+        options.push_back(std::make_pair("multiple_requests", "1"));
+
+        const auto& parameters = request().parameters;
+
+        for (size_t i = 0; i < parameters.size(); i++)
```
eloisebrosseau: You should use range-based for loop when possible

src/lib/image/TwkMovie/MovieIO.cpp
```diff
@@ -179,13 +225,18 @@ namespace TwkMovie
...
     void GenericIO::Preloader::addReader(const std::string_view filename, const Movie::ReadRequest& request)
     {
-        std::unique_lock<std::mutex> lock(m_schedulerThread_mutex);
+        std::shared_ptr<Preloader::Reader> newReader;
 
-        //        std::cerr << "PRELOADER ADD " << filename << std::endl;
+        {
+            std::unique_lock<std::mutex> lock(m_schedulerThread_mutex);
+
+            //        std::cerr << "PRELOADER ADD " << filename << std::endl;
```
eloisebrosseau: Since you're modifying this method, I believe you should remove this commented out line

src/lib/image/stream/stream/StreamPreloadPool.h
```diff
@@ -0,0 +1,119 @@
...
+
+    void reapFinished();
+
+    //
+    // Callback if connection gets interrupted, to not waste time if connection dies.
+    //
+
+    static int interruptCallback(void* opaque);
+
+private:
```
eloisebrosseau: You already declared `private` at line 62. You don't need to repeat it here

src/lib/image/TwkMovie/TwkMovie/MovieReader.h
```diff
@@ -160,6 +161,7 @@ namespace TwkMovie
     protected:
         std::string m_filename; /// derived class should set this to the filename
         ReadRequest m_request;
+        boost::mutex m_deferredOpenLock;
```
eloisebrosseau: You should use `std::mutex` instead of `boost::mutex`

src/lib/image/stream/StreamPreloadPool.cpp
```diff
@@ -0,0 +1,320 @@
...
+{
+#include <libavformat/avformat.h>
+#include <libavformat/avio.h>
+#include <libavutil/dict.h>
+#include <libavutil/error.h>
+#include <libavutil/mathematics.h>
+}
+
+// 32 KB
+const int readChunkSize = 32 * 1024;
```
eloisebrosseau: `readChunkSize` should be a constexpr variable declared directly in `download()` since it is only used there

src/lib/image/stream/StreamPreloadPool.cpp
```diff
@@ -0,0 +1,320 @@
...
+            worker_it = m_activeWorkers.erase(worker_it);
+        }
+        else
+        {
+            ++worker_it;
+        }
+    }
+}
+
+void StreamerPool::workerFunc(Job job)
```
eloisebrosseau, line 168: Why is `job` passed as a copy here? I believe it should be a const reference instead like you did for `download` and `downloadWindow`

eloisebrosseau:
I appreciate that you are trying to clean up some of the dead code in the IPCore directory, but I feel like it is unrelated to the purpose of this PR, since the removals are made on files that you are not touching otherwise. You could create a second PR for that cleanup if you want, and remove it from this one, to help us review only what is actually important for the feature to work. Same thing for `MovieFFMpegReader::scan()`. Have you checked why it was kept with just a comment instead of being removed? If it's not related to your changes, I would leave it alone unless you are 100% sure this can be safely removed

src/lib/mu/MuQt6/qt2mu.py
deltag0: I think it was auto formatted, because it was changed when I ran the pre-commit

eloisebrosseau: Are you sure of that? pre-commit only runs on files you are about to commit so there is no reason for you to have touched this file in the first place. Also, this is a symlink, so there is no reason for pre-commit to run on it

src/lib/image/stream/StreamPreloadPool.cpp
deltag0: We pass job from the `schedulerLoop` which is local to the function

src/lib/image/TwkMovie/MovieIO.cpp
deltag0: fixeed

src/lib/image/TwkMovie/MovieIO.cpp
deltag0: fixed

src/lib/image/TwkMovie/MovieIO.cpp
deltag0: fixed

src/lib/image/TwkMovie/MovieIO.cpp
deltag0: fixed

src/lib/image/stream/stream/StreamPreloadPool.h
deltag0: fixed

src/lib/image/TwkMovie/TwkMovie/MovieReader.h
deltag0: fixed

src/lib/image/stream/StreamPreloadPool.cpp
deltag0: fixed

src/lib/mu/MuQt6/qt2mu.py
deltag0: I ran it manually on all files

deltag0: 120000 9414f4a9c782f1526d63120dacfd55b461aac97b 0	src/lib/mu/MuQt6/qt2mu.py

I found that this means it's still a sym link

eloisebrosseau: But you shouldn't be touching it since it is unrelated to your PR. I just went back to your commit just before the reformat one and ran `pre-commit run --all-files`, and it did not touch this one.

src/lib/image/TwkMovie/MovieIO.cpp
eloisebrosseau: You forgot one `options.push_back` on line 129

src/lib/image/TwkMovie/MovieIO.cpp
eloisebrosseau: Can you do it for the rest of the code? This was just one example

src/lib/image/TwkMovie/MovieIO.cpp
deltag0: fixed

src/lib/mu/MuQt6/qt2mu.py
deltag0: Could it be because I ran it from my home machine? Either way, it's restored.

src/lib/image/stream/StreamPreloadPool.cpp
eloisebrosseau: Thank you for moving it but I think you forgot to also make it constexpr

src/lib/mu/MuQt6/qt2mu.py
eloisebrosseau: I have no idea! But thank you!

src/lib/image/TwkMovie/MovieIO.cpp
deltag0: fixed

src/lib/image/TwkMovie/MovieIO.cpp
deltag0: fixed

src/lib/image/stream/StreamPreloadPool.cpp
deltag0: fixed

src/lib/image/MovieFFMpeg/MovieFFMpeg.cpp
```diff
@@ -20,6 +20,10 @@
 #include <TwkUtil/PathConform.h>
 #include <TwkUtil/File.h>
 #include <TwkUtil/sgcHop.h>
+#include <stream/StreamPreloadPool.h>
+#include <boost/thread/lock_algorithms.hpp>
```
eloisebrosseau: Is this still being used in this file? If not, you should remove it. Also, you should put back `<boost/thread/mutex.hpp>` with the other boost header files where it was before

deltag0: fixed

src/lib/image/stream/StreamPreloadPool.cpp
deltag0: The object in the queue is destroyed when we pop it

bernie-laberge:
Closing this PR but saving this work preciously until we do the actual implementation.

PR #1384, opened 2026-08-20
fix: SG-40934: Defer -eval/-pyeval until display groups exist

cedrik-fuoco-adsk:
### Linked issues

SG-40934

### Summarize your change.

The `-eval` (Mu) and `-pyeval` (Python) command line expressions are no longer evaluated from `RvSession::postInitialize()`. They are split out into a new `RvSession::evalCommandLineScripts()` method, which the interactive path calls at the end of `RvApplication::newSessionFromFiles()`. After `setPhysicalDevices()` has created the display groups and the session is initialized.

`postInitialize()` runs while the `RvDocument` is still being constructed, well before `RVDisplayGroup` nodes exist. Only the default `RVOutputGroup` is present at that point. When media is loaded via `-pyeval`, `source_setup.py`'s `setDisplayFromProfile()` looks for an `RVDisplayGroup`, finds none, and silently applies nothing — so the Display Transfer Function stays "None" instead of "sRGB".

This is why enabling `RV_PROGRESSIVE_SOURCE_LOADING=1` works around the bug: it defers the real media load to an async job, by which time the display groups exist. It also explains the regression window — progressive source loading was on by default before RV 2022.0.0.

Fixing the ordering addresses the root cause, so no workaround is needed in `source_setup.py`. It also fixes the same problem for `-eval`, and for the reported case of applying `flop` to a source from the command line.

- [x] `rv -pyeval "from rv import commands;commands.addSourceVerbose(['some.mov'])"`
      → Display Transfer Function is "sRGB", with `RV_PROGRESSIVE_SOURCE_LOADING` unset

- `RvApp/RvSession.h`, `RvApp/RvSession.cpp`: new `evalCommandLineScripts()`; the
  `-eval`/`-pyeval` logic is moved verbatim out of `postInitialize()`.
- `RvCommon/RvDocument.cpp`: no longer evaluates the expressions during session
  construction.
- `RvCommon/RvApplication.cpp`: calls `evalCommandLineScripts()` at the end of
  `newSessionFromFiles()`.
- `rvio/main.cpp`, `MovieRV/MovieRV.cpp`: call it immediately after
  `postInitialize()`. These are headless, have no display groups to wait for, and
  keep their current timing.

**Needs attention:**

- **Behavior change** when a command line expression adds a source *and* file arguments are given, e.g. `rv pyeval "addSourceVerbose(['a.mov'])" b.mov`. Previously the expression ran first and its source was then wiped by the `clear()` in `rebuildSessionFromFiles()`, leaving only `b.mov`. Now both sources are loaded. Affects `-eval` and `-pyeval` equally.
- The call is placed *after* the `after-progressive-loading` block rather than immediately after `setPhysicalDevices()`. That block is gated on `s->loadTotal() == 0 && !files.empty()`; evaluating earlier could change `loadTotal()` and suppress an event that must close a previously emitted `before-progressive-loading`.
- The rvlink session-reuse path (`rebuildSessionFromFiles()` via `sessionFromUrl`) never called `postInitialize()`, so it already did not run `-pyeval`. That is unchanged here.
- Pre-existing quirk, intentionally not addressed: `m_initEval` clears itself after running while `m_pyInitEval` does not, so `-eval` is effectively once-per-process and `-pyeval` is per-session, despite both being documented as "at every session start".


bernie-laberge (approved):
Nice !

eloisebrosseau (approved):

PR #1385, opened 2026-08-20
feat: SG-44604: RVLinkLauncher - Save RV selection that was made in previous session

deltag0:
### Summarize your change.
* Added check if path was previously selected in preferences and select it
* Save last selected path in preferences

* Quality of life for users, so if RV is opened multiple times from shotgrid, users don't have to search for the last version they opened, it's selected by default.

* Tested on Mac CY2026


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #1397, opened 2026-09-01
fix: SG-44533: Insert new line on keyboard shortcut

eloisebrosseau:
### Summarize your change.

A new `on_text_new_line` method was added to insert a new line when the `shift-enter` keyboard shortcut is detected. This was a missing method that is present in the current annotation code in Mu. Note that both "enter" keys (i.e. the regular one and the numpad one) are detected to mirror the same logic as with the `enter` keyboard shortcut for committing text. `strip()` was added to `_commit_text` to make sure not to commit annotations that are only spaces or newlines. Note that the `shift_map` logic was removed because while investigating this shift key issue, I saw that it was actually not used at all. For instance, "!" is already received as `key-down--!` not `key-down--shift--1`, and "A" is `key-down--A` not `key-down--shift--a`.

The new annotation tool was missing support for multi-line text annotations. The `shift-enter` keyboard shortcut should add a new line, just like the current annotation tool does.

Adding new lines to text annotations using the `shift-enter` keyboard shortcut was tested on macOS.

### DIFF
--- a/src/plugins/rv-packages/annotate_beta/annotate_beta_engine.py
+++ b/src/plugins/rv-packages/annotate_beta/annotate_beta_engine.py
@@ -48,31 +48,6 @@ _SIZE_MIN = 0.001
 #   - existing text scales with the image as you zoom in (fixed WCS stored)
 _FONT_SIZE_WCS_BASE = {"small": 24.0 / 1080.0, "medium": 48.0 / 1080.0, "large": 72.0 / 1080.0}
 
-# US keyboard shift-symbol mapping used for text input
-_SHIFT_MAP = {
-    "1": "!",
-    "2": "@",
-    "3": "#",
-    "4": "$",
-    "5": "%",
-    "6": "^",
-    "7": "&",
-    "8": "*",
-    "9": "(",
-    "0": ")",
-    "-": "_",
-    "=": "+",
-    "[": "{",
-    "]": "}",
-    "\\": "|",
-    ";": ":",
-    "'": '"',
-    ",": "<",
-    ".": ">",
-    "/": "?",
-    "`": "~",
-}
-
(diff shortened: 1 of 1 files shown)


src/plugins/rv-packages/annotate_beta/annotate_beta_engine.py
```diff
@@ -147,6 +122,7 @@ def bindings(self):
             ("key-down--backspace", self.on_text_backspace, "Delete char"),
             ("key-down--delete", self.on_text_backspace, "Delete char"),
             ("key-down--space", self.on_text_space, "Insert space"),
+            ("key-down--shift--enter", self.on_text_new_line, "Insert a new line"),
```
bernie-laberge, line 125: Interesting comment from claude code that we might want to address: Shift+Numpad-Enter doesn't insert a newline (annotate_beta_engine.py:125) — Only "key-down--shift--enter" is bound to the new newline handler. RV's Qt event translator (QTTranslator.cpp) emits a distinct "-keypad-enter" suffix for the numeric keypad's Enter key, and the existing commit-binding already treats keypad-enter as equivalent to enter. The new newline binding doesn't mirror that, so Shift+Numpad-Enter is a silent no-op instead of inserting a newline.

bernie-laberge (approved):

src/plugins/rv-packages/annotate_beta/annotate_beta_engine.py
eloisebrosseau: `key-down--shift--keypad-enter` was added based on the discussion we had outside of this thread with design. Thank you for flagging it!

cedrik-fuoco-adsk (approved):

PR #1399, opened 2026-09-03
build: SG-45024 - Support C++20

cedrik-fuoco-adsk:
### Linked issues
none

### Summarize your change.
- Setting CPP version based on the VFX platform year.
- Ignore warning from PySide6. A lot of warning was logged on MacOS for PySide6.
- Stop using eigen for Imath with custom code to cover the functionality

OpenRV-pub PR: https://github.com/shotgunsoftware/openrv-pub/pull/22

Eigen will be re move from OpenRV-pub for both CY2025 and CY2026.

Fix issues while compiling OpenRV with C++20 while staying backwards compatible with C++17.

MacOS and Linux
Testing Windows compilation with CI


cmake/defaults/rv_options.cmake
```diff
@@ -81,6 +63,43 @@ SET_PROPERTY(
...
+  )
+ELSE()
+  SET(_RV_CPP_STANDARD
+      "17"
+  )
+ENDIF()
+
+SET(RV_CPP_STANDARD
+    "${_RV_CPP_STANDARD}"
+    CACHE STRING "RV's general C++ coding standard"
```
bernie-laberge: This is the only important finding  from our AI friend. Adding `FORCE` at the end would save us some trouble I think:
cmake/defaults/rv_options.cmake:82 — RV_CPP_STANDARD is set as a cache var without FORCE, unlike RV_VFX_PLATFORM right above it. Existing build trees reconfigured to a newer VFX platform won't actually pick up C++20.

bernie-laberge (approved):
LGTM thanks !
The only comment that we might address
1. cmake/defaults/rv_options.cmake:82 — RV_CPP_STANDARD is set as a cache var without FORCE, unlike RV_VFX_PLATFORM right above it. Existing build trees reconfigured to a newer VFX platform won't actually pick up C++20.

src/lib/app/RvPackage/RvPackage/PackageManager.h
```diff
@@ -117,8 +116,7 @@ namespace Rv
             QString url;
             QString icon;
             QString excludes;
-            QString
-                requires;
+            QString requires_;
```
eloisebrosseau: I'm not sure I like the addition of the underscore to fix the issue of `requires` being a protected keyword in C++20 since this is a convention used for member variables that we are not following anywhere else. Could it be renamed to something like `requiredPackages` instead?

src/build/make_pyside6.py
```diff
@@ -178,6 +178,36 @@ def get_fallback_clang_filename_suffix(version):
...
+    with open(old_shiboken_helpers_path) as old_shiboken_helpers:
+        old_content = old_shiboken_helpers.read()
+
+    new_content = old_content.replace(
+        'if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL  "GNU")\n'
+        '        set (gcc_warnings_options "${gcc_warnings_options} -Wno-cast-function-type")\n'
+        "    endif()",
+        'if ("${CMAKE_CXX_COMPILER_ID}" STREQUAL  "GNU")\n'
+        '        set (gcc_warnings_options "${gcc_warnings_options} -Wno-cast-function-type")\n'
+        '    elseif ("${CMAKE_CXX_COMPILER_ID}" STREQUAL "Clang")\n'
```
eloisebrosseau: Are you sure this is actually working? In `cxx_defaults.cmake`, we use `MATCHES`, which is a regex, not `STREQUAL`. On macOS, we use AppleClang, not Clang, so all of this would be a no-op in that case. (This is what I see in the configure step of the build: `The CXX compiler identification is AppleClang 17.0.0.17000013`)

src/build/make_pyside6.py
```diff
@@ -178,6 +178,36 @@ def get_fallback_clang_filename_suffix(version):
...
+        '        set (gcc_warnings_options "${gcc_warnings_options} '
+        '-Wno-error=cast-function-type-mismatch -Wno-cast-function-type-mismatch")\n'
```
eloisebrosseau: Why do you need both `-Wno-error=cast-function-type-mismatch` and `-Wno-cast-function-type-mismatch`?

cmake/defaults/rv_options.cmake
```diff
@@ -81,6 +63,45 @@ SET_PROPERTY(
...
+# The C standard stays at C17 on every VFX platform. It is deliberately not tied to RV_CPP_STANDARD: there is no C20, and CMake silently ignores an invalid
+# C_STANDARD value (emitting no -std flag at all, which would un-pin the C standard). C17 is the newest C standard with universal MSVC/GCC/Clang support, and
+# the VFX Reference Platform pins only the C++ standard.
+SET(RV_C_STANDARD
+    "17"
+    CACHE STRING "RV's general C coding standard"
+)
+SET_PROPERTY(
+  CACHE RV_C_STANDARD
+  PROPERTY STRINGS C99 11 17
```
eloisebrosseau: I know this was taking from existent code, but it should be 99, not C99.

src/lib/mu/MuMathLinear/MathLinearModule.cpp
```diff
@@ -16,25 +16,134 @@
...
+        }
+
+        template <typename M> void storeMatrix(const M& m, float* data) { memcpy(data, &m.x[0][0], sizeof(m.x)); }
+
+        void makeIdentity(float* m, size_t n)
+        {
+            fill(m, m + n * n, 0.0f);
+
+            for (size_t i = 0; i < n; i++)
+                m[i * n + i] = 1.0f;
```
eloisebrosseau, line 61: Nit: I think we agreed on always adding braces around statements even if they are one line. Could you fix it in the changes you made to this file?

src/lib/mu/MuMathLinear/MathLinearModule.cpp
```diff
@@ -16,25 +16,134 @@
...
+            makeIdentity(s.data(), n);
+
+            for (size_t i = 0; i < n; i++)
+            {
+                size_t pivot = i;
+                float pivotSize = fabsf(t[i * n + i]);
+
+                for (size_t j = i + 1; j < n; j++)
+                {
+                    const float mag = fabsf(t[j * n + i]);
```
eloisebrosseau: I know these are all math calculations so it makes sense to have one letter variables, but for things that are actual words, could we right it out completely instead of shortening them like you did for `mag`?

src/lib/app/RvPackage/RvPackage/PackageManager.h
cedrik-fuoco-adsk: I'll rename it

src/lib/mu/MuMathLinear/MathLinearModule.cpp
```diff
@@ -16,25 +16,134 @@
...
+        void invertSquareMatrix(const float* in, float* out, size_t n)
+        {
+            vector<float> t(in, in + n * n);
+            vector<float> s(n * n);
+            makeIdentity(s.data(), n);
+
+            for (size_t i = 0; i < n; i++)
+            {
+                size_t pivot = i;
+                float pivotSize = fabsf(t[i * n + i]);
```
eloisebrosseau: You should use `std::abs` instead of the C-style `fabsf`, since it is overloaded to handle floats directly.

src/build/make_pyside6.py
cedrik-fuoco-adsk: The warnings were gone, but you are right that MATCHES makes more sense here. I'll change it and re-test.

src/build/make_pyside6.py
cedrik-fuoco-adsk: Good catch! I added this while I was testing and forgot to remove it. I'll keep  `-Wno-cast-function-type-mismatch`

src/lib/mu/MuMathLinear/MathLinearModule.cpp
```diff
@@ -16,25 +16,134 @@
...
+
+        //
+        //  Mu fixed arrays and Imath matrices both store their elements
+        //  row-major, so the values can be copied verbatim.
+        //
+
+        template <typename M> M loadMatrix(const float* data)
+        {
+            M m;
+            memcpy(&m.x[0][0], data, sizeof(m.x));
```
eloisebrosseau: You should avoid using `memcpy` and use `std::copy_n` here instead: `std::copy_n(data, M::dimensions() * M::dimensions(), m.getValue())` and `std::copy_n(m.getValue(), M::dimensions() * M::dimensions(), data)` in `storeMatrix`.

cedrik-fuoco-adsk:
All comments should be covered!

eloisebrosseau (approved):

PR #1400, opened 2026-09-04
fix: SG-44508: Fix snapping shapes with stylus on hotkey

eloisebrosseau:
### Summarize your change.

Mirror the missing shift and move events from `pointer-1` (i.e. the mouse cursor) for the stylus pen. Since they were simply not there, nothing was allowing the shape to be move to 45-degree increments, which made the annotations appear frozen.

Holding the shift key while dragging a shape annotations would simply freeze the annotations at their current position instead of snapping and constraining them to 45-degree increments.

Adding shape annotations using a stylus pen with a Wacom Intuos Pro S and snapping it with the shift key was tested on macOS.


cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
LGTM

PR #1401, opened 2026-09-04
fix: SG-44509: Fix automatically mark frame

eloisebrosseau:
### Summarize your change.

If the "Automatically Mark Annotated Frames" option is enabled, mark the frame based on the global frame number instead of the source frame number since this is what `markFrame` is expected to get.

If the "Automatically Mark Annotated Frames" option was enabled, the mark would be added to the wrong frame or to no frames at all, depending on how far the source frame value was from the global frame value. Since the global frame always starts at 1, clips also starting at 1 would get their marks on the right frame, but those starting at 0, would get the marks on the previous frames, and those starting at something way bigger, like 864000, wouldn't get any marks.

Drawing annotations with the "Automatically Mark Annotated Frames" option enabled with different clips was tested on macOS.

### DIFF
--- a/src/plugins/rv-packages/annotate_beta/annotate_beta_engine.py
+++ b/src/plugins/rv-packages/annotate_beta/annotate_beta_engine.py
@@ -336,7 +336,7 @@ class AnnotateDrawEngine:
 
         if getattr(self._mode, "_auto_mark", False):
             try:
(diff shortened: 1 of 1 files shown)


bernie-laberge (approved):
LGTM

cedrik-fuoco-adsk (approved):

PR #1402, opened 2026-09-04
fix: SG-44522: Update QSS of the new annotation tool

eloisebrosseau:
### Summarize your change.

- Move all the hardcoded stylesheets definitions from the Python files to the different QSS files. Note that the template files are not matching properly the QSS files right now on main, so a follow-up PR to fix that will be made after this one. However they were still used generate the new properties in the QSS files, while removing everything that got generated because of the differences with the rest of the files. The legacy files were also updated, but to a minimum, to match more closely the overall theme instead of matching the new design that was based on the new themes. Windows and Linux are also sharing the same QSS files, which is the reason why there is only one for Linux and one for macOS. Some changes were made to the overall look of the UI to be closer to the design shown on Figma. 

**The majority of the changes are from all the QSS files. You don't need to review all of them since both template files are identical to each other, and that the only difference between them and the QSS files is the the variable names defined in `rv_theme_variable.conf` that were replaced with actual RGB values. The legacy QSS files contain a subset of what is defined in the non-legacy ones.** 

- Rename *colour* to *color* in `annotate_beta_engine.py` and `annotate_beta_widget.py`. The only needed change was the tooltip name that should use the American spelling instead of the Canadian spelling based on the Figma design. However, to avoid having two spellings (which was also the case because Qt is also using the American spelling), I felt like it would be more consistent to rename everything instead of just the tooltip.

- Remove unnecessary svg files that were not actually needed.

- Remove anything related to PySide2 in favor of PySide6 since there is no reason to support both anymore.

- `annotate_beta_widget.py` was updated to use the components defined in the QSS files. The file was also partially cleaned up to remove, modify or add methods based on the migration of the style sheets to QSS and the use of Qt's components as much as possible instead of custom ones. For instance, instead of using a custom slider, we are now using a QSlider, so a lot of functionality that was added to that custom widget was removed or updated to only the calculation of the handle position. For the same reason, the custom title bar was also replaced with a QDockWidget.

The new annotation tool had all the stylesheets define directly in the Python code instead of reusing or defining it's own properties in the application's QSS files. This is important to make sure the panel matches with the overall theme of the application that is customizable. Also, some parts of the UI could have been a little closer to the Figma design.

The whole tool bar was tested on both macOS and Rocky Linux 9 to make sure the functionalities are still behaving the way they should.

**BEFORE**

Screenshot 2026-09-04 at 16 33 37

**AFTER**

Screenshot 2026-09-04 at 16 32 13


src/plugins/rv-packages/annotate_beta/annotate_beta_color_picker.py
```diff
@@ -1,10 +1,7 @@
-# Copyright (c) 2025 Autodesk, Inc. All Rights Reserved.
+# Copyright (c) 2026 Autodesk, Inc. All Rights Reserved.
 # SPDX-License-Identifier: Apache-2.0
 
-try:
-    from PySide2 import QtCore, QtWidgets, QtGui
-except ImportError:
-    from PySide6 import QtCore, QtWidgets, QtGui
+from PySide6 import QtCore, QtWidgets, QtGui
```
bernie-laberge: Only supporting Qt6: Won't that break the main branch built with VFX CY2023 ?

src/plugins/rv-packages/annotate_beta/annotate_beta_widget.py
```diff
@@ -412,154 +129,58 @@ def mousePressEvent(self, event):
...
         self.setSizePolicy(QtWidgets.QSizePolicy.Fixed, QtWidgets.QSizePolicy.Expanding)
         self.setCursor(QtCore.Qt.PointingHandCursor)
         self.setFocusPolicy(QtCore.Qt.StrongFocus)
 
-    def _handle_y(self):
-        top, bot = self._PAD, self.height() - self._PAD
-        t = (self._value - self._min) / max(1, self._max - self._min)
-        return int(bot - t * (bot - top))
+    _HANDLE_LENGTH = 18
+    _GROOVE_MARGIN = 3
```
bernie-laberge: Slider hit-test math doesn't match rendered geometry (annotate_beta_widget.py:155) — _GROOVE_MARGIN = 3 is hardcoded but the QSS sets margin: 0px. Drag position and value are off by ~6px total span, causing the handle to drift from the cursor and saturate early near the track ends.

Feel free to ignore if this is not signifcant to cause an issue.

bernie-laberge (approved):
LGTM

src/plugins/rv-packages/annotate_beta/annotate_beta_widget.py
eloisebrosseau: Fixed! I also updated the legacy QSS files to make sure the geometry matches there as well! Thank you for flagging it!

cedrik-fuoco-adsk (approved):

PR #1407, opened 2026-09-08
fix: SG-45139: Fix viewport scaling when moving RV between displays with different DPI

cedrik-fuoco-adsk:
###

### Linked issues
none

### Summarize your change.
Moving the RV window to a display with a different scale factor left the viewport rendering at
the wrong scale: the image was either magnified into a corner, or drawn small in the bottom-left.
It stayed wrong until the window was manually resized. Dragging *across* a display boundary showed
the same corruption for as long as the window straddled the two screens.

Two independent defects, both introduced when the viewport became a native `QOpenGLWindow`:

1. **The native drawable was never resized.** Qt sizes the surface from *logical* geometry, and
   only a geometry change pushes a new size down to the platform window. A DPI change leaves the
   logical geometry untouched, so the drawable kept its old pixel size while everything derived
   from the video device (`glViewport`), the default `GLFBO`, the render geometry, correctly used
   the new one. GL clipped the frame to the stale surface.

2. **Two competing sources of truth for the device pixel ratio.** `QTGLVideoDevice` sized every
   frame from a cached ratio refreshed only by a screen lookup on the viewport's *centre point*,
   while Qt sized the surface from the live ratio. AppKit flips a window's backing scale on its
   own rule, which does not coincide with that centre-point probe, so the two disagreed for the
   whole time the window straddled two displays.

`GLWindow` now tracks the ratio and, when it changes, re-establishes the drawable and notifies the
session, and does so *before* the frame is painted rather than a turn later. `QTGLVideoDevice`
takes the ratio for the window-backed control viewport straight from its own surface.

`#1375` moved the viewport off `QOpenGLWidget` onto a native `QOpenGLWindow` embedded via
`QWidget::createWindowContainer()`. That PR's own description flagged this area as untested:

> **HDPI.** Coordinate translation now goes through the container widget while rendering goes
> through the window, so `devicePixelRatio` paths deserve a mixed-DPI pass.

This is that pass. The relevant Qt 6.5.3 behaviour, for reviewers:

- `QOpenGLWindow::resizeGL()` has exactly one call site, `resizeEvent()`
  (`qopenglwindow.cpp:641`), and `processGeometryChangeEvent()` only sends a `QResizeEvent` when
  the **logical** size changes (`qguiapplication.cpp:2670`). A DPI change therefore produces no
  `resizeGL()` at all. `QWindowContainer` does not help — it never re-pushes the embedded window's
  geometry on a screen change (`qwindowcontainer.cpp:42,267`).
- `QCocoaGLContext` only calls `[m_context update]` when `m_needsUpdate` is set, and that is set by
  `NSViewFrameDidChange`, `NSWindowDidChangeScreen` and `NSApplicationDidChangeScreenParameters` —
  **not** by a backing-properties change (`qcocoaglcontext.mm:383-398`). A one-pixel resize round
  trip posts `NSViewFrameDidChangeNotification`, which is the portable way to force that update.
  It is the same remedy `RvDocument::showEvent()` already applies to the first-show case.
- `QWindow::devicePixelRatio()` is **not** cached in 6.5.3 — it round-trips to AppKit on every call
  (`qcocoawindow.mm:1927`), and `QOpenGLWindowPrivate::beginPaint()` uses it to set
  `glViewport(0, 0, w*dpr, h*dpr)` immediately before each `paintGL()` (`qopenglwindow.cpp:226`).
  Sourcing the device's ratio from the same place makes divergence impossible by construction.

**macOS 15 (Apple silicon), Qt 6.5.3, Release**, on a genuinely mixed-DPI setup: DELL U2725QE
3840x2160 at DPR 1.0 (main) plus the built-in Retina display at DPR 2.0.

Dragging the window back and forth across the display boundary is now visually clean, with no artifact while straddling.

**`GLWindow`** (`GLWindow.h`, `GLWindow.cpp`)

- New `syncDevicePixelRatio()` plus `m_devicePixelRatio` / `m_syncingDevicePixelRatio`. On a change
  it does a one-pixel resize round trip on the viewport window to re-establish the drawable, then
  calls `RvDocument::viewSizeChanged()` so the renderer's image FBOs flush and `view-size-changed`
  fires. Poking the viewport window rather than the document keeps it off the main window's layout.
- Called from `GLWindow::event()` **before** delegating to the base class, on `UpdateRequest`,
  `Paint`, `Expose` and `Move`.
- Queued `QWindow::screenChanged` connection as an additional trigger.

**`QTGLVideoDevice`** (`QTGLVideoDevice.h`, `QTGLVideoDevice.cpp`)

- `devicePixelRatio()` becomes out-of-line; for the window-backed control viewport it returns
  `m_window->devicePixelRatio()`. Widget-backed presentation (`ScreenView`) and worker devices keep
  the cached value, so their behaviour is unchanged.
- The `RV_NO_QT_HDPI_SUPPORT` static is hoisted into a shared `qtHighDPISupportDisabled()` helper
  so both paths honour it.
- `setPhysicalDevice()`'s screen lookup is left in place — it still drives the refresh rate and the
  `video-device-changed` event, it just stops being load-bearing for the ratio.

### DIFF
--- a/src/lib/app/RvCommon/GLWindow.cpp
+++ b/src/lib/app/RvCommon/GLWindow.cpp
@@ -23,6 +23,7 @@
 #include <TwkApp/VideoDevice.h>
 #include <TwkGLF/GLVideoDevice.h>
 #include <QOpenGLContext>
+#include <QScreen>
 #include <QtGui/QGuiApplication>
 #include <QKeyEvent>
 #include <QResizeEvent>
@@ -35,6 +36,37 @@ namespace Rv
     using namespace std;
     using namespace TwkApp;
     using namespace IPCore;
+
+    namespace
+    {
+        //
+        //  Sets a flag for the duration of a scope.
+        //
+        //  Holds a reference to the caller's flag, so the destructor clears the
+        //  one and only copy of it on every exit path, including an exception
+        //  thrown from a nested Qt or script callback. Copying is deleted
+        //  because the first destructor to run would clear the flag while the
+        //  other guard still believes it holds it.
+        //
+        //  Declare it as a named local: an unnamed temporary is destroyed at
+        //  the end of its own statement and guards nothing.
+        //
+        struct ScopedFlag
+        {
+            explicit ScopedFlag(bool& f)
+                : m_flag(f)
+            {
+                m_flag = true;
+            }
+
+            ~ScopedFlag() { m_flag = false; }
+
+            ScopedFlag(const ScopedFlag&) = delete;
+            ScopedFlag& operator=(const ScopedFlag&) = delete;
+
+            bool& m_flag;
+        };
+    } // namespace
 
     GLWindow::GLWindow(QOpenGLContext* sharedContext, RvDocument* doc, bool stereo, bool vsync, bool doubleBuffer, int red, int green,
                        int blue, int alpha, bool noResize)
@@ -50,6 +82,8 @@ namespace Rv
         , m_firstPaintCompleted(false)
         , m_postFirstNonEmptyRender(noResize)
         , m_stopProcessingEvents(false)
+        , m_devicePixelRatio(1.0f)
+        , m_syncingDevicePixelRatio(false)
         , m_sharedContext(sharedContext)
     {
         setFormat(GLView::rvGLFormat(stereo, vsync, doubleBuffer, red, green, blue, alpha));
@@ -60,6 +94,18 @@ namespace Rv
 
         m_eventProcessingTimer.setSingleShot(true);
         connect(&m_eventProcessingTimer, SIGNAL(timeout()), this, SLOT(eventProcessingTimeout()));
+
+        m_devicePixelRatio = static_cast<float>(devicePixelRatio());
+
+        //
+        //  Moving to another display is the usual way the ratio changes.
+        //  QWindow emits screenChanged() recursively, so this embedded viewport
+        //  gets it when the top level is dragged to another monitor.
+        //
+        //  Queued: screenChanged() is emitted before devicePixelRatio() reports
+        //  the new value, so a direct call would always see the old ratio.
+        //
+        connect(this, &QWindow::screenChanged, this, [this](QScreen*) { syncDevicePixelRatio(); }, Qt::QueuedConnection);
     }
 
     GLWindow::~GLWindow() {}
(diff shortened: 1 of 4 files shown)


src/lib/app/RvCommon/GLWindow.cpp
```diff
@@ -130,6 +176,59 @@ namespace Rv
             m_doc->viewSizeChanged(w, h);
     }
 
+    void GLWindow::syncDevicePixelRatio()
+    {
+        const float dpr = static_cast<float>(devicePixelRatio());
+
+        if (dpr == m_devicePixelRatio || m_syncingDevicePixelRatio)
```
bernie-laberge: Potential infinite update loop (GLWindow.cpp:183) — syncDevicePixelRatio() gates on exact float equality (dpr == m_devicePixelRatio). If a platform ever returns a non-bit-identical ratio across calls (e.g. Windows fractional-scaling rounding jitter — only macOS was tested), the early-return never triggers, causing a resize→viewSizeChanged()→requestUpdate()→re-entry loop that pegs a CPU/GPU core and flickers continuously.

src/lib/app/RvCommon/RvCommon/GLWindow.h
```diff
@@ -68,6 +68,14 @@ namespace Rv
 
         float devicePixelRatioF() const;
 
+        //
+        //  Re-establish the surface and notify the session when the device
+        //  pixel ratio changes without a change of logical size. This is what
+        //  moving to a display with a different scale factor does, and which Qt
+        //  reports through no resize of its own. See the definition.
+        //
+        void syncDevicePixelRatio();
```
bernie-laberge: Minor: unnecessarily public method (GLWindow.h:77) — syncDevicePixelRatio() is public but only ever called internally; should be private.

bernie-laberge (approved):
LGTM But I would fix the float comparison to be safe

src/lib/app/RvCommon/GLWindow.cpp
```diff
@@ -60,6 +95,18 @@ namespace Rv
 
         m_eventProcessingTimer.setSingleShot(true);
         connect(&m_eventProcessingTimer, SIGNAL(timeout()), this, SLOT(eventProcessingTimeout()));
+
+        m_devicePixelRatio = static_cast<float>(devicePixelRatio());
```
eloisebrosseau, line 99: `m_devicePixelRatio` should be initialized in a member initializer of the constructor (https://clang.llvm.org/extra/clang-tidy/checks/cppcoreguidelines/prefer-member-initializer.html)

src/lib/app/RvCommon/GLWindow.cpp
```diff
@@ -130,6 +177,69 @@ namespace Rv
             m_doc->viewSizeChanged(w, h);
     }
 
+    void GLWindow::syncDevicePixelRatio()
+    {
+        const float dpr = static_cast<float>(devicePixelRatio());
```
eloisebrosseau, line 182: Please rename this variable with a more descriptive name. There is no reason for it to be abbreviated.

src/lib/app/RvCommon/GLWindow.cpp
```diff
@@ -130,6 +177,69 @@ namespace Rv
...
+        //  Compare with a tolerance rather than exactly. Only macOS, which
+        //  reports whole-number ratios, has been tested; a platform that
+        //  derives the ratio from fractional scaling (Windows at 125%, 150%)
+        //  can return values that differ in the last bits from one call to the
+        //  next. An exact comparison would never take the early return there,
+        //  and since the work below ends in requestUpdate() (and the session notification) 
+        //  that brings us back here, that would spin continuously. 
+        //  The tolerance is far below any real difference in scale factor between two displays.
+        //
+        if (std::abs(dpr - m_devicePixelRatio) < 1e-4f || m_syncingDevicePixelRatio)
```
eloisebrosseau, line 194: `1e-4f` is a magic number. Can you replace it with a named constant?

src/lib/app/RvCommon/GLWindow.cpp
eloisebrosseau: Also, you should use `auto` when initializing with a cast to avoid duplicating the type name.

src/lib/app/RvCommon/GLWindow.cpp
```diff
@@ -50,6 +83,8 @@ namespace Rv
         , m_firstPaintCompleted(false)
         , m_postFirstNonEmptyRender(noResize)
         , m_stopProcessingEvents(false)
+        , m_devicePixelRatio(1.0f)
```
eloisebrosseau, line 86: Do you really need to initialize it to 1.0f just to assigned devicePixelRatio() a few line below of the constructor's body? It could be initialized directly in the header file instead since it's just a default value (i.e. `float m_devicePixelRatio{1.0f`).

eloisebrosseau (commented):
You added a lot of comments in the code. Are they really all needed? Especially the comments you added in both the header and the source files for the same functions. Can we keep the comments to a few lines that are actually needed to understand the code. Otherwise, I feel like it is a lot of bloat.

src/lib/app/RvCommon/GLWindow.cpp
```diff
@@ -40,18 +40,12 @@ namespace Rv
...
-        //  thrown from a nested Qt or script callback. Copying is deleted
-        //  because the first destructor to run would clear the flag while the
-        //  other guard still believes it holds it.
-        //
-        //  Declare it as a named local: an unnamed temporary is destroyed at
-        //  the end of its own statement and guards nothing.
-        //
+        //  Below any real difference between two displays, above the
+        //  last-bit jitter a fractionally scaled ratio can show.
+        constexpr float DevicePixelRatioTolerance = 1e-4f;
```
eloisebrosseau: Why are you declaring `DevicePixelRatioTolerance` here instead of just before the if statement in `syncDevicePixelRatio` since it's only called there?

eloisebrosseau (approved):

PR #1415, opened 2026-09-16
fix: SG-45240: Fix annotation cursor

eloisebrosseau:
### Summarize your change.

- Update `setCursor` in `MuUICommands.cpp` to set the cursor shape on the QOpenGLWindow.
- Add `sessionGLWindow()` in `PyUICommands.cpp` to access the QOpenGLWindow Python through `qtutils.py`
- Define and set a cursor shape for each tool of the new annotation package in `annotate_beta_widget.py` and `annotate_beta_mode.py` to match the old annotation package cursor's behaviour.

The viewport (i.e. `GLView`) used to be a `QOpenGLWidget`. However, since #1375, it became a `QWidget`, and the actual rendering is now done by `GLWindow`, which is a `QOpenGLWindow`. Because of this change, `setCursor` was now a no-op. This means that both the old annotation tool and the manual layout view, that were both setting cursor shapes for some actions, lost that behaviour. Also, in `annotate_beta_mode.py`, the cursor shape for the eyedropper was applied on `sessionWindow()` which caused the cursor shape to be applied on the whole application window (i.e. QMainWindow) instead of just the viewport. Moreover, other tools were missing a specific cursor shape to match the old annotation package behaviour.

Selecting tools with different cursor shapes was tested on macOS.

**BEFORE**

https://github.com/user-attachments/assets/4b6d962d-b2ae-4273-be39-a782a82303f8

**AFTER**

https://github.com/user-attachments/assets/799fee43-ca70-4b35-aa8d-743d3b5ddd64


cedrik-fuoco-adsk (approved):

src/lib/app/RvCommon/MuUICommands.cpp
```diff
@@ -955,7 +956,7 @@ namespace Rv
     {
         Session* s = Session::currentSession();
         RvDocument* rvDoc = (RvDocument*)s->opaquePointer();
-        rvDoc->view()->setCursor(QCursor(Qt::CursorShape(NODE_ARG(0, int))));
+        rvDoc->view()->glWindow()->setCursor(QCursor(Qt::CursorShape(NODE_ARG(0, int))));
```
bernie-laberge, line 959: This code could crash: rvDoc->view()->glWindow() could be potentially null: GLView::m_glWindow can become null during window teardown and setCursor is called from mode deactivate() handlers, this is a plausible crash path during exactly that teardown window.
In the refactoring of RV viewport (https://github.com/AcademySoftwareFoundation/OpenRV/pull/1375), GLView became a facade.
I think we should add the setCursor() implementation in /Users/labergb/openrv_bernie3/OpenRV/src/lib/app/RvCommon/RvCommon/GLView.h/.cpp. This way you could make sure that m_glWindow is not nullptr before executing the setCursor(). If we do that, then we won't need to modify MuUICommands.cpp. 
What do you think ?

src/plugins/rv-packages/annotate_beta/annotate_beta_mode.py
```diff
@@ -341,17 +343,16 @@ def _on_tool_changed(self, tool):
...
-        sw = qtutils.sessionWindow()
-        if tool == TOOL_EYEDROPPER:
-            sw.setCursor(QtCore.Qt.CrossCursor)
-        else:
-            sw.unsetCursor()
+        self._set_cursor(TOOLS[tool].cursor)
+
+    def _set_cursor(self, cursor):
+        session_window = qtutils.sessionGLWindow()
+        session_window.setCursor(cursor)
```
bernie-laberge, line 350: qtutils.sessionGLWindow() could return None.
So session_window.setCursor(cursor) would be called on a None object. 
If we go ahead with adding an implementation for GLView::setCursor(), then you could use a qtutils.sessionGLView() object instead. sessionGLWindow() would no longer be necessary.

src/plugins/rv-packages/annotate_beta/annotate_beta_widget.py
```diff
@@ -28,35 +30,30 @@
...
+    page: int
+    tooltip: str
+    cursor: QtCore.Qt.CursorShape
+
+
+TOOLS = {
+    TOOL_CURSOR: Tool(page=_PAGE_EMPTY, tooltip="Cursor", cursor=QtCore.Qt.ArrowCursor),
+    TOOL_EYEDROPPER: Tool(page=_PAGE_EMPTY, tooltip="Eyedropper", cursor=QtCore.Qt.CrossCursor),
+    TOOL_PEN: Tool(page=_PAGE_PEN, tooltip="Pen", cursor=QtCore.Qt.CrossCursor),
+    TOOL_AIRBRUSH: Tool(page=_PAGE_BRUSH, tooltip="Airbrush", cursor=QtCore.Qt.CrossCursor),
```
bernie-laberge, line 50: In the previous code, TOOL_AIRBRUSH was associated with the _PAGE_PEN page. 
And in the code it says:
`_PAGE_PEN = 3  # pen and airbrush (size/opacity/blend mode)`
Perhaps we should keep the TOOL_AIRBRUSH associated with the _PAGE_PEN page ?
And rename the _PAGE_PEN to _PAGE_BLEND maybe because this is so tempting to do what you just did ! I would have done it.

src/plugins/rv-packages/annotate_beta/annotate_beta_widget.py
```diff
@@ -28,35 +30,30 @@
...
+    TOOL_ERASER: Tool(page=_PAGE_ERASER, tooltip="Eraser", cursor=QtCore.Qt.CrossCursor),
+    TOOL_RECT: Tool(page=_PAGE_SHAPE, tooltip="Rectangle", cursor=QtCore.Qt.CrossCursor),
+    TOOL_CIRCLE: Tool(page=_PAGE_SHAPE, tooltip="Circle", cursor=QtCore.Qt.CrossCursor),
+    TOOL_ARROW: Tool(page=_PAGE_SHAPE, tooltip="Arrow", cursor=QtCore.Qt.CrossCursor),
+    TOOL_LINE: Tool(page=_PAGE_SHAPE, tooltip="Line", cursor=QtCore.Qt.CrossCursor),
```
bernie-laberge, line 55: In the previous code, TOOL_ARROW and TOOL_LINE were associated with the _PAGE_BRUSH page.
And in the code it says:
`_PAGE_BRUSH = 0  # arrow, line`
I think we need to keep them associated with the _PAGE_BRUSH page.
And rename the _PAGE_BRUSH to _PAGE_SIZE_OPACITY

PR #1416, opened 2026-09-16
fix: SG-44534: Fix missing text caret on push events

eloisebrosseau:
### Summarize your change.

Add the "|" placeholder on `on_push()` events that are triggered when clicking on the viewport through `_ensure_text_node()`. Note that some other changes were needed to make sure undo-redo was properly updated with that new behaviour to avoid registering the text caret as a real annotation:

- `_ensure_text_node()` is now only responsible for creating a text node, and `_register_text_undo()` is responsible for updating the undo stack when a new text annotation is registered. This also fixed another issue where every text node was pushed onto the undo stack twice.
- `_frame_order_and_undo()` was split into `_add_to_order()` and `_register_undo_entry()` to avoid counting a click without typing as an annotation.
- `_cancel_text` was updated to fix an issue where clicking on the `Escape` key after typing would leave a stale undo entry that could be displayed again by clicking on undo, followed by redo.
- Add `commit_text_if_active()` to `clear_frame()` and `clear_all_frames()` to ensure any in-progress text annotation is committed or canceled before the frame is cleared, and avoid a desync between the node state and the undo stack.

When clicking on a frame with the text tool selected from the new annotation panel, no "|" placeholder was added to confirm that text would start being displayed at that position once the user would start to type.

Adding text annotations was tested on macOS.


cedrik-fuoco-adsk (approved):

src/plugins/rv-packages/annotate_beta/annotate_beta_engine.py
```diff
@@ -549,17 +568,22 @@ def _update_text_display(self, cursor=True):
         except Exception as e:
             print(f"[annotate_beta] _update_text_display error: {e}")
 
+    def _reset_text(self):
```
bernie-laberge, line 571: Good eye. nice cleanup !

bernie-laberge (approved):
LGTM

bernie-laberge:
FYI: From our robot friend but feel free to ignore if not relevant/significant:
1. Confirmed data-loss bug (annotate_beta_engine.py:550) — pressing Ctrl+Z while mid-typing a text annotation soft-deletes the live node but leaves the editing session active, so the user keeps typing into a deleted node and loses the text silently on commit.
2. Plausible regression (annotate_beta_engine.py:878) — on_push now eagerly creates and network-syncs an empty placeholder text node on every click of the Text tool, even if the user never types anything. This both leaks orphaned properties locally and pushes a stray empty annotation to Live Review peers.

eloisebrosseau:
> FYI: From our robot friend but feel free to ignore if not relevant/significant:
> 
> 1. Confirmed data-loss bug (annotate_beta_engine.py:550) — pressing Ctrl+Z while mid-typing a text annotation soft-deletes the live node but leaves the editing session active, so the user keeps typing into a deleted node and loses the text silently on commit.
> 2. Plausible regression (annotate_beta_engine.py:878) — on_push now eagerly creates and network-syncs an empty placeholder text node on every click of the Text tool, even if the user never types anything. This both leaks orphaned properties locally and pushes a stray empty annotation to Live Review peers.

@bernie-laberge I just pushed two small fixes to address both comments! For your second point, the issue you are talking about is the same behaviour that has always been present in the Mu annotation package, so I only removed the "|" placeholder from being saved as a char to match Mu's behaviour of saving it as an empty annotation. We can discuss if this is something we want to address in a further ticket, but I believe this is not a regression.

PR #1422, opened 2026-09-21
fix: SG-44988: Fix annotations drawn on empty canvas with new annotation package

eloisebrosseau:
### Summarize your change.

The Mu annotation package uses a tag so that `imagesAtPixel` only returns RVPaint nodes associated with the active view instead of any image rendered at that pixel. This is useful to properly detect the source we are currently on when there a multiple images (e.g. stack or layout view), and also contributes to correctly detecting an empty canvas. This behaviour was missing from the new annotation package. `_find_paint_node()` was also mistakenly returning the first paint node from the list of RVPaint nodes if none were found from the view node info. It should have simply returned None instead of an arbitrary paint node. This caused strokes to be registered when clicking on empty Sequence view canvas because the Default Layout node was the first in that list.

Note that a typo in the bindings for committing text was also fixed from `before-graph-view-changed` to `before-graph-view-change`.

With the new annotation package, strokes were registered on empty canvas or on the wrong node.

Drawing was tested on macOS for the following:

- Empty canvas for Sequence, Stack, and Layout views
- Movie clip loaded before and after opening the new annotation tool
- On a movie clip after clearing the first one


cedrik-fuoco-adsk (approved):

src/plugins/rv-packages/annotate_beta/annotate_beta_engine.py
```diff
@@ -847,6 +861,8 @@ def on_push(self, event):
                 return
             self._current_source_name = name
             paint_node, frame = self._find_paint_node()
+            if paint_node is None:
+                return
```
bernie-laberge, line 865: Should we event.reject() here like on line 886 or not ?

PR #1423, opened 2026-09-22
fix: SG-45168: Fix dock settings not saved between sessions

eloisebrosseau:
### Summarize your change.

Following `annotate_mode.mu`, the Python-based annotation package now saves and reads two dock settings between sessions: `dock_area` and `top_level`. They represent, respectively, whether the dock is in the right or the left dock area, and whether the dock is floating. the default area is now the left to match `annotate_mode.mu`. `_save_dock_settings` is called in `deactivate`, just like `_save_configure_settings`. Unlike the other settings that are read in `_load_settings`, the dock settings are read in `_create_dock`. `addDockWidget` needs `dock_area`, and `load_settings` can only be called once the dock has been created.The dock is set as floating after `show` in `_show_toolbar` so that Qt can reuse the dock's docked geometry for the floating panel. Note that `self._dock.hide()` was removed from `_create_dock` because this is no longer necessary since the mode is only created when the panel is first toggled, not at startup.

Note that some cleanup was done around the usage of the `g` variable used as an alias for `_SETTINGS_GROUP`, and the try catches around `commands.writeSettings` since it was inconsistent in the file. I was using both in the code I was adding so it made sense to fix it here. The `g` had no benefit, and it made the code less readable, since you had to scroll up to see what g stood for. `writeSettings` never throws, so the try catches were dead code.

The new annotation panel would always appear docked on the right of the app, no matter where is was docked last before closing the app.

Opening and closing both the annotation panel and the app with the panel docked on the right, left, and undocked (i.e. floating) was tested on macOS.


cedrik-fuoco-adsk (approved):

bernie-laberge (approved):
LGTM

